PR #45, opened 2015-10-10
Replaced a fixed hydraulic motor with a variable hydraulic motor

ratanakso:


cmichelen:
Are you replacing it or having both options? Also there are conflicts. It might be best if you cancel this pull request, fix the conflicts, and the pull request again.

Thanks!

-Carlos

ratanakso:
Need to fix the conflicts first.

PR #53, opened 2015-10-28
nonlinear force different step size

bradling:
# Overview
This pull request implements change that allow for the nonlinear forces to be calculated at a different sampling rate than the rest of the model. This is especially helpful if other aspects of your model, like a PTO, must be run at a smaller step size than is required by the hydrodynamics.

It is essentially a zero-order hold implemented on the nonlinear FK and hydrostatic forces, reducing the number of times the calculation must be performed. Shown below is the implementation for the FK Force:
```matlab
persistent oldForce
persistent oldWp
persistent oldWpMeanFs

if isempty(oldForce)
    [f, wp, wpMeanFS] = calc_force(x,...);
    oldForce = f;
    oldWp = wp;
    oldWpMeanFs = wpMeanFS;
else
    if mod(t, dtFe) < dt/2
        [f, wp, wpMeanFS] = calc_force(x,...);
        oldForce = f;
        oldWp = wp;
        oldWpMeanFs = wpMeanFS;
    else
        f = oldForce;
        wp = oldWp;
        wpMeanFS = oldWpMeanFs;
    end
end
```

### How To Use
The desired sampling rate for the nonlinear calculations is declared as a property of the simulationClass:
```matlab
simu.dtFeNonlin = 0.01;
```

If this property is not declared in the wecSimInputFile, simu.dtFeNonlin is set to simu.dt in the setupSim method.

### Verification
I did a quick test using RM3 to make sure that the code would still run when incorporated with some of the other changes that have been made recently. I did not upload any detailed verification results


cmichelen:
That is great! I have a few questions?
* Can you list everything you changed in the WEC-Sim library? My guess is that it was only the nonlinear FK and the nonlinear hydrostatic matlab functions. Is this correct?
* Does ``simu.dtFeNonlin`` have to be a multiple of ``simu.dt``? what happens if it isn't?

bradling:
Hi Carlos, let me answer your questions
* I changed these three matlab function blocks:
  * /Rigid Body/Nonlinear Wave Elevation/Linear or Instantaneous Free Surface/Instantaneous Water Free Surface/Instantaneous FS
  *  /Rigid Body/Nonlinear Wave Elevation/Linear or Instantaneous Free Surface/Instantaneous Water Free Surface  /Diffraction and Excitation Forces
  *  /Rigid Body/Hydrostatic Restoring Force Calculation/Linear and Nonlinear Restoring Force Variant Subsystem/Nonlinear Hydrostatic Restoring Force /nonLinearBuoyancy
  * Additionally - I had to add some inputs to each of these three blocks. Specifically, simu.dt, simu.dtFeNonlin, and time if it wasnt already input
* It is possible to specify a simu.dtFeNonlin that is not a multiple of simu.dt, however, you may get odd results. I just did a couple of quick tests, and what happens is the nonlinear forces are not calculated at an even sample rate. Instead, it is calculated whenever `mod(t, dtFe) < dt/2 ` is true. Depending on the ratio of dt/dtFe, the behavior will change. I used this logic to ensure that the excitation forces are calculated for each internal solver step when it should be calculated.
* Also, I should note, this probably won't be stable if a variable step size solver is used.

cmichelen:
Great. Thanks for the answers.

cmichelen:
The second block should be:
* WECSim_Lib/Body Elements/Rigid Body/Wave Diffraction and Excitation Force Calculation/Nonlinear FK Force Variant Subsystem/With Nonlinear FroudeKrylov Force/Diffraction and Excitation Forces

bradling:
Yes - I corrected my previous comment

Issue #74, opened 2015-12-03
Morison Elements not working for irregular waves


bradling:
When trying to run a simulation including a morison element using a bretschneider spectrum, I get the following error:

```
Error using wecSim (line 192)
        The input signal to port 6 of subsystem 'azura_half/float/Morrison Element and
        ViscousDampingForceCalculation/Morrison Element Variant Subsystem/Morrison Element On/Irregular Wave
        Excitation Force/MorrisonElements/MorisonElementIrregular' has a width of 1. This width must be a
        multiple of 301, the number of iterations expected by another subsystem input/output signals.
```

I think the problem is simulink is trying to loop through different time values for each wave component, instead of treating time as a constant.

PR #101, opened 2016-03-11
Fixed Morison Elements for Irregular waves

bradling:
This pull request fixes morison elements for irregular waves, which would close Issue https://github.com/WEC-Sim/WEC-Sim/issues/74.

The problem was the for each block that was calculating the water particle kinematics was trying to loop over time. I discussed this briefly with @yuyihsiang at the training course at OSU.

I also added `mooring` to the `clear` statement at the top of `wecSim.m`. I was having problems with a mooring object still being in memory, but running a model without a mooring in the slx file, causing a crash.


bradling:
Updated to fix conflicts. See pull request https://github.com/WEC-Sim/WEC-Sim/pull/103

yuyihsiang:
Brad created a new pull request (#103), issue solved.

PR #126, opened 2016-09-10
Free surface view, nlHydro fix/edits, changes to MCR 

cmcnatt:
Hi, I'm creating a pull request mostly to add my update of a free surface visualization. It also contains some other changes, which I'll describe below. I've never done a pull request before, so let me know if there is anything else I need to do.

1) added simu.nlHydro = 3, which runs a nonlinear hydrostatic force but a linear FK

2) modified the nonlinear FK calculation (and broke it into a separate function so that it was easier for me to test). In the original, the linear FK was computed at the global origin (0,0,0), but if the body drifts to some new horizontal location (x,y,0), then the body-exact FK was computed at (x,y,0) but the linear FK was computed at (0,0,0). I changed it to translate the computation of the linear FK to (x,y,0)

3) I modified the behavior of wecSimMCR to go through an ordered pair of [T, H] combos (rather than make a matrix), because we often run RAOs of waves of constant steepness.

4) Added a some more functionality to visualize the plots of the .stl files

5) Finally, added a free surface visualization. It is set with simu.viewExactFS = 1 (defaut is 0). It only works for unidirectional waves at 0 degrees, but it works for regular and irregular. It could be extended to directional seas at the cost of loading time in the simulation explorer.  It works by creating a free-surface out of 64 extruded rigid bodies lined up like a treadmill. These bodies are then translated and rotated according to the equations for the free surface. Right now it has to be 64 pieces, and when you look at the block's you'll see how I set it up. This is because the SimMultiBody bodies cannot be created or modified in a loop; each one has to be an actual block. I also created a separate block set for the Free Surface because I needed to be able to have a "Free Surface Unit" block which could be replicated, but where all the copies are linked to the original so that changes to the original are propagated throughout.


yuyihsiang:
Hi Cameron

That is great. Thank you for the contribution, and we really appreciate that.

I talked to Kelley today and we will take a look and make sure there is no conflict with our current version. Given that I am flying out for a meeting in Europe tomorrow and won't be back until the end of this week, I'll make it as a priority on my todo list next week and get back to you asap.

Thanks,

--Yi-Hsiang

cmcnatt:
Hi Yi-Hsiang,

No rush on the merge, and just let me know if there are any changes I can
make to facilitate it.

Are you going to the DTOcean meeting in Edinburgh by any chance?  I ask
because I live in Edinburgh.

Cheers,
Cameron

bradling:
Hi,

With regards to point number 2, I thought that the linear portion of FK was calculated at (0,0,0) and then removed because the linear excitation force calculated using BEM coefficients is used, which is calculated at (0,0,0). And this is done so that the linear diffraction terms are included in the calculation. My understanding is the linear FK calculation is done to remove the FK portion of the excitation force coming from BEM coefficients.

So would it be correct to remove the linear portion of Fe at (x,y,0) in the instantaneous calc, and then add in the linear portion of Fe from BEM results calculated at (0,0,0)? I think this change would introduce an inconsistency into the excitation force calculation.

And secondarily, if diffraction coefficients could be imported from BEM results, why calculate the linear portion at all in WEC-Sim? It would be more computationally efficient to calculate instantaneous FK once (linear + nonlinear), and then include the diffraction force by using the diffraction coefficients from BEM results instead of total Fe coefficients. This would require modifying BEMIO to import diffraction coefficients. The upside is faster nonlinear hydro simulations. 

I would be willing to made my proposed modifications to the library to do this, but since I don't have access to WAMIT or AQWA, I wouldn't be able to modify and test BEMIO to import diffraction coefficients.

Thanks,

Brad

cmcnatt:
Hi Brad,

I think I'll address your second point first, that of computing the linear
portion of the FK force at run time in WEC-Sim. I agree with you that it
should be pre-computed and then either 1) use the scattering force (linear
excitation - linear FK) and the exact FK force or 2) use the excitation
force and subtract the the precomputed linear FK from the exact FK.  WAMIT
can compute the linear FK force, or WEC-Sim could pre-compute the linear FK
force with the .stl file (which may be more consistent).

In WEC-Sim, right now, it does the second, except that as I said it
computes the linear FK at run time. I did not set it up this way; that's
how WEC-Sim was. I tried to correct it, but wasn't able to sort out a
simple fix in the time I had.

Now, let me try and explain the first point. One could compute the
excitation force at (x,y,z) = (0,0,0) for all run time, no matter the exact
position of the body - that is linear wave theory (LWT). To get WEC-Sim
results to exactly agree with LWT, that's what you have to do.

But in the time domain model, as the body move in the x-y plane, it ends up
at different phases of the wave. So, in fact, you could modify the linear
excitation force by multiplying by the phasor: exp(-i*k*(x*cos(beta) +
y*sin(beta))) for a positive time exponent, exp(i*w*t). Then as the body
moves in the x-y plane it experiences a different force depending on its
position. This is a nonlinearity, and it's not currently implemented in
WEC-Sim, but I think that it should be as an option. (It is the same thing
as modifying the excitation force to account new incident wave directions
for body yaw position).

I am attaching a picture that I quickly drew up to explain this. It's a bit
messy, but hopefully it makes the point.

So, now onto the nonlinear FK force. Looking at my messy picture, in the
dashed body position, the body is in a wave trough, but the linear position
is in a wave crest. The exact FK computes the pressures for the wave
trough, and so to get the nonlinear component of this, one needs to
subtract linear FK for the wave trough. Subtracting the mean position
results in major errors. That's how I found the problem - as it was, as my
WEC drifted from the mean position, it eventually went crazy.

Then, you can either use the linear excitation force at the mean position
or the linear exciation force at the exact position. The exact position
force is probably more correct, and can be obtain via the phasor I
described above. However, the time domain theory with the nonlinear FK is
not completely consistent with regards to physics anyway, so I don't think
there is an absolute right answer (don't get me wrong, I think the model
generally works and is needed for speed).

Ok, hope that wasn't too long.

I'd also like to add a thank you to the development team for WEC-Sim! It's
been really handy to have something off the shelf that is open source, so
that we can get in and tinker.

Cheers,
Cameron

bradling:
Hi Cameron,

Yi-Hsiang and I exchanged an email each on this that for some reason did not make it on to the thread. The takeaway is the reason that the WEC-Sim team does not use the diffraction force from BEM + exact FK force calculated at runtime is they still support wamit 6X versions, which apparently doesn't output the diffraction coefficients.

I don't see your sketch, but I think I can imagine what you are talking about. And I still disagree about calculating the linear portion of the FK at runtime at the instantaneous x,y position. The nonlinear portion of the FK force adjusts the force to to horizontal displacement as well as vertical displacement and instantaneous water height.

Currently, WEC-SIm implements the following: (sorry for my sloppy notation, hopefully you can follow)

Fe(t) = Diff( BEM,x=cgx, y = cgy) + FK( BEM, x=cgx, y = cgy) + FK(exact, x=x(t), y = y(t)) - FK(linear STL, x=cgx, y=cgy)

If we assume FK(BEM) = FK(linear STL), then we get:

Fe(t) = Diff(BEM, x=cgx, y = cgy) + FK(exact, x=x(t), y = y(t)).

However, if at runtime the linear excitation force is calculated with the instantaneous horizontal position, we will get:

 Fe(t) = Diff( BEM,x=cgx, y = cgy) + FK( BEM, x=cgx, y = cgy) + FK(exact, x=x(t), y = y(t)) - FK(linear STL, x=x(t), y=y(t))

Since FK( BEM, x=cgx, y=cgy) != FK(linear STL, x=x(t), y=y(t)), this term could be considered an error term.

I think in your case where the body is displaced by half a wavelength, the assumption that we can combine the linear scattering force and nonlinear FK force is not really valid. I think the errors come from this assumption breaking down, not the calculation of the linear portion of the FK force. Maybe we can get @yuyihsiang to chime in with his take on this?

cmcnatt:
Hi Brad,

I think in reality, what is most correct and should be implemented is:

Fe(t) = Diff( BEM,x=x(t), y=y(t)) + FK( BEM, x=x(t), y=y(t)) + FK(exact,
x=x(t), y = y(t)) - FK(linear STL, x=x(t), y=y(t)),

which is of course:

Fe(t) = Diff( BEM,x=x(t), y=y(t)) + FK(exact, x=x(t), y = y(t))

I understand your need to be mathematically consistent, but the way it was
implemented blew up. The nonlinear FK theory isn't mathematically
consistent with linear theory anyways. It's a "blended method", which is
really designed to "work and be fast", at least that's my perspective.

bradling:
I would agree that calculating BEM excitation force at the instantaneous position is probably best, but if the BEM excitation force is calculated based on (0,0,0), then I believe the current implementation is most consistent. My approach is to simply not trust simulations with large horizontal displacements relative to wavelength, although that may be more problematic for other designs.

I previously had some poor simulation results due to bemio incorrectly interpreting nemoh results. Bemio did not previously account for the opposite excitation phase sign convention as wamit and aqwa  (https://github.com/WEC-Sim/bemio/commit/daf06845d2ba61eae6accba39b5ae5424ed02ee2), and this incorrect excitation phase combined with the nonlinear hydro calculations created erroneous results in some situations.

It would be nice to have some test data that exhibits some of this behavior, and the different approaches could be compared to reality!

yuyihsiang:
Hi Cameron and Brad

If I understand both of you correctly, I agree more with Brad on his comment on (2) . Maybe let me start from the beginning.

(1) First of all, the option of simu.nlHydro = 3 for nonlinear hydrostatic force but a linear FK sounds wired to me. Cameron, it will be great if you can clarify what nonlinear hydrostatic force means here. Are you talking about nonlinear effect from just the movement of the body position or including the wave elevation? If it is just the movement of the body, simu.nlHydro=1 gives what you want. If it is the later one, where wave elevation is considered in one term but not the other one seems problematic.

(2) I agree 

Fe(t) = Diff( BEM,x=x(t), y=y(t)) + FK( BEM, x=x(t), y=y(t)) + FK(exact, x=x(t), y = y(t)) - FK(linear STL, x=x(t), y=y(t)),

But keep in mind that the FK( BEM, x=x(t), y=y(t)) is obtained from BEM and it assumes that the body is fixed (at the original location). Therefore, in order for Fe(t) to be correct, FK(linear STL, x=x(t), y=y(t)) has to be calculated at the origin, otherwise it is not correct mathematically and may introduce more errors than using just linear theory in my opinion. 

I agree that the nonlinear FK theory isn't mathematically consistent with linear theory anyways. But what do you mean that the way it was implemented blew up, which worries me.

Thanks,

--Yi-Hsiang

cmcnatt:
Hi guys,

To continue discussion...

My nlHydro = 3 option is for nonlinear hydrostatic for the exact wetted
geometry using z = 0 as the reference and using the linear (from BEM)
excitation force. It was my understanding that nlHydro = 1 was my version
for hydrostatic, but used the body exact FK with z=0 as the reference.
(While nlHydro = 2 is exact free surface with Weber stretching).

I wanted to look at this option because it separates the effect of
hydrostatic force from the excitation force. They are not necessarily
connected except in the case of Weber stretching. I also have implemented
(although I'm not sure it's working correctly) a totally different
nonlinear option that uses BEM coefficients of different submergences of a
given geometry.  And this requires using just nonlinear hydrostatics and
not the nonlinear FK as implemented by nlHydro = 1.

Back to the other point...

I understand your all's perspective. You are implementing: (linear
diffraction force) + (body exact FK):

Fe(t) = Diff(BEM, x=cgx, y = cgy) + FK(exact, x=x(t), y = y(t))

What I am implementing is slightly different: (linear excitation force) +
(nonlinear portion of body exact FK):

Fe(t) = Fe(BEM, x=cgx, y=cgy) + [FK(exact, x=x(t), y = y(t)) - FK(linear,
x=x(t), y = y(t))]

The term inside the square brackets is the additional force due to a change
in wetted geometry in heave, pitch, roll as well as considering the
pressure based on the exact free surface.

The case I am running is hinged raft (the Mocean WEC: www.moceanenergy.com).
With nlHydro=2, there is a mean drift in surge from the origin, which we
observed in physical tests, and the WEC-Sim result seems plausible. We
cannot count on body oscillating about the origin.  I am not saying what I
have implemented above is correct, and I think I will implement:

Fe(t) = Diff(BEM, x=x(t), y = y(t)) + FK(exact, x=x(t), y = y(t)).

using the phasor I described in a previous email. But I think what I
implemented doesn't blow up, while the original implementation does, is
because the term in [ ] is much smaller in magnitude than  FK(exact,
x=x(t), y = y(t))

cmcnatt:
I acidentally hit send there... but I just wanted to add that I don't want
to belabor the point. You are of course under no obligation to merge what I
implemented.

It sounds like there was an update to the FK calc that I basically
overwrote, because of the challenges of merging Simulink Libraries.

Keep me posted on if/how you want to integrate my changes, and I'm happy to
help when necessary.

Cheers,
Cameron

yuyihsiang:
Hi Cameron

You have done several different changes, and we really appreciate that. We never expected one person would ask for pull request on multiple things at once, which make the pull request more complicate. Therefore, I have a request if that is possible. It will be great if you can re-submit the pull request and separate the changes 3~5 from 1~2.

Regarding the pull request number 1 and 2, I think I have a better understanding on what you did. But it will great if we can have a quick chat to make sure we have a complete understanding. Also, I am worry that the Fe you want to calculate will contain a large error if the body motion is large. 
  
I will be available between 9:00~10:00 and 11:00~13:00 tomorrow. Anytime next Monday will work for me. Let me know what time works for you.

Thanks,

--Yi-Hsiang

cmcnatt:
Hi Yi-Hsiang,

I am happy to resubmit the pull requests. I can't do it right away as I
have a lot on my plate right now, but hopefully in a week or so.

I am happy to give you a call today. Could you send me your number at:
cameron.mcnatt@moceanenergy.com?

Thanks,
Cameron

PR #133, opened 2016-10-26
User defined frequency range for irregular wave simulations.

bradling:
Created an option to allow users to define the range of frequencies used to build irregular wave spectra in irregular wave simulations.

I included some error checking, and default to the frequency bounds in the BEM data if invalid inputs are provided.


kmruehl:
@bradling This revision has been added. I manually moved the waveClass.m files over, so that I would not overwrite the the MCR revisions. See [this commit.](https://github.com/WEC-Sim/WEC-Sim/commit/820d055832f2b1e789a11e8d83314681dd9f2ca4)

PR #137, opened 2016-11-01
NL Hydro Addl Option - Improve Runtime

bradling:
This pull requests adds the additional option to set `simu.nlHydro = 3`. This option calculates the full FK force at each time step using the STL file, and adds that to the linear diffraction force (as previously discussed in #126 thread).  This approach gives the same results as setting `simu.nlHydro = 2`, but eliminates the need to calculate and remove the linear portion of the FK force at each time step using the STL file.

I did some testing in regular waves and a single irregular wave class, and saw very close agreement. I can upload the test cases if that would be helpful. The change in runtime was small in regular waves, but more significant in irregular waves (~40% reduction in runtime). Position and excitation forces were nearly identical, with differences small enough to be attributed to mesh discretization error.

Also included in the pull request is a modification to BEMIO to read in diffraction coefficients from NEMOH results. I did not modify the AQWA or WAMIT read modules, as I do not have either of those software packages. Diffraction force coefficients in the H5 are necessary to implement `simu.nlHydro = 3`.


bradling:
Cool. Let me know if you need me to make any additional changes.

j-vanrij:
OK, so after going through your code changes, we’ve decided that we like the concept, and want to incorporate the nonlinear option you’ve suggested. But, if we implement this for Nemoh, we’d want to do the same for WAMIT and AQWA, so as to avoid any BEM specific code (in both BEMIO and WEC-Sim). With this in mind, and all of the other recent BEMIO changes, we’ve decided it would be easier to just make the nonlinear BEMIO/WEC-Sim changes in the current master code, rather than try to pull your code in. So, I’m going to close the request, but did want to let you know that your suggestions WILL be incorporated shortly – thanks for all of your work!

bradling:
That sounds fine to me. If there is anything additional I could do to help with this effort, let me know. I would be glad to help.

Issue #145, opened 2016-12-01
Potential MCR Speed Improvement - Changing Hydro Data?


bradling:
Hello WEC-Sim Community,

This is more of a question about a potential improvement than an issue. I often use wecSimMCR to sweep through various regular wave cases. I have noticed that a significant portion of the run time in regular waves is dedicated to loading the h5 file for each instance (~20 seconds of the 30 second total run time per case). It would be possible to load the .h5 file only once when using wecSimMCR, and that 20 seconds of run time could be eliminated for each additional case tested.

So my question is this:

Does anyone use wecSimMCR to change the hydrodynamic file being used?

If that is not a common use case, then it would be easy to modify wecSimMCR to load the h5 file once. Another option would be to include a flag in the simu object that tells wecSimMCR whether or not it needs to reread the h5 file each time.

If this is something the WEC Sim team is interested in including in the code I would be happy to make a pull request including this improvement. However, if this is something the development team is not interested in including, I don't want to waste my time making a pull request.

kmruehl:
@bradling I think this is a great suggestion, and as you mentioned, it would reduce the run time for MCR runs a lot. If this is a change you are interesting in making, I recommend you make a pull request to do so. I would also suggest that you make it a simu flag so that it is an option that users can toggle on/off,  with the default such that the h5 file is only loaded once. This would allow flexibility as there may be instances when a user would want to load it between runs (the only example I can think of is if the user has different BEM runs they want to compare). Also, you'll need to check what variables are being cleared between MCR runs, and verify that the hydrodynamics properties are stored between runs (but nothing else from the previous run).

bradling:
OK I will work a little bit on this when I have time, and then make a pull request, keeping in mind your suggestions when implementing this.

kmruehl:
@bradling thank you

Toby-D:
@bradling: Good call on not reloading the .h5 data. 

It's also worth noting that most (~90%) of the CPU time associated with loading the .h5 data is actually spent checking the MATLAB version. 

Lines 59-62 of h5load.m address a compatibility issue with versions from 2008 and earlier. Using MATLAB's profile tool on a representative run, it looks like the current software version gets checked 379 times while loading the .h5 data, which adds up to 8.6 s of the 9.6 s it took to load the data. 

So if you're not planning on going back to R2008 anytime soon, you can replace those lines with:
%   v = ver('MATLAB');
[...]

PR #146, opened 2016-12-05
Store hydroData between runs for MCR - Improved MCR runtime

bradling:
Speed improvement for MCR:
* hydroData is stored in memory between uses to eliminate redundant h5 file read operations.
* Significantly improves overall MCR runtime when each case has a short runtime.
* if h5File is changed between runs, then hydroData is always reloaded from h5 file.

I have two test cases I can provide if requested.

See Issue https://github.com/WEC-Sim/WEC-Sim/issues/145 

@kmruehl

### DIFF
--- a/source/functions/wecSimMCR.m
+++ b/source/functions/wecSimMCR.m
@@ -64,9 +64,24 @@ else
     end
 end
 
+% Check to see if changing h5 file between runs
+% If one of the MCR headers is body(#).h5File, then the hydro data will be
+% loaded from the h5 file for each condition run. 
+reloadHydroDataFlag = true;
+if isempty(cell2mat(regexp(mcr.header, 'body\(\d+\).h5File')))
+    reloadHydroDataFlag = false;
+    clear hydroData
+end
+
 % Run WEC-Sim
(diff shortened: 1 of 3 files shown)


kmruehl:
@bradling thanks for doing this. This code modification speeds up the mcr case runs _a lot_. I made a minor revision, so that instead of the flag being turned on in the mcr case file, it is instead done in the simulation class, under `sim.reloadH5Data`. Thanks again!

bradling:
Hi Kelley, glad you like the changes. 

I am a little concerned that placing the flag in the simu object could lead to a case where a user sets up an MCR case that changes the h5 file between runs, and either doesn't realize or forgets to change to set `simu.realoadH5Data = false`. That is why I chose to scan the mcr header to check if the h5 file was being changed between runs to set the flag. Ultimately it is your decision, but I want to make sure my concern is voiced. @kmruehl

kmruehl:
@bradling Thank you for your feedback. That's a good point, but I wanted to keep this feature addition consistent with other code features, where the numerical flags are toggled on/off within the class definition as opposed to modified within the executable (`wecSim.m` or `wecSimMCR.m`) directly. I'll go ahead and add a note to the MCR documentation about how/when the hydro data is read. That should (hopefully) help.

Issue #158, opened 2017-02-09
Porting to plain Matlab: What does B2B block do exactly?


crobarcro:
Hello,

I'm porting WEC-Sim to use only plain matlab for the calculation of forces, and using MBDyn for the multibody dynamics. You can see the current status of this work [here](https://sourceforge.net/p/rnfoundry/hgrepo/ci/default/tree/wave/matlab-octave/wec-sim/). I'm essentially extending the bodyClass to also include methods for calculation of the dynamic forces. I copied the bodyClass to a new class hydrobody which has the extra methods. I've also had to tweak the simulationClass for this which isn't uploaded anywhere yet.

I am working my way through making sure the force results match WEC-Sim (I'm using the RM3 example as a reference at this stage) but am having some trouble with the added mass forces, and am hoping you could clarify a couple of things for me. 

In my Matlab code, to calculate the added mass I have the following:

`F_addedmass = body.hydroForce.fAddedMass * accel(:);`

I haven't yet applied the equivalent of the transport delay.

where `accel` is a (6 x n) matrix (when doing body to body interactions), and therefore `accel(:)` is a (6*n x 1) column vector, which appears to be the right format. The issue is that I cannot get the right forces with this. When I compare the forces ffrom simulink (before post-processing) the forces for the first body are correct (match the original WEC-Sim output) but for body 2 are very slightly different. 

If i then do the post-processing like:

```
for iBod = 1:simu.numWecBodies
    body(iBod).restoreMassMatrix
    output.bodies(iBod).forceTotal = output.bodies(iBod).forceTotal + output.bodies(iBod).forceAddedMass;
    output.bodies(iBod).forceAddedMass = body(iBod).forceAddedMass(output.bodies(iBod).acceleration,simu.b2b);
    output.bodies(iBod).forceTotal = output.bodies(iBod).forceTotal - output.bodies(iBod).forceAddedMass;
end; clear iBod
```
The forces are then all wrong (scaled up too much). Is there anything obvious I might be missing here?

My other question is, what is the purpose of the transport delay on the accelerations for the added mass calc?

kmruehl:
`simu.b2b` turns on/off [body2body interactions](http://wec-sim.github.io/WEC-Sim/features.html#body-to-body-interactions). If body2body interactions are on, the dimension of the velocity and acceleration terms change, and thus the matrix definition of added mass and damping changes as well.  Refer to the above website link for more info on b2b.

[...]

PR #160, opened 2017-02-10
simulationClass tweaks

crobarcro:
Hi,

I tweaked simulationClass.m so you don't necessarily have to start in the same directory as the case directory when launching a simulation. It's backward compatible with the previous behavior (used `pwd()`) if no case dir supplied. 

I also added more checking for slx files existence, and tweaked how some file names are constructed. I've also added the possibility of supporting MBDyn in future as an alternative to SimMechanics by adding a 'multibodySolver' property.

I won't be offended if you don't like these changes, but would rather have them in WEC-Sim than maintain a fork.


kmruehl:
#158

yuyihsiang:
Hi crobarcro

Sorry for taking so long o reply. We are very interested in added the possibility of supporting MBDyn in future as an alternative to SimMechanics. We have spent some time look into MBDyn and it looks very promising and can be used not only for WEC-Sim type of simulations but also CFD simulations, as far as I understand.

However, as you pointed out, this requires some changes to the WEC-Sim source code. We would not merge it for now, but I'll create a project for it and hopefully we will be able to further investigate this more down the road.

Thanks,

-Yi-Hsiang

crobarcro:
Thanks, actually I've already successfully swapped simmechanics with mbdyn for multibody dynamics and am just refining the system. In fact no simulink is required at all. It currently only works on Linux, but I am working on (cross)compiling mbdyn for windows and adding winsock support (communication between matlab and mbdyn is done via sockets). This involved implementing everything in plain Matlab. 

You can see a video of the RM3 simulated using the system here:

https://www.youtube.com/watch?v=ZHil5qh4rGA

This work is shared online here:

https://sourceforge.net/p/rnfoundry/hgrepo/ci/default/tree/wave/matlab-octave/wec-sim/

the mbdyn stuff is here:

https://sourceforge.net/p/rnfoundry/hgrepo/ci/default/tree/common/multibody/

but really you need parts of the whole code base RenewNet Foundry to make it work.

however there's not much documentation yet so it would be a struggle for anyone to pick up. Only the convolution algorithm method is tested and working yet for radiation forces, but I will begin testing the state space method soon.

I would gladly accept help on this if it is of interest.

yuyihsiang:
Hi Richard 

That looks really cool. I'll see if we can find someone from our end to dig into this further.
Let me know how the state space method works. 

Thanks,

-Yi-Hsiang

shadigoodarzi0048:
Hello
I'm forming geometry. When I execute, I get this error. Can someone help me?

The variant control 'sv_b_hydroBody' of the variant block 'RM31/Rigid Body' must be a condition expression or name of a Simulink.Variant object in the global scope workspace of the model.
Component:Simulink | Category:Model error

Issue #159, opened 2017-02-10
Morison Element state derivative 0 at time = 0.0


bradling:
Hello,

I am getting an error when trying to use morison drag terms in an irregular wave simulation. If I turn off morison drag then the error does not occur. The error message is:

`Error using wecSim (line 248)
Error occurred while simulating the model 'single_float_model' in rapid
accelerator mode. Rerun the simulation in normal mode to diagnose this
error. Continuous state derivative of 'element 3' is non-finite`

I spent some time tracing signals in the morison drag calculation block. I believe it to be the same error as was happening in https://github.com/WEC-Sim/WEC-Sim/issues/84, but its occurring in the Morison drag orbital velocity calculation. I am under a bit of a time crunch to make get this working, so I am going to rewrite the orbital velocity function in matlab code instead of simulink, and use the deep-water formula for large wavenumbers.

bradling:
See pull request https://github.com/WEC-Sim/WEC-Sim/pull/164. This seems to fix the problem.

I do have an additional question though. Why is the irregular wave amplitude (`waves.A`) multiplied by a factor of 2 when input to the orbital_velocity calculation:

![image](https://cloud.githubusercontent.com/assets/10558137/23191576/336b68c4-f853-11e6-9eed-7b4bc3efdef1.png)

This same factor does not exist in other calculations using `waves.A`, for example in the `waveClass` method `waveElevIrreg`, we have:

`tmp=sqrt(obj.A.*df);`
`tmp1 = tmp.*real(exp(sqrt(-1).*(obj.w.*t + obj.phaseRand)));`
`obj.waveAmpTime(i,1) = t;`
`obj.waveAmpTime(i,2) = sum(tmp1);`, 

and in the Nonlinear F-K calculation we have (where `AH = waves.A`):

`z=(center(:,3)-elv).*wDepth./(wDepth+elv);`
`f_tmp = rho.*g.*sqrt(AH(i)*dw).*cos(k(i).*center(:,1)-w(i)*t-phaseRand(i));`
`f = f + f_tmp.*(cosh(k(i).*(z+wDepth))./cosh(k(i).*wDepth));`

Is this difference due to how the orbital velocity calcuation implemented, or is it an error?

bradling:
@yuyihsiang  Any update on this? I see you approved the pull request that changes the calculation of orbital velocity to use deepwater wave approximation. But what about the term `2*A` discussed above?

yuyihsiang:
[...]

PR #165, opened 2017-03-10
Fix Morison Orbital Vel Connections

bradling:
Fixes the connections between orbital velocity function and subsystem
outputs.

Reference Issue https://github.com/WEC-Sim/WEC-Sim/issues/159


bradling:
Any update on the status of this? It is a minor change, and the Morison Drag calculation is incorrect without this fix.

kmruehl:
@bradling thank you for this pull request

PR #166, opened 2017-03-10
add missing fclose() to responseClass

bradling:
when reading moorDyn results, "Lines.out" is not closed.


bradling:
Any update? This is a minor change, and can cause weird and annoying memory issues if many cases are run with moordyn.

PR #189, opened 2017-07-13
Update bodyClass.m

gparisella:
The change is related to 'user-defined' wave elevation cases. 

When 'length(obj.hydroData.simulation_parameters.wave_dir) == 1' BUT differs from the default waveDir=0 (elseif condition, line 367).
In those cases, the run would originally break since obj.userDefinedExcIRF is not produced.

Therefore, specifying the correct value in the wecSimInputFile.m fixes the problem.


kmruehl:
@gparisella Thanks for this!

PR #240, opened 2018-03-23
Froude krylov and Diffraction body tag fix

serag-abdulmotaleb:
The Read_AQWA function gives a matrix index error (the index is not a positive integer) when I try to model an array with more than 9 bodies. I found that the problem is with the 156th and 178th lines in the Read_AQWA function since they take only the (end-8)th character of the string which will be equal to 0 for structure number 10 and 1 for structure number 11 and so forth. This can be easily fixed by replacing (end-8) with (end-9:end-8) so as to capture the tens digit.


nathanmtom:
@serag-abdulmotaleb Thank you for bringing this to the attention of WEC-Sim developers. We have reviewed your code modifications and believe you are correct in how to fix the bug in number the bodies in AQWA with over 9 bodies. We have incorporated your code changes into WEC-Sim. Again, thank you again for your help improving the WEC-Sim code.

PR #259, opened 2018-07-23
Directional waves

emiliofa:
Changes of the linear wave excitation force in the MATLAB function of the Wec-Sim model to account the wave directions. Including the reading of the hydrodynamic excitation coefficients for all the directions of the hydrodata file in bodyclass. Including a wave spread and random phase generation for all the wave directions.

### DIFF
--- a/source/lib/WEC-Sim/WECSim_Lib.mdl
+++ b/source/lib/WEC-Sim/WECSim_Lib.mdl
@@ -1,7 +1,12 @@
 Library {
-  Name			  "WECSim_Lib_update"
-  Version		  8.6
+  Name			  "WECSim_Lib"
+  Version		  8.8
   SavedCharacterEncoding  "windows-1252"
+  LogicAnalyzerGraphicalSettings ""
+  LogicAnalyzerPlugin	  "on"
+  LogicAnalyzerSignalOrdering ""
+  DiagnosticSuppressor	  "on"
+  SuppressorTable	  "22 serialization::archive 11 0 3 0 0 0 8 0"
   LibraryType		  "BlockLibrary"
   ScopeRefreshTime	  0.035000
   OverrideScopeRefreshTime on
@@ -19,6 +24,7 @@ Library {
   "ional PTO Actuation Torque||__slOldName__||WECSim_Lib/PTOs/Rotational PTO Prescribed Motion (Local RY)||__slNewName_"
   "_||WECSim_Lib/PTOs/Rotational PTO Actuation Motion"
   MaxMDLFileLineLength	  120
+  LastSavedArchitecture	  "win64"
   Object {
     $PropName		    "BdWindowsInfo"
     $ObjectID		    1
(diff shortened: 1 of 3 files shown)


yuyihsiang:
Hi emiliofa

Thank you for providing the directional wave functionality. It look good to me and I have merged it to the official release. If I understand correctly, the one thing that is needed when using this is the waves.waveSpread vector. Is that correct?  In addition, it will be very helpful if you can also provide an example if you happen to have one.

Best,

Yi-Hsiang

emiliofa:
Hi Yi-Hsiang,

yes it's correct. I can send one of the tutorial case (OSWEC) with this functionality. About the theory of the spreading function can be found in http://rspa.royalsocietypublishing.org/content/465/2111/3361.
I attach the case here:
[OSWEC.zip](https://github.com/WEC-Sim/WEC-Sim/files/2235476/OSWEC.zip)
It will be interesting also to do some validation case. Please let me know for some opportunity.

Thanks,
Emilio

yuyihsiang:
Thanks you, that will be great,

Yi-Hsiang

jo4c:
Hi Emilio et al,

This looks like a great addition to WEC-Sim.  Is there a way to export the wave surface to paraview in this case?  It looks like this would involve a re-write of the waveElevationGrid function in waveClass, but I don't really have a feel for how involved this would be...

Kind regards,
Jo

emiliofa:
Dear Jo,

thanks for your interest in this. I have attached an example of waveclass updated for the Paraview visualization for this case. Please see the lines 293-300.
[waveClass.zip](https://github.com/WEC-Sim/WEC-Sim/files/2681334/waveClass.zip)

It is just my modification and not checked yet by the WEC-Sim team. So consider it carefully please.

Kind regards,
Emilio

jo4c:
Hi Emilio,

This appears to be working well.  I've also had it working well using a directional distribution function to create spread sea states similar to ones we have used in the tank at FloWave - some validation required of course, but from a visual point of view it all looks very plausible so far.

A quick question:  am I correct in thinking that, in it's current implementation, all directional components will have the same Tp?  i.e. you couldn't currently use this to model, for example, a long-wavelength sea at 0 degrees, with a smaller short-wavelength component at 90 degrees?

Kind regards,
Jo

emiliofa:
Dear Jo,

yes is correct. In your case you will have a directional distribution that depends also on the frequency. 
The case that I considered is simplified with a directional distribution dependent only on the direction but you can find a more general case for example in the attachment. It will be really interesting to further expand this topic. Let me know if you need anything from my side, will be great if i can do something to help. 
[1520-0485%281998%29028%3C0495%3Aootdso%3E2.0.co%3B2.pdf](https://github.com/WEC-Sim/WEC-Sim/files/2690500/1520-0485.281998.29028.3C0495.3Aootdso.3E2.0.co.3B2.pdf)

Kind regards,
Emilio

PR #292, opened 2019-04-11
Update responseClass.m - Add missing fclose(fid)

bradling:
When reading moorDyn outputs, each "LineX.out" file, the file is left open when reading. This can cause matlab to crash when repeatedly running models with moorDyn


kmruehl:
@bradling Thank you for making this revision.

PR #408, opened 2020-08-28
Morison

lilnguy:
Update Morison Element block to decompose the fluid and body motion to tangential and normal components to consider the magnitude in the calculation for the Morison force.


nathanmtom:
@lilnguy Thank you for this pull request! The changes you made to the Body and Response classes look good. Unfortunately, I cannot open your WECSim library block as it has been saved as a Matlab 2020 version. We save the WECSim library block as a 2015b version to allow use with past versions of Matlab. If possible can you export the library block as a 2015b version and add to the pull request for review?

kmruehl:
@lilnguy I just created a [dev](https://github.com/WEC-Sim/WEC-Sim/tree/dev) branch and modified this PR so that it submits change to 'dev' instead of 'master'

kmruehl:
@lilnguy  Can you let @nathanmtom push to your fork so that he can make revisions too? [see info here](https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork). @nathanmtom this gives you the option to make minor changes to the PR, like saving the `*.slx` file to 2015b.

kmruehl:
@lilnguy my apologies! I misspoke, I thought this PR was from your fork, not from a branch on the main repository. @nathanmtom should be able to save the `*.slx` file to a previous version of MATLAB by pushing directly to the `Morison` branch.

nathanmtom:
@kmruehl and @yuyihsiang I've realized that I need to make an update to the WEC-Sim library. So wait until the next push before taking a deeper dive into the Morison Element update.

nathanmtom:
@kmruehl @yuyihsiang I've updated the WECSim_Lib.slx to fix the broken links in the Flex Body block. With this last commit, the update is now ready for review.

kmruehl:
@nathanmtom I get an error when I run `simu.morisonElement    = 1; ` with the Free_Decay 1m-ME application case. Do you also get this error?

```matlab
Error using wecSim (line 316)
Derivative of state 'sphere.Floating_3DOF.CONSTRAINT_Planar_Joint.Px.v' in block
'sphere/Floating (3DOF)/CONSTRAINT (Planar Joint)' at time 0.0 is not finite. The simulation
will be stopped. There may be a singularity in the solution.  If not, try reducing the step
size (either by reducing the fixed step size or by tightening the error tolerances)
```

nathanmtom:
@kmruehl Yes, because the application case hasn't been updated to include the following:

body(1).morisonElement.z    = [0 1 0];

with 

simu.morisonElement = 1; 

without defining the .z vector to be a normal vector, not [0 0 0] which is the default initialization, there will be a situation where there will be a division by 0 and thus a singularity in the solution. 

I'll see if I can have the code check the variable is non zero and is a unit normal.

I did edit the applications case to show how you need to run both option 1 and option 2, but haven't pushed it since this pull request is not yet incorporated into master branch.

nathanmtom:
@kmruehl @yuyihsiang  The updates to the Morison Element calculation which has set simu.morisonElement = 1 option to the original formulation have been completed. The WEC-Sim_Applications Free Decay example now works with backwards compatibility.

kmruehl:
@nathanmtom thanks for adding the warning glad related to not defining `body(1).morisonElement.z = [0 1 0];`

nathanmtom:
@kmruehl The update library with the corrected Flex Body block links has been uploaded to the Morison branch.

Issue #384, opened 2020-07-08
Morrison Body


ffrancesco:
This is an improvement request and not an actual issue.
I believe it might be interesting to create a third body type, along with the Hydro and non-Hydro, which includes Morrison forces but does not require the results from BEM solver.

dforbush2:
This is in progress!

kmruehl:
@ffrancesco this should be resolved with @dforbush2 pull-request #423

PR #423, opened 2020-10-07
Drag body implementation, set nhBody = 2

dforbush2:
This is to address issue #384 and creates an additional dragBodyPre method of bodyClass, and rigid body library block, and the necessary flags in simulationClass. Set body(num).nhBody = 2 in WEC-Sim input file to utilize. Acts like non-hydro body but allows for linear and viscous damping forces to be calculated.


kmruehl:
Thanks @dforbush2!

kmruehl:
@dforbush2 I just created a [dev](https://github.com/WEC-Sim/WEC-Sim/tree/dev) branch and changed this PR so that it submits change to 'dev' instead of 'master'

nathanmtom:
@dforbush2 Thanks for pulling this pull request together. If possible, can you save the updated WEC-Sim Library to a Matlab 2015b version? Currently it appears the library is saved as a 2020a version.

Cheers, 
Nathan

source/objects/bodyClass.m
```diff
@@ -183,6 +183,21 @@ function loadHydroData(obj, hydroData)
             obj.dof_gbm   = obj.dof-6;
         end
         
+        function dragForcePre(obj,rho)
```
nathanmtom, line 186: @dforbush2 This implementation allows for a non-hydrodynamic body to include a linear and a quadratic viscous drag forcing term; however, based on the conversation on issue #384 the requester is asking for the Morison Element forcing which includes a drag and added inertia coefficient that also includes the flow velocity and acceleration of the undisturbed wave velocity. Therefore, this modification may not be fully addressing the issue; however, it should be relatively straightforward to adapt for the Morison Element.

nathanmtom: @dforbush2 My apologies, believe I misunderstood the implementation. Please disregard my previous comment.

dforbush2: I did not need to modify anything to specifically incorporate the morison elements, but I did test the new implementation with morison elements on/off and they seem to be behaving correctly.

nathanmtom: @dforbush2 Yes, I realized that the linear and quadratic damping are attached to the body.hydroforce variable, while all of the Morison Element coefficients stay within the body object and is not placed within body.hydroforce. Perhaps the Morison Element definition should also be moved into body.hydroforce to be consistent with the rest of the hydrodynamic forces.

nathanmtom (approved):
@dforbush2 Your pull request looks good, apologies for the prior misunderstanding on my part.

PR #425, opened 2020-10-08
Speed wave gauge calculation

akeeste:
Created a new variable that holds a 3x1 array of booleans to track if each wave gauge has been placed by the user. Wrapped wave gauge amplitude calculations in if statements to speed the pre-processing calculations. If a wave gauge location is at the default of 0,0 it is not updated.

The etaImport and noWave cases are left alone so that they still preallocate an array of zeroes for the amplitude at each gauge. This matches the behavior of the other wave cases when a gauge is not placed.


source/objects/waveClass.m
```diff
@@ -65,6 +65,7 @@
         waveAmpTime1 = [];  % Wave elevation time history at a wave gauge 1 location specified by user [m] 
         waveAmpTime2 = [];  % Wave elevation time history at a wave gauge 2 location specified by user [m] 
         waveAmpTime3 = [];  % Wave elevation time history at a wave gauge 3 location specified by user [m] 
+        waveGaugeExist = [false, false, false]; % Boolean array to track if wave gauges have been specified (loc ~= 0,0)
         A = [];             % Wave amplitude [m]. For regular waves or 2*(wave spectrum vector) for irregular waves
```
kmruehl, line 69: Instead of creating a new property within the waveClass I recommend checking to see if each if any of the wave gauges have been initialized, for example`if true(waves.wavegauge1loc ~= [nan,nan])`. This removes the middle step of defining a new property, and helps keep the class a little cleaner/simpler.

source/objects/waveClass.m
```diff
@@ -551,7 +555,7 @@ function setWaveProps(obj,wDepth)
         end
         
         function waveElevNowave(obj,maxIt,dt)
-            % Set noWave levation time-history
+            % Set noWave elevation time-history
             obj.waveAmpTime         = zeros(maxIt+1,2);
```
kmruehl, line 559: Thanks for catching a typo!

source/objects/waveClass.m
```diff
@@ -188,6 +189,9 @@ function waveSetup(obj,bemFreq,wDepth,rampTime,dt,maxIt,g, rho, endTime)
             %
             obj.bemFreq    = bemFreq;
             obj.setWaveProps(wDepth)
+            obj.waveGaugeExist(1) = boolean(max(obj.wavegauge1loc ~= [0,0]));  % Check if user has set wave gauge 1
+            obj.waveGaugeExist(2) = boolean(max(obj.wavegauge2loc ~= [0,0]));  % Check if user has set wave gauge 2
+            obj.waveGaugeExist(3) = boolean(max(obj.wavegauge3loc ~= [0,0]));  % Check if user has set wave gauge 3
             switch obj.type
```
kmruehl, line 195: See above comment, I don't think this logic is necessary

source/objects/waveClass.m
kmruehl: Also, I think we should change the way the wave gauge locations are initialized. Instead of setting them to `[0,0]`, I think we should set them to `wavegauge1loc = [nan,nan];`

source/objects/waveClass.m
```diff
@@ -575,35 +579,53 @@ function waveElevReg(obj, rampTime,dt,maxIt)
...
                     obj.waveAmpTime(i,1)    = t;
                     obj.waveAmpTime(i,2)    = obj.A*cos(obj.w*t);
-                    obj.waveAmpTime1(i,1)   = t;
-                    obj.waveAmpTime1(i,2)   = obj.A*cos(obj.w*t-obj.k*(obj.wavegauge1loc(1).*cos(obj.waveDir*pi/180) + obj.wavegauge1loc(2).*sin(obj.waveDir*pi/180)));
-                    obj.waveAmpTime2(i,1)   = t;
-                    obj.waveAmpTime2(i,2)   = obj.A*cos(obj.w*t-obj.k*(obj.wavegauge2loc(1).*cos(obj.waveDir*pi/180) + obj.wavegauge2loc(2).*sin(obj.waveDir*pi/180)));
-                    obj.waveAmpTime3(i,1)   = t;
-                    obj.waveAmpTime3(i,2)   = obj.A*cos(obj.w*t-obj.k*(obj.wavegauge3loc(1).*cos(obj.waveDir*pi/180) + obj.wavegauge3loc(2).*sin(obj.waveDir*pi/180)));
+                    if obj.waveGaugeExist(1)
+                        obj.waveAmpTime1(i,1)   = t;
```
kmruehl, line 583: replace with `if true(waves.wavegauge1loc ~= [nan,nan])`

kmruehl (changes requested):
@akeeste Thanks for getting started on this PR! 

I have a few recommended revisions, and I've already made some of these changes locally. **Can you let me to push to your fork so that I can share my revisions with you?** 

I've made specific comments below, but generally I think this could be a bit cleaner. With these revisions, the `waveClass` is still generating wave gauge output of zeros for all three gauges, e.g. `waveswaveAmpTime1        = zeros(maxIt+1,2);`. I think this logic can be cleaner so that we don't generate _any_ wave surface elevation data at a wave gauge unless the user requests it by specifying `waves.wavegauge1loc ~= [nan,nan]`.

kmruehl:
@akeeste I just created a [dev](https://github.com/WEC-Sim/WEC-Sim/tree/dev) branch and modified this PR so that it submits change to 'dev' instead of 'master'

kmruehl:
@akeeste since you've already started making the changes here, and I don't want to over-write them I went ahead and uploaded my suggestions to my own fork https://github.com/kmruehl/WEC-Sim/tree/dev_adam for reference. Let me know if you have any questions!

akeeste:
@kmruehl Thanks for the advice and changes. I'm not sure if you are yet able to push to my fork so I updated it myself with your suggestions.

In the new commit I simplify the conditions for setting the wave gauge amplitudes. Locations now default to empty arrays which are easily checked with the isempty() function. I also remove etaImport from the case statement in line 444 because it prevents the last case from running when etaImport is chosen.

akeeste:
@kmruehl The latest commit updates the default wave gauge location from [] to [NaN, NaN]. The waveClass write_paraview_vtp function is fixed so that waveElevationGrid is called correctly with all arguments. Previously, using Paraview output with the etaImport option gave a wave elevation of 0. The updated function calculates the correct wave height for the Paraview output.

source/objects/waveClass.m
kmruehl: @akeeste your solution is much cleaner, thanks!

kmruehl (approved):
@akeeste this looks great and I'm going to go ahead and merge it into the dev branch. 

My only comment is that when we generate wave gauge data it's only stored to `waves` object, and we should probably also output it to the `output.waves` object. Feel free to make this change to the `responseClass` as a separate PR. For now I'll add it to the project board. 

``` matlab

>> waves

waves = 

  waveClass with properties:

                  type: 'regular'
                     T: 8
                     H: 2.5000
          spectrumType: 'NOT DEFINED'
                 gamma: []
             phaseSeed: 0
      spectrumDataFile: 'NOT DEFINED'
           etaDataFile: 'NOT DEFINED'
             freqRange: []
               numFreq: []
               waveDir: 0
            waveSpread: 1
                   viz: [1×1 struct]
    statisticsDataLoad: []
              freqDisc: 'EqualEnergy'
         wavegauge1loc: [25 0]
         wavegauge2loc: [50 0]
         wavegauge3loc: [75 0]
         currentOption: 3
          currentSpeed: 0
      currentDirection: 0
          currentDepth: 0
               typeNum: 10
               bemFreq: [1×260 double]
            waterDepth: 200
         deepWaterWave: 1
           waveAmpTime: [4001×2 double]
          waveAmpTime1: [4001×2 double]
          waveAmpTime2: [4001×2 double]
          waveAmpTime3: [4001×2 double]
                     A: 1.2500
                     w: 0.7854
                 phase: 0
                    dw: 0
                     k: 0.0629
                     S: []
                    Pw: 4.7864e+04
```

``` matlab
>> output.wave

ans = 

  struct with fields:

         type: 'regular'
         time: [4001×1 double]
    elevation: [4001×1 double]
```

PR #428, opened 2020-10-14
Add wave gauge amplitudes to response class

akeeste:
I add the wave gauge amplitudes to the output responseClass object if the locations are set. This requires changing the responseClass arguments. I lumped the wave type, elevation, wave gauge elevations and wave gauge locations into a waveOutput structure for easier passing. 

I also placed the hydrostatic and FK pressures (hspressure, wpressurenl, wpressurel) into the bodyOutput structure so the responseClass initialization is more clean and easy to follow. The format of the arguments is more consistent this way. It will also be easier to add variables to the responseClass in the future because additional parameters can be inserted into the ___Output structs without changing the responseClass arguments


source/functions/postProcess.m
```diff
@@ -25,6 +25,21 @@
     if iBod == 1; bodiesOutput = body1_out; end
     bodiesOutput(iBod) = eval(['body' num2str(iBod) '_out']);
     eval(['clear body' num2str(iBod) '_out'])
+end
+% Add hydrostatic and FK pressures to bodiesOutput if required.
+for iBod = 1:length(body(1,:))
+    if simu.nlHydro~=0 && body(iBod).nhBody==0 && simu.pressureDis == 1 
+        % hydrostatic pressure
```
akeeste, line 32: Note that this condition doesn't allow the pressure on the body to be output unless the nonlinear hydro option is also on. This is not what the simu.pressureDis option implies in the simulationClass

kmruehl: @akeeste that's a good point. We can only provide the option to output pressure on the body if non-linear hydro is on, so we should add a warning flag letting users knows pressure will not be provided if `simu.nlHydro == 0` & simu.presssureDis ==1`.

kmruehl (changes requested):
@akeeste I ran the WEC-Sim RM3 example with your updates and it creates and unstable soltuion. WEC-Sim should have a simple sinusoidal response for a regular wave input. I'm not sure what's causing the issue, but I recommend running the unit tests with the `runtests` command to compare your updates to the previous version of WEC-Sim. Until the tests pass we cannot merge into dev. 
image

Also, I recommend keeping the PRs as simple as possible, perhaps just add the wave gauge data to the response class output in one pr, and then modify the body output in another.

examples/RM3/wecSimInputFile.m
```diff
@@ -9,6 +9,9 @@
 simu.solver = 'ode4';                   % simu.solver = 'ode4' for fixed step & simu.solver = 'ode45' for variable step 
 simu.dt = 0.1; 							% Simulation time-step [s]
 
+simu.nlHydro = 1;
+simu.pressureDis = 1;
+
```
kmruehl, line 14: @akeeste Okay, I see what's going on now. Please do not modify the input files for the examples, that's what is causing the unexpected response for the RM3. You can change them locally to test your updates, but we don't want to push those changes to the repo. I should be able to resolve this issue by pushing to your branch.

kmruehl (changes requested):
@akeeste I ran the WEC-Sim RM3 example with your updates and it has an unexpected solution. WEC-Sim should have a simple sinusoidal response for a regular wave input. I'm not sure what's causing the issue, but I recommend running the unit tests with the `runtests` command to compare your updates to the previous version of WEC-Sim. Until the tests pass we cannot merge into dev. 
image

Also, I recommend keeping the PRs as simple as possible, perhaps just add the wave gauge data to the response class output in one pr, and then modify the body output in another.

examples/RM3/wecSimInputFile.m
kmruehl: FYI, the tests are passing, so with this minor update to the input file we should be good to merge the PR. 

``` matlab
   25 Passed, 0 Failed, 0 Incomplete.
   373.901 seconds testing time.
```

kmruehl: @akeeste can you let me push to your fork to make these changes?

kmruehl (approved):

kmruehl:
@akeeste take a look at the minor changes, and if you're good with them we can merge this. And can you test the warning flag?

PR #440, opened 2020-10-28
BEMIO Update

nathanmtom:
In the project board under BEMIO there was a request to remove commands within BEMIO.m.

Based on the note, it looked like only clc; close all; clear all was requested to be removed.

Let me know if there was more to this request than described in the project note.


kmruehl:
@nathanmtom can you modify this PR so that it merges into dev, not master? Also, the only file that has been changed is `bemio.m`, correct?

PR #441, opened 2020-10-28
Removing a single line of code from bemio.m

nathanmtom:
As requested, deleting the ``clc;clear all;close all;`` commands from the 20th line of the .m file.


nathanmtom:
@kmruehl Bear with me as I'm trying to ensure I've got the process down to merge with dev rather than master.

kmruehl:
@nathanmtom did you locally compile the documentation? This PR includes a bunch of docs files changes as well: https://github.com/WEC-Sim/WEC-Sim/pull/441/files, and I think it should just be `bemio.m`. This means `git add bemio.m` instead of  `git add all`. Alternatively we could/should update the `.gitignore` files to ignore a local docs compile.

PR #443, opened 2020-10-29
Prefer Static Links to Download Directive in Sphinx Docs

H0R5E:
This commit switches from using the :download: directive to static links for the webinar slides files (in docs/man/webinars.rst). The value of this changes is that the link paths will not be recalculated every time that the docs are built and only a single copy of the files is stored (per branch).


kmruehl:
@H0R5E thank you, merging now

PR #444, opened 2020-10-29
Fix for CI-B2B bug

### LINKED ISSUE #437: Bug when using B2B and CI calculation together
When running the WEC-Sim_Applications repo body-to-body case 4 (B2B and CI calculation). A dimension error is generated in the convolution integral calculation within the radiation force calculation block.

the full error message is: 
Error using wecSim (line 292)
Non-singleton dimensions of the two input arrays must match each other.
	Error in 'RM3/Float/Hydrodynamic Body/Wave Radiation Forces Calculation/SS CI and Constant-Damping-CoeVariant
 Subsystem/Convolution Integral Calculation/MATLAB Function1'.

Within that function, line 21 throws an error because IRKB_interp and velocity are different dimensions:
time_series = bsxfun(@times, IRKB_interp, velocity);

akeeste:
This is a fix and cleanup of the convolution integral function. There is a version bug between MATLAB R2020a and R2020b in the bsxfun() function.

This closes #437 .

Note it does not pass all tests, 8 fail. 6 test have minor differences close to the tolerance limit, and two in the passive yaw irregular case fail with larger errors.
Pos_diff = 2.5e-4
Force_diff = 7.9e+3


jleonqu:
I was able to check this pull request locally in my computer. I run the case using Matlab 2020b and it works.

akeeste:
@kmruehl I have discussed this issue with Mathworks support. They confirmed that R2020a contained a bug allowing Simulink to compile without an error message. This allowed bsxfun() to function incorrectly and did not give the intended results. The bug was fixed in R2020b.

My update permutes the velocity and IRKB dimensions within the convolution integral function so that the dimensions match correctly for multiplication. 8/25 continuous integration tests still fail, but the updated library should give more accurate results than previously.

akeeste:
Mathworks support also confirmed to me that this bug first appeared in R2017b (released September 2017). In 2017b, the requirements for dimension expansion were relaxed. It became possible for MATLAB to miss a runtime error for incompatible dimension operations, such as bsxfun().

I checked releases v2.2 and v3.0 which were before/after R2017b released. The convolution integration function does change between these versions, but I have not confirmed if v2.2 should have had this same error. It is likely the error in the changed function was missed because it overlapped with the simulink bug release which allowed it.

I am currently testing the accuracy of this change using the B2B application. This app compares regular wave, regular+B2B, regularCIC, regularCIC+B2B, regularCIC+SS, regularCIC+B2B+SS cases. The new CI update should match the state space response very closely. However there are some significant differences now between the CI and SS cases as shown in the radiation damping force plots:

Radiation damping in Surge:
surge

Radiation damping in Heave:
heave

Radiation damping in Pitch:
pitch

akeeste:
I solved the last problems with this bug. In addition to the permutations in the CI block, there was an issue with the selector block that chose the velocity values to pass into the CI function. I have solved both issues and the function now works correctly. All continuous integration tests are passing. 

The below figures show that the CI calculations (black lines) are now more closely aligned with the state space calculations (light blue) and likely more accurate than previously. 

Surge:
surge

Heave:
heave

Pitch:
pitch

PR #450, opened 2020-11-03
Simulink library documentation update

akeeste:
Mooring - updated masks and internal annotations
PTO, Frames, Constraints - updated internal annotations
Body - updated masks, internal annotations and some library function doc strings

No functionality changed. All tests pass.


kmruehl:
@akeeste thank you!

Issue #468, opened 2020-11-21
Error in postprocessing - ParaviewVisualization


Rodimus88:
I got the following error when I try to create paraview file for wecsim example RM3 and OSWEC.

Undefined variable "hspressure" or class "hspressure".

Error in paraViewVisualization (line 53)
            body(ii).write_paraview_vtp(TimeBodyParav, PositionBodyParav, bodyname, simu.simMechanicsFile,
            datestr(simu.simulationDate), hspressure{ii}, wpressurenl{ii}, wpressurel{ii},
            simu.pathParaviewVideo,vtkbodiesii);

Error in wecSim (line 334)
paraViewVisualization

![Capture2](https://user-images.githubusercontent.com/69836354/100194407-a8599e00-2f1b-11eb-8437-40d7cc85197a.JPG)

akeeste:
Hi @Rodimus88 

Thank you for bringing this bug to our attention. In a previous commit, the way that hspressure, wpressurenl, and wpressurel are passed internally in the post-processing functions was changed. These three references were not fully updated when creating Paraview output. 

I have created a pull request (#472) to fix these references to the cell pressures. It will soon be reviewed by another member of the WEC-Sim team and merged with master to fix this issue. Please let me know if the same problems occurs after this update.

Adam

Rodimus88:
Hi @akeeste 

thank you for the attention given, So i just need to update my wec-sim directory to avoid this error. I will inform if any error occurs afterwards.

PR #472, opened 2020-12-01
Paraview pressure fix

akeeste:
This PR fixes issue #468. A previous PR (#428) changed how the pressure outputs for paraview (hspressure, wpressurenl, wpressurel) are passed in postProcess and paraViewVisualization. The references to these three pressures are now consistent.

All continuous integration tests pass. I confirmed that the Paraview application outputs correctly for every possible combination of     
simu.nlHydro, simu.pressureDis , and simu.paraview


kmruehl:
@akeeste what do you recommend as the best way to check this PR?

akeeste:
@kmruehl I tested the PR with all possible combinations of simu.nlHydro, simu.presDis and simu.paraview, so I am not worried about the PR's ability to run without crashing. 

The aspect that I did not check is to actually load the linear, nonlinear and HS pressure output in Paraview and compare to a previous run. Agreed, the OSWEC paraview application case is probably best for this

kmruehl:
@akeeste this is ready for a merge, thank you! I made some changes to the vtk path, which should resolve issues with loading the provided Paraview state files from the applications repository. .

Issue #476, opened 2020-12-07
ERROR in reading Read_AQWA file (line194)


akashch293:
Hello,
I'm using Aqwa version 2020 R2 and Matlab 2018a
I got the following error when executed the BEMIO code

Index exceeds array Bounds

Error in Read_AQWA (line 194) 
               ind = tmpl(1:3) ; tmpl(1:3)=[ ];

I tried to figure out the problem and also compared my .LIS and .AH1 file with other correct files like aqwa_example_data.The compared files look similar but still Error is produced for my file.
I'm attaching my .AH1 and .LIS file below :
[ANALYSIS.zip](https://github.com/WEC-Sim/WEC-Sim/files/5650860/ANALYSIS.zip)

akeeste:
Hello @akashch293 

Thanks for posting your issue and uploading your AQWA files. 

We have had versioning issues with AQWA input to BEMIO in the past. It seems ANSYS makes slight changes in the .AH1 and .LIS files in each new version of AQWA and it can sometimes cause issues with reading the text files correctly with BEMIO. This is not necessarily the problem here, but just something I wanted to point out. We are working to resolve this issue. 

Regardless, I will take a look at your input and see if I can find the issue in the Read_AQWA() function.

Adam

akeeste:
Hi @akashch293 

I have found the bug with the read_AQWA script and submitted a fix (PR #478). The script tried to read a line of non-numeric characters as numbers, and caused `tmp1` to be empty and non-indexable in line 194. The fix corrects the lines of read_AQWA that are read for the Froude-Krylov force.

Once this PR is reviewed and merged, the issue will be solved. Please let me know if this does not resolve your issue.

Adam

sugarl14:
Hi @akashch293,

In the meantime (assuming you're working under a university or industry license), you could maybe reach out to your technical help desk and ask them if they have access to Ansys Workbench 18.1. GUI isn't too different between versions :)

Good luck!

akeeste:
@sugarl14 Thanks for the suggestion! and @akashch293 I hope the issue is resolved with the PR.

We have had a lot of problems of AQWA versioning in the past, so having different output versions is very helpful. I currently have output from Ansys 2015, 17.0, 19.0, 20.0. If either of you do use Ansys 18.1, I would appreciate any sample .AH1 or .LIS files you can provide!

sugarl14:
Hi @akeeste,
 I do use Ansys 18.1!

[...]

PR #478, opened 2020-12-08
update Read_AQWA

akeeste:
This PR updates the Read_AQWA BEMIO function. It fixes an issue with re-reading lines in the .LIS file that caused the FK, and scattering forces to be read incorrectly. It also updates the Froude-Krylov force definition to be the FK component, not FK+scattering.

This resolves issue #476

### DIFF
--- a/source/functions/BEMIO/Read_AQWA.m
+++ b/source/functions/BEMIO/Read_AQWA.m
@@ -165,7 +165,9 @@ hydro(F).Vo = [];
 hydro(F).cb = [];
 lnlog=zeros(length(raw2),1); % initialize line find log
 for ln=1:length(raw2)
-    if isempty(strfind(raw2{ln},'MESH BASED DISPLACEMENT'))==0
+    if isempty(find(ln==lnlog,1))==0
+        continue
+    elseif isempty(strfind(raw2{ln},'MESH BASED DISPLACEMENT'))==0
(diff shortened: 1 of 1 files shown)


kmruehl:
@akeeste I believe we should retain the 'FROUDE KRYLOV + DIFFRACTION' string, because the excitation force is incident wave (aka [Froude Krylov](https://en.wikipedia.org/wiki/Froude%E2%80%93Krylov_force)) plus the diffracted wave. Once that's resolved, I think this is ready for a merge.

akeeste:
@kmruehl Lines 182-208 are setting the values of the FK force, so this should look for the header that is only 'FROUDE KRYLOV FORCES'. Then the diffraction force searches for 'DIFFRACTION FORCES'

The total excitation could be read from the 'FROUDE KRYLOV + DIFFRACTION...' table, but we actually read this from the 'FORCERAO' in the *.AH1 file because the layout is easier to read in consistently

kmruehl:
@akeeste Thank you for the clarification. That makes sense. It's counterintelligence that the excitation force is referred to as 'FORCERAO' in the AH1 file. Since this is a bug, I'm going to merge this PR now, but we still need to do some work to comment the `READ_AQWA` file so that it's clear what's happening in each section of the script, and which file we're parsing the data from.

PR #485, opened 2020-12-16
PTO-Sim library update resolves #247

### LINKED ISSUE #247: Torque OSWEC hydraulic PTO
Hi!
I have a question about the hydraulic PTO for the OSWEC example.
When I increase the values of the piston top and bottom area, I saw that the power keeps increasing.
When I plot the velocity of the flap in pitch together with the torque applied by the PTO (calculated inside 'rotary to linear adjustable rod'), I noticed that the **torque is positive when the velocity is positive**, and **negative when the pitch velocity is negative**. Shouldn't this be the other way around?
Thanks in advance!
Regards
Nicolas

jleonqu:
This PR corrects the Rotary to Linear blocks in the PTO-Sim. The torque calculation is corrected in both blocks, the formulation was briefly described in the issue #247 

The blocks revised are the Motion Conversion blocks in the PTO-Sim library. The changes are highlighted in the next figure:

image


akeeste:
Hi @jleonqu 

I'm having a couple issues running this case. I think that in the PTO-Sim library blocks a few variables were renamed. For example it looks like ptosim.motionMechanism.crank became ptosim.crank, etc. However the ptosimClass was not updated to reflect this naming that so I can't run the case yet.

Adam

jleonqu:
Hi @akeeste 

It's kind of weird, because I didn't change the names in the blocks. This is how the block looks like originally:
image

And this is how it looks like after my changes:
image

I used the same notation. The original variable name is ptosim.crank in the blocks and I used the same name for updated version.

Jorge.

akeeste:
@jleonqu I see, this is strange. In the current [ptosimClass.m on dev](https://github.com/WEC-Sim/WEC-Sim/blob/9a7aed8305bda51c2d36dd77a943c013b6859a3e/source/objects/ptoSimClass.m#L95) I  see these variables under the ptosim.motionMechanism struct. When I tried to add them to the ptosimClass itself in the ptoSimInputFile:

   ptosim.crank = 3;
   ptosim.offset = 1.3;
   ptosim.rodInit = 5;

I get the error:
"Unrecognized property 'crank' for class 'ptoSimClass'.

Error in ptoSimInputFile (line 6)
ptosim.crank = 3;

Error in wecSim (line 136)
    ptoSimInputFile"

jleonqu:
@akeeste it looks like in the ptoSimInputFile you should add the variables like this:

ptosim.motionMechanism.crank = 3;
ptosim.motionMechanism.offset = 1.3;
ptosim.motionMechanism.rodInit = 5;

I just checked a couple of examples in the WEC-Sim_Applications repository, and that's the notation.

kmruehl:
> 
> 
> @akeeste it looks like in the ptoSimInputFile you should add the variables like this:
> 
> ptosim.motionMechanism.crank = 3;
> ptosim.motionMechanism.offset = 1.3;
> ptosim.motionMechanism.rodInit = 5;
> 
> I just checked a couple of examples in the WEC-Sim_Applications repository, and that's the notation.

@jleonqu, do we need to change the inputs to the PTO-Sim example on the applications repository too, or are they unchanged?

jleonqu:
@kmruehl the variables are unchanged, this is the original notation used in the ptoSimInputFile in all the application cases.

kmruehl:
> 
> 
> This PR corrects the Rotary to Linear blocks in the PTO-Sim. The torque calculation is corrected in both blocks, the formulation was briefly described in the issue #247

@jleonqu can you modify this description to include a screenshot of the blocks you revised, since GitHub doesn't easily track changes to binary likes, like the Simulink Library. Thanks!

kmruehl (approved):
@jleonqu thank you so much for getting to the bottom of this issue! I was able to run the [OSWEC_Hydraulic_Crank_PTO](https://github.com/WEC-Sim/WEC-Sim_Applications/tree/master/PTO-Sim/OSWEC/OSWEC_Hydraulic_Crank_PTO) PTO-Sim case and confirm that this PR resolves the sign error in the kinematic linkage.

image

``` matlab
figure()
plot(output.ptos.time,output.ptos.forceActuation(:,5))
hold on;
plot(output.ptos.time,output.ptos.forceInternalMechanics(:,5)) 
hold on;
plot(output.ptosim.time,output.ptosim.pistonNCF.force)
legend('Actuation Torque','Internal Torque','PTO-Sim Torque')
```

akeeste:
Hi @jleonqu @kmruehl I found my issue. Sorry this is not actually related to what you implemented in this PR, but to a strange way that the motionMechanism mask was set-up previously that may be worth changing to prevent confusion.

In my case, I was using both the "Rotary to Linear Adjustable Rod" and the "Angular to Linear Velocity" blocks from PTO-Sim. Normally, "Angular to linear velocity" is normally contained within the mask of "Rotary to Linear Adjustable rod". The mask takes a "ptosim.motionMechanism" struct as an input, but assigns it to the variable "ptosim" under the mask. This was very confusing as then it appears that the blocks are using ptosim.crank, ptosim.rodInit, etc. I think if the mask instead assigned the input to "ptosim.motionMechanism" or just "motionMechanism" this would prevent confusion. 

I setup simulink this way because I only needed to convert an angular velocity, not an additional force. I can work around this another way, but wanted to point out where my confusion was coming from. 

image

jleonqu:
@kmruehl I updated the description of the PR to include a figure of the blocks that were revised.

akeeste (approved):
Approved!

PR #486, opened 2020-12-17
Variant subsystem label fix resolves #479

### LINKED ISSUE #479: Variant Subystem Rendering
There's an issue with the rendering of variant subsystem blocks. For example, the `Rigid Body` block is labeled as a `Drag Body` when pulled from the WEC-Sim Library (even though it hasn't been initialized yet). This issue has been temporarily resolved on the `master` branch by "refreshing" the variant subsystem in the WEC-Sim Library, and retaining the variant subsystem variables in the MATLAB Workspace. This is not an ideal solution, and we need to come up with a more robust solution on the `dev` branch.

akeeste:
This PR fixes issue #479 where the 'Drag body' label appears on a rigid body if the Variant variables are not yet defined. This causes confusion and can make Simulink bodies appear to be a different type than they actually are.

This fix uses the 'Allow zero active variant controls' option in the rigid body which allows no label to be assigned until a body is one of the Variants is correctly activated. This requires the library to be in R2017b or later.


kmruehl:
> 
> 
> This PR fixes issue #479 where the 'Drag body' label appears on a rigid body if the Variant variables are not yet defined. This causes confusion and can make Simulink bodies appear to be a different type than they actually are.
> 
> This fix uses the 'Allow zero active variant controls' option in the rigid body which allows no label to be assigned until a body is one of the Variants is correctly activated. This requires the library to be in R2017b or later.

@akeeste, can you add a screenshot of what was changed in the library for reference?

kmruehl (approved):
Thanks @akeeste this PR resolves #479! And thanks for the other cleanup.

akeeste:
I'm adding a few screenshots of what has been changed in Simulink for easy reference. The main change that fixed the actual subsystem labeling issue is changing the block parameter option "Allow zero active variant controls". I changed this in the properties table and in the mask:
image
image


As a small clean-up item, I also changed the annotations for the excitation force in each case to reflect that the mean drift force is also included. For example:
image

This is done for each of the F_exc types:
image

PR #491, opened 2021-01-06
Clean-up responseClass

akeeste:
This PR cleans up the eulXYZ2RotMat and rotMatXYZ2Eul functions appended on to the responseClass. New functions are created in one location that can be called by the classes and Simulink.

Similarly, rotateXYZ and offsetXYZ functions are created so they can be tracked in a single location. The old methods are removed from the classes and simulink, and changed to point to the scripts in the function directory.


nathanmtom:
@akeeste Just a quick clarification, but is the intent here to move more of the functions coded within the the WEC-Sim Library to external MATLAB functions that allow for better version control? 

Therefore, you haven't changed any of the code just where it is found?

source/functions/orbital_velocity.m
```diff
@@ -1,5 +1,5 @@
 function [u,v,w,xa,ya,za] = orbital_velocity(disp,k,omega,waveAmp,theta,t,rampT, beta, h, g)
```
nathanmtom, line 1: @akeeste I'm struggling to find where this function is used within WEC-Sim. Although, we need to calculate the flow velocity and acceleration to calculate the Morison Element forces, we do this in the ME block and do not reference a separate function. Trying to determine if this is perhaps an outdated function that could be removed?

akeeste:
@nathanmtom Yes the functions are moved to a single location so that they are easier to track. They are not changed functionally.

I hadn't really looked at orbital_velocity, just noticed a typo and corrected it. I just searched the entire source directory but can't find anywhere that it is called

source/functions/orbital_velocity.m
akeeste: I just looked at the orbital velocity calculation in the ME block, it looks like an updated / rewritten version of the stand-alone function. It seems that it's likely outdated and not used

source/objects/bodyClass.m
```diff
@@ -727,31 +727,31 @@ function setMassMatrix(obj, rho, nlHydro)
...
-            rotMat(1,3) = ax(3)*ax(1)*(1-cos(t))    - ax(2)*sin(t);
-            rotMat(2,1) = ax(1)*ax(2)*(1-cos(t))    - ax(3)*sin(t);
-            rotMat(2,2) = ax(2)*ax(2)*(1-cos(t))    + cos(t);
-            rotMat(2,3) = ax(3)*ax(2)*(1-cos(t))    + ax(1)*sin(t);
-            rotMat(3,1) = ax(1)*ax(3)*(1-cos(t))    + ax(2)*sin(t);
-            rotMat(3,2) = ax(2)*ax(3)*(1-cos(t))    - ax(1)*sin(t);
-            rotMat(3,3) = ax(3)*ax(3)*(1-cos(t))    + cos(t);
-            xn = x*rotMat;
-        end
+%         function xn = rotateXYZ(obj,x,ax,t)
```
nathanmtom, line 730: Currently, these lines are commented out. After this pull request is approved and before pushing to the dev branch will these lines be removed from the bodyClass.m? This comment also goes for `offsetXYZ.m` file below.

source/objects/constraintClass.m
```diff
@@ -126,34 +126,34 @@ function setInitDisp(obj, x_rot, ax_rot, ang_rot, addLinDisp)
...
-            rotMat(1,3) = ax(3)*ax(1)*(1-cos(t))    - ax(2)*sin(t);
-            rotMat(2,1) = ax(1)*ax(2)*(1-cos(t))    - ax(3)*sin(t);
-            rotMat(2,2) = ax(2)*ax(2)*(1-cos(t))    + cos(t);
-            rotMat(2,3) = ax(3)*ax(2)*(1-cos(t))    + ax(1)*sin(t);
-            rotMat(3,1) = ax(1)*ax(3)*(1-cos(t))    + ax(2)*sin(t);
-            rotMat(3,2) = ax(2)*ax(3)*(1-cos(t))    - ax(1)*sin(t);
-            rotMat(3,3) = ax(3)*ax(3)*(1-cos(t))    + cos(t);
-            xn = x*rotMat;
-        end
+%         function xn = rotateXYZ(obj,x,ax,t)
```
nathanmtom, line 135: Repeating comment, but suggest removing the commented lines before final push to the dev branch.

source/objects/mooringClass.m
```diff
@@ -72,43 +72,43 @@ function setInitDisp(obj, x_rot, ax_rot, ang_rot, addLinDisp)
...
-            rotMat(1,3) = ax(3)*ax(1)*(1-cos(t))    - ax(2)*sin(t);
-            rotMat(2,1) = ax(1)*ax(2)*(1-cos(t))    - ax(3)*sin(t);
-            rotMat(2,2) = ax(2)*ax(2)*(1-cos(t))    + cos(t);
-            rotMat(2,3) = ax(3)*ax(2)*(1-cos(t))    + ax(1)*sin(t);
-            rotMat(3,1) = ax(1)*ax(3)*(1-cos(t))    + ax(2)*sin(t);
-            rotMat(3,2) = ax(2)*ax(3)*(1-cos(t))    - ax(1)*sin(t);
-            rotMat(3,3) = ax(3)*ax(3)*(1-cos(t))    + cos(t);
-            xn = x*rotMat;
-        end
+%         function xn = rotateXYZ(obj,x,ax,t)
```
nathanmtom, line 83: Repeat comment, about removing commented sections before pushing to dev branch.

source/objects/ptoClass.m
```diff
@@ -140,34 +140,34 @@ function setInitDisp(obj, x_rot, ax_rot, ang_rot, addLinDisp)
...
-            rotMat(1,3) = ax(3)*ax(1)*(1-cos(t))    - ax(2)*sin(t);
-            rotMat(2,1) = ax(1)*ax(2)*(1-cos(t))    - ax(3)*sin(t);
-            rotMat(2,2) = ax(2)*ax(2)*(1-cos(t))    + cos(t);
-            rotMat(2,3) = ax(3)*ax(2)*(1-cos(t))    + ax(1)*sin(t);
-            rotMat(3,1) = ax(1)*ax(3)*(1-cos(t))    + ax(2)*sin(t);
-            rotMat(3,2) = ax(2)*ax(3)*(1-cos(t))    - ax(1)*sin(t);
-            rotMat(3,3) = ax(3)*ax(3)*(1-cos(t))    + cos(t);
-            xn = x*rotMat;
-        end
+%         function xn = rotateXYZ(obj,x,ax,t)
```
nathanmtom, line 149: One more documentation of removing commented lines to clean up code.

source/functions/orbital_velocity.m
nathanmtom: Thanks, that was my initial conclusion after reviewing this file and noticing this is not called in the ME block. If no one else on the WEC-Sim team has a reason to keep this function then suggest removing before pushing to the dev branch.

nathanmtom:
@akeeste All of the changes `.m` files look good from my review. The last item I want to review is the library, but since it is more difficult to see where you made changes could you quickly identify which blocks you had to modify?

source/functions/eulXYZ2RotMat.m
```diff
@@ -0,0 +1,22 @@
+function rotMat = eulXYZ2RotMat(phi, theta, psi)
```
dav-og, line 1: This is probably a broader point...but have we decided on a naming convention for functions?

My personal preference would be eulxyz_2_rotmat(), but I think I used eulXYZ2RotMat() originally in keeping with the rest of the code. But IIRC there were mixed conventions and we wanted to resolve that problem? Are we going for camel case function names?

@kmruehl

dav-og: in fact, the orbital_velocity() function below has an underscore. I would prefer underscores for functions, camel case for variables. Let me know what you think 👍

source/objects/bodyClass.m
akeeste: Yes I meant to remove those comments, I'll push that before merging with dev

source/functions/eulXYZ2RotMat.m
akeeste: I agree the naming convention is a bit inconsistent, especially with variables. I think this is a broader cleanup item on the project board. 

I think we have waited to address this so far because there were concerns about backwards compatibility if a lot of the class variables are renamed

akeeste:
@nathanmtom I just pushed a commit to cleanup the commented functions.

The simulink blocks that I removed the functions from are:
- nonlinear wave elevation / instantaneous water free surface
- hydro body / nonlinear hydrostatic restoring force
- hydro body / diffraction force / nonlinear FK force
- flex body / nonlinear hydrostatic restoring force
- flex body / diffraction force / nonlinear FK force

In each case, the function is changed at the lowest level in the imbedded matlab function of each block.
Let me know if I can clarify further so that you don't have to dig for them

source/functions/eulXYZ2RotMat.m
dav-og: yeah that's a good point, it will take time to propagate the convention throughout the whole code. But perhaps this PR is a good moment to decide on a convention? (If we haven't already)

dav-og:
> 
> The simulink blocks that I removed the functions from are:
> 
>     * nonlinear wave elevation / instantaneous water free surface
> 
>     * hydro body / nonlinear hydrostatic restoring force
> 
>     * hydro body / diffraction force / nonlinear FK force
> 
>     * flex body / nonlinear hydrostatic restoring force
> 
>     * flex body / diffraction force / nonlinear FK force
> 
> 
> In each case, the function is changed at the lowest level in the imbedded matlab function of each block.
> Let me know if I can clarify further so that you don't have to dig for them

I think the quaternion2EulXYZ.m file you've created should be called in Hydrodynamic Body/Structure/Motion Sensor as well?

Also in the description of that function:
```
%   Q : 1 x 4 float vector
%       List of quaternions
```

I don't think its a list of quaternions? Just one quaternion

source/functions/eulXYZ2RotMat.m
akeeste: @dav-og I was going to bring up the convention today, but ran out of time.

How about we leave these camel case for now? The rest of the functions in /source/functions are currently camel case. Though I agree, we should have a different convention for variables, functions, classes. Maybe the team can discuss during the working meeting.

akeeste:
@dav-og I'll update the descriptions for the quarternions. There's not really a change here, as I didn't end up making it an external function. 

I can make it an external function, so that git can track a text file. Before we do that across the library though, I'm curious to know if there is a speed difference in how simulink calls functions (interpreted matlab function vs embedded matlab function vs calculated in simulink, etc).

source/functions/eulXYZ2RotMat.m
H0R5E: I would love to see some kind of coding guidelines introduced. It would be great to see a Python style line length restriction, as some of the lines are way too long at the moment. Consistent naming style is great too - I'm not sure if you necessarily need to distinguish every type of variable, but consistency helps a lot.

I wouldn't recommend implementing this in a big bang style, however, as you won't be able to properly review such a huge change. This kind of thing is best implemented as other changes are being made, at least until there is very little left to modify.

I think someone should open an issue to discuss it. We might need a doodle poll to vote on the options!

H0R5E: Another thing is you can get a handle on the "accepted" naming conventions from the MATLAB docs. E.g:

* [Classes, classMethods and Properties](https://uk.mathworks.com/help/matlab/matlab_oop/create-a-simple-class.html)
* [functionDefinitions](https://uk.mathworks.com/help/matlab/ref/function.html#bvfsr6o)
* [variables_whoKnows](https://uk.mathworks.com/help/matlab/matlab_prog/variable-names.html)

There isn't any clarity on how a variable should be named. Personally I prefer UPPERCASE GLOBALS, for instance.

source/functions/orbital_velocity.m
kmruehl: @nathanmtom I agree

source/functions/eulXYZ2RotMat.m
kmruehl: @dav-og and @akeeste we decided on camel case for WEC-Sim a long time ago (many years ago), and I think we should stick with it. Yes, we do have functions/variables that did not follow this naming convention, and we should try to clean them up this year as we we come across them. This is the general cleanup on the project board @akeeste referred do.

kmruehl:
@akeeste I'm not sure about the speed difference, but I'm included to move to do an external function just so that we can make our jobs easier on SCM. I'm willing to take the hit on speed, assuming it's not a huge difference. If you're curious you could compare the time :)

akeeste:
@dav-og  @kmruehl 
I updated the description of the quaternion function and changed it from an 'interpreted matlab function'.
According to [this description](https://www.mathworks.com/help/simulink/ug/comparison-of-custom-block-functionality.html), the interpreted function is slightly slower but more importantly it does not support code generation for rapid-accelerator mode.

My solution was to keep an embedded matlab function (untracked), but only use it point to an external matlab function that can be tracked, i.e.:
`function E = fcn(Q)`
`% Call quaternion matlab function. `
`% quaternion2EulXYZ() saved as separate text file for Github tracking purposes.`
`E = quaternion2EulXYZ(Q);`
`end`

I'm curious to see Hunter's solution in the PR that he will post. Hopefully that will give us another solution, or fix the tracking problem entirely as is.

dav-og:
hey @akeeste 

> @dav-og @kmruehl
> I updated the description of the quaternion function and changed it from an 'interpreted matlab function'.

the updated description states:

> Euler Angle Type II or Tait-Bryan Angle for sequence 321 or zyx (Nautical or Cardan angles).

I think its 'Euler/Tait-Bryan angles; x-y-z extrinsic convention'

akeeste:
@dav-og 
Yes you're right, I believe it uses the xyz convention. I need to merge dev into this PR, I'll change the description when I do that and then merge.

PR #499, opened 2021-01-18
Paraview bug with simMechanics file name

akeeste:
This PR fixes a bug with the simMechanics file name in paraview output. The simMechanics file was previously changed to include the path for compatibility with WEC-Sim PCT. Paraview output requires just the file name to define the CASE.pvd file. This is resolved by this PR.

.pvd and .pvsm files can be opened successfully with Paraview now.


kmruehl:
@akeeste Thank you very much! This resolves the Paraview bug that kept us from opening the PVD file.

PR #503, opened 2021-01-21
Update allowing simulations ran from Simulink file

jhbates:
This update allows for simulations to be ran straight from the Simulink file. A masked subsystem, Parameters, in the Simulink model allows for all input parameters to be selected. There are two options for parameters, load from input file or custom parameters. The new model is saved in a new example, RM3withButton. All new functions used by the model are stored in the source/functions/SimulinkModelFunctions folder. These functions include Init and Stop functions for the model as well as all callbacks for the masked subsystem parameters.


akeeste:
Hi @jhbates 

Sorry for the delay on this, I reviewed your PR this week. I had a couple issues pulling in the most recent version of wecSim. into the wecSimInitFunction.m and wecSimStopFunction.m but I think they are resolved. However I'm still getting an issue running the RM3withButton from Simulink that I can't resolve:

" Undefined function 'RM3withButton_sfun' for input arguments of type 'double'. "

It is very strange, I get this error every other time I run the Simulink model. If I run with a clean workspace it appears, but if I immediately attempt rerunning the model from Simulink after this, the error does not appear.
Have you encountered this before? It seems like 'RM3withButton_sfun' is created when Simulink is run but after it should be..


Also, I am working to replace wecSim.m with essentially 3 calls:
wecSimInitFunction.m
wecSimStopFunction.m

In this way, wecSim.m can still be called and have the same functionality without all commands duplicated in two places. Do you see any problem with this approach?

If you like I can contribute to this PR and push here, or I can push any changes to a different branch and we can compare.

Let me know what you think, thanks for your help on this!

Adam

jhbates:
@akeeste 

Unfortunately I have been between projects since the start of the month and do not have access to Matlab. I should have it within the next week or two, and then I will be able to help with the "sfun" issue. I have not seen that issue before.

In regards to replacing wecSim.m,
- wecSimInitFunction.m utilizes bdroot in order to get the correct .slx file name. This function requires the Simulink file to be open at the time it is called. This could certainly be replaced in order to allow wecSimInitFunction.m, and therefore wecSim.m, to be called without having the Simulink file open.
- If you want to be able to run the model from the Simulink file itself, wecSimInitFunction.m and wecSimStopFunction.m have to stay in the callbacks of the model. Therefore, with the current wecSim.m and the proposed wecSim.m both will run duplicated code.
- The only option I can think of right now to reduce running duplicated code, would be to change wecSim.m to simply run the Simulink file with no other commands. 

As far as your work, I would just have you contribute to this PR. I will work on solving the issues as soon as I can.

Thank you,
Hunter

akeeste:
@jhbates 

We weren't sure what you would prefer, so we ended up creating a feature branch from the main code called 'feature_simulink_run'. I pulled your PR and commits into that branch and have been pushing my own changes there. I think this will be easier in the long run as more of the team pulls the code and may want to push any updates. 

I did end up changing wecSim.m essentially as you suggest. Right now it just calls an input file to get the simulink model name, and then runs the model. The model then calls wecSimInitFunction.m and wecSimStopFunction.m. My biggest change so far was to integrate your nice masking into the library blocks themselves so that the number of bodies, constraints, etc does not have to be pre-set. This way, every WEC-Sim simulink model will have these capabilities without requiring users to add another block. 

There's still a few changes we may make, depending how we want to integrate this into the library, but the main issue is this '_sfun' bug. I think it has to do with upgrading the simulink model to 2020b, but I'm uncertain... If you get a chance to look, that would be great. Otherwise I'll keep working on this and trying to figure it out!

Thanks for the great work!
Adam

akeeste:
Continuing notes of my progress on the feature_simulink_run branch here.

I have worked with MATLAB to identify the 'RM3withButton_sfun' error. This is a bug that first appears in R2020b. There is a timing issue with using Simulink initialization callbacks when said callbacks also define a stateflow or variant subsystem in the model. The timing is why the model works when run once.

* 1st run - the variant subsystems parameters are set in wecSimInitFunction.m but not early enough for the simulink model to process them.
* 2nd run- the variant parameters are already defined so the model works. Variants deleted after successful simulation
* 3rd run - variants again not defined, same issue as 1st run. 

These issues can be avoided by calling 'wecSimInitFunction' in the command line, and then running the simulink file as normal, or using the 'double-run' method above. The command line + simulink option is equivalent to the current method of putting a breakpoint at the 'sim(...' line in wecSim.m and then running the Simulink file manually--though with the added benefit of being able to define parameters in Simulink.

If this feature is incorporated into the main WEC-Sim branch, it will prove problematic as R2020b will always have this bug.

akeeste:
See the branch feature_simulink_run and new PR (#548 ) for that branch to continue any discussion on this PR.

PR #506, opened 2021-01-29
Dev Update to the bodyClass.m

nathanmtom:
Updated the initialization of the Morison Element values from NaN to 0. The initialization of NaN was causing issues when runing a multibody system where only one body has Morison Elements. The body without Morison Elements would error out even though the body did not have active elements.


nathanmtom:
@kmruehl I wasn't sure if we wanted this to go to Dev and then merge with Master for a bug fix release. I can also make the same changes and push to Master.

kmruehl:
@nathanmtom I think submitting the PR into dev is a good approach, and we should tag a bug fix release ASAP. Thank you for taking care of this.

PR #509, opened 2021-02-08
add WecSim if for MCR loadFile

dforbush2:
This is a 3 - line change to wecSim.m that adds an additional flag for MCR runs. Right now, there is not a good workflow that I am aware of to step through a variety of user-input files (as needed for a "From Workspace" simulink input block, for instance), unless those input files can be specified in such a way that they can fill an existing field in a class, e.g. waves.etaDataFile. This implies some constraints on array size and data type that can create some annoying and inconsistent work arounds that are likely prohibitive for the broader user base. 

This would be specifically useful in the forthcoming system identification examples to prescribe an actuation time series to an actuator or a set of matrices to act as controllers, and I would argue is of potentially broad utility to the user-base as a whole, since it allows easy MCR runs iterating over a commonly-used block type. In this way, multiple condition runs become more flexible.

This does not affect any existing WEC-Sim functionality, and only operates if mcr.header={'LoadFile'}.


dforbush2:
Example usage: 
Create 3 unique time series, of a format recognized by a From Workspace block. Each time series needs the same variable name (e.g., variableName), but to be saved as a unique file name, say 'FilePath1', 'FilePath2', and 'FilePath3'. The From Workspace block should reference 'variableName'.

Now, make an mcr file like so: 
mcr.header{1,1}= 'LoadFile';
mcr.cases{1,1} = 'FilePath1';
mcr.cases{2,1} = 'FilePath2';
mcr.cases{3,1} = 'FilePath3';
And save it. Specify this mcr file in your wecSimInputFile.

Then, running 'wecSimMCR' will for imcr ==1, load the file at FilePath1 and simulate (with the  From Workspace block referencing the variable loaded from FilePath1) . For imcr ==2, loads file at FilePath2, simulates (with From Workspace reading variable loaded from File Path2). and so on. We don't update the variable name in From Workspace each time, that's why its important to have the same variable name be loaded from each file path.

yuyihsiang:
Dom, Can you push it to Dev instead as suggested by Kelley.

yuyihsiang (changes requested):
The changes look good and should resolve the load file issue. However, please push it to the Dev branch instead.

PR #512, opened 2021-02-15
Update allowing simulations ran from Simulink file #503

kmruehl:
This update allows for simulations to be ran straight from the Simulink file. A masked subsystem, Parameters, in the Simulink model allows for all input parameters to be selected. There are two options for parameters, load from input file or custom parameters. The new model is saved in a new example, RM3withButton. All new functions used by the model are stored in the source/functions/SimulinkModelFunctions folder. These functions include Init and Stop functions for the model as well as all callbacks for the masked subsystem parameters.

from @jhbates


kmruehl:
Hi @jhbates

Sorry for the delay on this, I reviewed your PR this week. I had a couple issues pulling in the most recent version of wecSim. into the wecSimInitFunction.m and wecSimStopFunction.m but I think they are resolved. However I'm still getting an issue running the RM3withButton from Simulink that I can't resolve:

" Undefined function 'RM3withButton_sfun' for input arguments of type 'double'. "

It is very strange, I get this error every other time I run the Simulink model. If I run with a clean workspace it appears, but if I immediately attempt rerunning the model from Simulink after this, the error does not appear.
Have you encountered this before? It seems like 'RM3withButton_sfun' is created when Simulink is run but after it should be..

Also, I am working to replace wecSim.m with essentially 3 calls:
wecSimInitFunction.m
wecSimStopFunction.m

In this way, wecSim.m can still be called and have the same functionality without all commands duplicated in two places. Do you see any problem with this approach?

If you like I can contribute to this PR and push here, or I can push any changes to a different branch and we can compare.

Let me know what you think, thanks for your help on this!

Adam
@akeeste

Issue #516, opened 2021-02-18
setInitDisp function


akeeste:
The setInitDisp function can't be used in the wecSimInputFile.m without first defining a body's cg. Since the cg of the body is read from the H5 file after the input file is called, this function can't be used properly.

Manually setting the body.initDisp maxtrix is limited in that any rotation is defined about the cg and not any other point.

We should modify this function so that it can be used in the input file, without defining a cg that will go on to be overwritten when calling wecSim.m

PR #535, opened 2021-03-12
Update to Resolve Issue #516

nathanmtom:
The body class was updated so the setInitDisp function does not look to call the center of gravity of the body. Rather the user is now required to define the distance from the point of rotation to the center of gravity which would be a pre-calculated value.


akeeste (approved):
@nathanmtom 

This PR looks good. I checked it with one of the OSU course homeworks and it works well without redefining the body cg. I didn't see any mention in the docs or Applications repo that needs to be changed, so it looks good to go!

Adam

jleonqu:
@nathanmtom 

I used this PR to do the problems of the OSU course homework and it works well for me, the issues that I had were solved and the angular free decay tests for the OSWEC run well.

kmruehl:
@nathanmtom  both @akeeste and @jleonqu  have reviewed this PR and said it's good to go. I'm going to merge it. Thank you for this clean fix!

PR #542, opened 2021-03-17
Resolve the issue of running CAPYTAINE examples on OSX and Linux

yuyihsiang:
- Fix case-sensitive issue for the folder on OSX and Linux
 - Modify the run_cases.py file


akeeste (approved):
@yuyihsiang 

All of these cases still run on Windows! It looks okay to merge.

Adam

dav-og:
I get the following error message:

image

This has come up on Capytaine's issues before:
https://github.com/mancellin/capytaine/issues/2

Workaround seems to be changing

`    cpt.io.xarray.separate_complex_values(capyData).to_netcdf(ncFName)
`

to

`    cpt.io.xarray.separate_complex_values(capyData).to_netcdf(ncFName,
                      encoding={'radiating_dof': {'dtype': 'U'},
                                       'influenced_dof': {'dtype': 'U'},
                                'convention': {'dtype': 'U'}})
`

akeeste:
Based on the capytaine issue, this seems to be an xarray bug. However, my version of xarray (v 0.16.1) is not supporting the suggested 'convention' encoding. Although it works without any specified encoding and with just 'radiating_dof' and 'influenced_dof' defined. 

@dav-og  What version are you using?

Adam

dav-og:
hey Adam, yeah I also get a KeyError for 'convention', not exactly sure whats going on there...but leaving that out and just defining 'radiating_dof' and 'influenced_dof' seems to work.

I did have xarray v0.15, but I upgraded to v0.17 and still got the same error - weird ¯\_(ツ)_/¯ ... at least passing the encoding dictionary seems to work for all versions?

dav-og (approved):

mancellin:
Hi!

Just FYI, "convention" was something that was in Capytaine at the time of mancellin/capytaine#2, but has been moved since.

Long version: Wamit and Nemoh have different conventions for e.g. time dependency and I meant to store with the data a hint about the convention that had been used. At the time of mancellin/capytaine#2, each data point had its own string naming the convention, but it was a bit wasteful and having more than one convention in a dataset would have been a mess. So it has been moved to be a  single metadata of the xarray dataset (`capyData.attrs['convention']` or something like that). And today, this is one of the half-baked features of Capytaine with which I don't know what to do.

kmruehl:
@yuyihsiang please remove the two conflicting files so we can merge this PR

kmruehl:
@yuyihsiang this has been merged and conflicts should be resolved

Issue #546, opened 2021-03-23
Matlab R2021a bug - mask initialization


akeeste:
There seems to be a change in when Matlab R2021a handles mask initialization in Simulink. There are currently displacement, velocity and acceleration goto tags in each body block. If a user creates multiple bodies, the mask initialization takes the number input by the user, and updates the names of all relevant numbered blocks using that number.
So vel1, disp1 become vel2, disp2, etc for body #2.

It seems Matlab is now (in 2021a) checking for duplicate goto tags before running the mask initialization. There maybe be a way around this, but it is difficult to know Simulink's order of operations behind the scenes. The easy solution is to report a bug to Mathworks and wait for it to be patched. R2021a was just released 3/15/21, so this version should not be widely used and hopefully patches are released quickly. I'm making this issue so that WEC-Sim users/dev team are aware that R2021a is not currently usable.

I will submit a Mathworks issue on this and if necessary look into ways around this bug.

Adam

akeeste:
This issue is actually not a bug on the Mathworks side, but on how we check for version compatibility in the hydro/nonhydro/drag/flex body block masks. In R2021a, the Matlab version becomes v9.10. However, when this is converted to a numerical value it becomes '9.1' which incorrectly passes our version test of being '<9.7' and a nonexistent parameter is set. This led to an error in the mask initialization, which stopped early before renaming the goto tags. This gave a misleading error message related to the goto blocks but not the mask initialization error.

The version check is changed to:
ver = split(version,'.');
ver = str2num(ver(1:3));
if ver(1) < 9 || (ver(1)==9 && ver(2)<7)
   set_param(h,'ExtGeomFileType',geomType);
end

from the previous check:
ver=version;
ver=str2double(ver(1:3));
if ver<9.7
   set_param(h,'ExtGeomFileType',geomType);
end

mberk06:
@akeeste what's the schedule for releasing the fix? No rush, just excited to keep exploring the library.

akeeste:
@mberk06 

This is a small fix, but we want to have all changes to the master branch reviewed by a different team member to prevent creating another issue. Hopefully this can be implemented next week.

akeeste:
@mberk06 this R2021a bug should now be resolved! Please reopen this issue if you have any problems.

PR #548, opened 2021-03-26
Run WEC-Sim from Simulink

akeeste:
This PR expands upon @jhbates work (PR #512) and allows WEC-Sim to run from Simulink.

There are now five ways to run WEC-Sim (two additional):
1. 'wecSim' in command line. Uses standard wecSimInputFile.m
2. 'wecSimInitFunction' in command line + running in Simulink using custom parameters (no input file)
3. 'wecSimInitFunction' in command line + running in Simulink and loading an input file
4. 'wecSimMCR' in command line
5. 'wecSimPCT' in command line

If starting in Simulink (2,3 above), the user must run the new WEC-Sim pre-processing command 'wecSimInitFunction' prior to running every simulation. This extra step avoids a previously mentioned bug in R2020b. MCR and PCT can be used with the command line, but currently not when running WEC-Sim from Simulink. Regardless of the run method used, the WEC-Sim post-processing is run as the Global Reference Frame's stop function callback in Simulink.

Custom parameters may be set inside of the Simulink file using the mask for each WEC-Sim block. The masks currently contain most (but not every) options available in the text-based input. Whether the Simulink model runs an input file or uses custom parameters is determined by a single Global Reference Frame mask setting. This currently defaults to using an input file. Additionally, if custom parameters are used and WEC-Sim is run from Simulink then the mask parameters are written to a 'wecSimInputFile_simulinkCustomParameters.m' which is used for that simulation. To make initializing custom mask parameters easy, there is a button in the Global Reference Frame that prepopulates mask values using a previous input file.

New files
* wecSimInitFunction - contains all wecSim.m pre-processing (before former wecSim.m line with `sim(simulinkModel)...`)
* wecSimStopFunction - contains all wecSim.m post-processing (after former wecSim.m line with `sim(simulinkModel)...`)
* SimulinkModelFunctions directory - contains all mask callback functions

This PR passes all continuous integration tests and also works with the various application case tests in PR #508. It should have the same backwards compatibility as the current dev branch does.

@yuyihsiang Can you test this PR with wecSimPCT? I have tested with MCR but do not have the parallel computing toolbox installed.
@jleonqu I believe you looked at this work a bit previously. Can you take a look at this PR generally and let me know your suggestions?


jleonqu (approved):
I liked the way you implemented this feature in Simulink, it is very convenient, easy to use, easy to understand, and very useful. Regarding MCR, I think it's not a big deal if you have to run the simulation with the command line, actually I prefer that way when I am running multiple simulations with the same model because for me it is easy to adjust the multiple runs using a Matlab script.

Just as a comment, the changes worked when I used Matlab 2020b, but they didn't work with Matlab 2021a (I had the same issue described in #546 ), which I think make sense, since #553 has not been merge into master.

kmruehl (approved):
@akeeste this should be ready for your final review. Here are my notes:
- updated docs to add Running from Simulink to advanced features and simplified workflow/examples
- changed `wecSimInitFunction` to `wecSimInitialize`
- added to .gitignore
- added brief section on `wecSimFcn` related to #598

@akeeste To Do:
- [ ] add custom WEC-Sim error if run from Simulink by hitting play and haven't executed `wecSimInitialize`
- [ ] open issue or project to see if we can run from Simulink by hitting play (and run executing `wecSimInitialize`)

akeeste:
@kmruehl I am ready to merge. I added a few small fixes and created project board cards for the last two items:
https://github.com/WEC-Sim/WEC-Sim/projects/53#card-62514170
https://github.com/WEC-Sim/WEC-Sim/projects/53#card-62514133

kmruehl:
@akeeste thank you! I made a minor revision and now I'm merging into dev

Issue #599, opened 2021-06-07
[BUG] Read_CAPYTAINE doesn't allow <6 DOFs


akeeste:
**Note: italicized text below is include as an example and should be updated before submission. If you feel any section is not applicable to your request, please replace with 'N/A' rather than delete a section.**

**Describe the bug**
The current Read_CAPYTAINE() BEMIO function can read standard Capytaine output with 6 DOFs, but does not work with output using <6 DOFs.

**WEC-Sim file(s)**
Read_CAPYTAINE.m

**To Reproduce**
With a given Capytaine .nc file containing less than 6 DOFs:
1. hydro = Read_CAPYTAINE(hydro,'PATHTONCFILE');
2. hydro = Radiation_IRF(hydro,[],[],[],[],[]);
3. hydro = Radiation_IRF_SS(hydro,[],[]);
4. hydro = Excitation_IRF(hydro,[],[],[],[],[]);
5. Write_H5(hydro)
6. Plot_BEMIO(hydro)

**Expected behavior**
A saved .h5 file with no error.

**Screenshots**
N/A

**Workstation Information (this section MUST BE completed):**
 - OS: Windows 10
 - MATLAB 2020b
 - WEC-Sim Release v.2

**Additional context**
N/A

akeeste:
@dforbush2 PR for this coming shortly

PR #600, opened 2021-06-07
Bugfix read capytaine

akeeste:
There's a bug in the Read_CAPYTAINE BEMIO function that doesn't allow <6 DOFs.
This PR fixes this problem and solves issue #599.


akeeste:
@dforbush2 This should solve your issue with the OWC BEM data. I checked using the output you sent me and BEMIO seems to be working now.

PR #604, opened 2021-06-08
Library fix for fixed constraint mask

akeeste:
Somehow during the course of the Run From Simulink development or merging various library changes into that PR, the mask initialization for the Fixed Constraint was removed. This issue affects the WECCCOMP case because the error only appears with more than 1 fixed constraint where both would attempt to use the same To Workspace name: constraint1_out.

This was the only mask initialization to change, all others are the same as the previous version of WEC-Sim.

This PR adds the correct mask initialization back into the Fixed Constraint and adds the WECCCOMP application to the Compilation CI Tests.


kmruehl:
example code from mask initialization for fixed constraints
``` matlab
if exist('constraint','var')
    h=[gcb '/output'];
    responseName=['constraint' num2str(constraint.constraintNum) '_out'];
    set_param(h,'VariableName',responseName);
end

constraint.checkLoc('W');
````

PR #608, opened 2021-06-09
WEC-Sim Installation/activation/deactivation using Startup and Cleanup scripts

salhus:
Hi All,

I have created two scripts, that can simplify the installation process,
[cleanup_m.txt](https://github.com/WEC-Sim/WEC-Sim/files/6627005/cleanup_m.txt)
[startup_m.txt](https://github.com/WEC-Sim/WEC-Sim/files/6627006/startup_m.txt)

startup.m : This scripts initiates the WEC-Sim environment and adds all necessary sub-folders on the MATLAB path. This script also sets simulink to put all the extra slrpj files in a 'work' folder.
cleanup.m : This script removes all the folders and sets the simulink back to default settings.
An empty folder titles 'work' needs to be created so that the startup.m script runs without errors.
The advantage of these two scripts is that it simplifies the installation process and the activation every time a user needs to use WEC-Sim, but then change the working folder to some other folder and do other work.

A new user will just need to download the wecsim folder from GitHub and run the startup script and not go through setting up the paths and MATLAB startup directory manually.

I am attaching the files here.

Cheers,
Sal


nathanmtom:
@salhus Thanks for the updates to the startup/cleanup files for WEC-Sim. I want to review them, but it looks like you have pushed some additional changes to your `dev` branch for WAMIT and Capytaine that are now in this branch. If possible, you may want to close this pull request and create a new branch, just for the start-up/clean-up changes.

PR #610, opened 2021-06-09
Water Visualization Block 

salhus:
Hi,

I noticed on the potential projects under new features board, that a water surface visualization is desired.

Here is a simple version of a water surface visualization. The water surface can become its own block that gets connected to the global coord frame. Attaching screenshots of the model.

Cheers,
Sal
water_surface_block
water_surface_block_inside

https://user-images.githubusercontent.com/84348506/121437857-ee72ce00-c950-11eb-89f3-b50f97cf6286.mp4


nathanmtom:
@salhus Thanks for starting the discussion and implementation of visualization of the free surface in WEC-Sim. The implementation is very straight forward and appreciate including a sample animation to show how this would look after a WEC-Sim simulation is completed. However, the WEC-Sim development team feels that the planar surface moving at the wave elevation at the origin of the model is a bit confusing as it doesn't capture the spatial variation. Perhaps for this animation we can make the planar surface smaller to be more of a point measurement that is helpful to see in the Simscape Multibody visualizer. We would also like to be able to describe the the variation in the wave profile around the floating body to better capture the wave profile, but felt that using the outputs from WEC-Sim to generate a video as a post processing function would be easier than in the Simscape Multibody visualizer. Let us know if the comments and requests above are clear and happy to discuss further if you have questions or comments. 

Cheers, 
Nathan
National Renewable Energy Laboratory

PR #615, opened 2021-06-17
Clear temporary variable added in startup.m

H0R5E:
I was reinstalling WEC-Sim on a new operating system and I remembered that one of the annoying aspects of using the `startup.m` file was the temporary variable `wecSimPath` being placed into the global workspace. 

I don't believe this variable is needed in any other situation, so my preference is to clear it from the workspace, which this PR does automatically.

### DIFF
--- a/source/wecSimStartup.m
+++ b/source/wecSimStartup.m
@@ -4,4 +4,4 @@
 
 wecSimPath = '<wecSim>';
 addpath(genpath(wecSimPath));
-
+clear wecSimPath


akeeste (approved):
I would agree, the ``wecSimPath`` variable has no use outside of the startup file. I never noticed this because I clear it myself already.  Since we direct users to copy the wecSimStartup.m file though, it makes sense to remove it in this same script.

Issue #619, opened 2021-06-21
[Developer Issue] WEC-Sim Testing Master Plan (Proposal)


H0R5E:
Following an offline (ish) disussion with @akeeste, I would like to make some proposals for improvements to the WEC-Sim testing processes. The ultimate goal of these changes would be as follows:

> * To have an efficient, high coverage test suite
> * That can be run automatically
> * That could potentially be used for automatic PR merging (for a whitelist of users)

To this end, Adam and I discussed a number of changes, as follows:

1. <s>**Testing the WEC-Sim_Applications repository separately from WEC-Sim**</s>
    - This is probably the most controversial suggestion, but I think there are a number of good reasons
    - For one, the dependency of the WEC-Sim_Applications repository for WEC-Sim is the wrong way around. WEC-Sim_Applications should require WEC-Sim, not visa versa
    - Also, requiring to check-out WEC-Sim_Applications makes the test suite very fragile
    - Moving the WEC-Sim_Applications tests will remove a burdensome load from the WEC-Sim tests (with some caveats, to be addressed below)
    - In the WEC-Sim_Applications repo, each application could be a separate test that can be triggered on a schedule or by pushes to WEC-Sim.
    - Some of the applications tests do provide useful scope for the WEC-Sim integration tests and selected tests should be copied to (and managed as part of) the WEC-Sim code.
1. <s>**Copying of files for testing should stop**</s>
    - Although this is done with good intentions, I think there are a few reasons why it should stop
    - Notably it is difficult (following the copying) to determine what should and shouldn't be checked-into the repo, risking files being added that should not be.
    - It also uses extra space on the user's drive unnecessarily
    - Copying should either be made to temporary space and then deleted or output files should be directed to temporary files and deleted.
1. <s>**The test directory should not be added to the path**</s>
    - Adding the test directory to the path can be risky behavior as it can cause file shadowing
    - A way of conducting the tests without changing the path is required
1. **Clearing the console should be suppressed when testing**
    - If tests fail it is very difficult to determine the cause of failure because the history has been deleted
    - Maybe adding an option for WEC-Sim to run in a quiet mode would reduce the amount of output
1. <s>**Figures should always be closed or not opened at all**</s>
    - Figure generation can be suppressed (see [here](https://github.com/SNL-WaterPower/WecOptTool-MATLAB/blob/master/tests/examples/RM3/ExamplesTest.m) for an example)
1. **Integration tests (i.e. most of the current tests) should be streamlined**
[...]

PR #620, opened 2021-06-21
restructure CI tests into class-based unit tests

akeeste:
This PR restructures the CI tests into class-based continuous integration tests. Right now there are classes for each: 
- regression tests
- compilation tests
- BEMIO
- rotation tests
- run from Simulink tests

Each class has specific setup, teardown and test methods. I also added tests for the BEMIO functions to ensure they are working correctly. This should make it easier to properly skip long tests (irregular Passive Yaw, MCR application, etc) by marking them "incomplete" instead of "failed".

I still need to work on improving the test runtime. Creating a draft PR so @H0R5E can review.


akeeste:
In a few local trials I have the following approximate breakdown of runtime per test class, including the contributions of the longest individual tests. In total, the tests take about ~11 minutes when run on my laptop.

- BEMIO tests: 10%
- Compilation (application) tests: 51%
    - B2B case 4: 18%
    - MCR case: 8.5%
    - mooring case: 7.5%
- Regression tests: 36%
    - regular cases: 5%
    - irregular cases: 4%
    - passive yaw cases: 27%
- Rotation tests: 0.03%
- Run from Simulink tests: 3%

If the application repo tests are removed as suggested in issue #619, in addition to skipping the passive yaw cases, this decreases the test runtime by ~80%. The integration tests can be gradually broken down into unit tests over time to prevent duplicate tests. This will improve runtime even further in the future.

H0R5E:
See https://github.com/akeeste/WEC-Sim/pull/7

akeeste:
A few additions since this PR has been opened:
- removed applications tested from the WEC-Sim repo. Applications Repo PR 7 adds these tests to that repo.
- moved the passive yaw regression tests to a separate class so that they can be turned on/off easily. By default they are skipped to decrease run time
- Changed how the tests are called in wecSimTest. Now this function uses the ``run(suite)`` command instead of ``testRunner.run(suite)`` so that appropriate debugging information is collected in the test results.

As indicated in the above discussion, removing the applications and passive yaw tests significantly decreases run time. Now the tests take ~2 minutes locally, only 15% of the previous run time.

H0R5E:
Adam, how much time is saved by not running the passive yaw regression test? My concern is that if they are not run by default, then they will never be run at all.

I would also consider replacing the test runner. If we are going to collect coverage info (which I think we should soon) we need to use a plugin on a test runner:

https://uk.mathworks.com/help/matlab/ref/matlab.unittest.plugins.codecoverageplugin-class.html

What is the difference in output between using run and using the test runner?

akeeste:
@H0R5E 
That is a fair point, most users might not remember to turn on passive yaw when appropriate. Maybe we just remove the passive yaw regression altogether? The functionality will be covered in the Applications repository, though without a comparison to the previous result.

My problem with the test runner is that it currently does not give the same diagnostic information as the run command (in R2021a). For example, right now the first BEMIO test fails (the fix is already in dev). 

When using the testRunner, `results(1).Details` is empty. Whereas the `run(suites)` command gives useful debugging information in ``Details.DiagnosticRecord`` such as ``Exception, Stack, Scope, Report`` that provide details on where the error occurred and the error message received.

This almost seems like a Matlab bug, I don't know why the testRunner would give less information than the standard run command. If the testRunner will be required for code coverage information, I can look again at how to get around this issue. However until code coverage is implemented, I think the run command is more insightful.

H0R5E:
I think I might have been calling the runner incorrectly. Can you try the
method here:

https://uk.mathworks.com/help/matlab/ref/matlab.unittest.testrunner-class.html#mw_8602d5ca-8748-4425-8fa1-9c2fdaef84e3

I.e.

runner = TestRunner.withTextOutput;
result = run(runner,suite);

T

H0R5E:
It's weird though, they should be equivalent ways of calling the same
thing. Do you mind if I take a look at this tomorrow morning?

Mat

akeeste:
Hi Mat,

Yes I'll take a look at this today and can discuss with you over the next few days.

akeeste:
@H0R5E 

I looked at the test plugins more today and discovered the one that seems to be recording diagnostic information in the testResult struct. I tested the following TestRunner options that contain the respective plugins:

    TestRunner.withNoPlugins;      % no plugins
    TestRunner.withDefaultPlugins; % ConciseProgressPlugin, DiagnosticsOutputPlugin, DiagnosticsRecordingPlugin
    TestRunner.withTextOutput;     % TestRunProgressPlugin, DiagnosticsOutputPlugin

DiagnosticsRecordingPlugin is the option that adds the diagnostic information to the testResult object. This is included as the default plugin when ``run(suites)`` is used and why we were seeing different output. The testRunner class also has a note that the 'withDefaultPlugins' option will change across versions. 

Given these items, I opted for the following setup, which will contain TestRunProgressPlugin, DiagnosticsOutputPlugin, DiagnosticsRecordingPlugin and CodeCoveragePlugin. This results in sufficient debugging information and allows code coverage in the future.

    runner = TestRunner.withTextOutput;
    runner.addPlugin(DiagnosticsRecordingPlugin);
    runner.addPlugin(CodeCoveragePlugin.forFolder('./source'));
.

H0R5E:
@akeeste, good work. It's kind of weird that the DiagnosticsRecordingPlugin gets added in certain situations but not others. I would say that is probably a bug.

Do you want me to review the PR?

akeeste:
@H0R5E Yes that would be great. There should not be much changed since you pushed to this branch. 

Last note--I am aware of one BEMIO test currently failing. The dev branch already contains the fix for this and the tests should all pass once the PR is merged.

tests/bemioTest.m
```diff
@@ -0,0 +1,109 @@
...
+
+            cd(fullfile(testCase.capytaineDir,'Cylinder'))
+            
+            hydro = struct();
+            hydro = Read_CAPYTAINE(hydro,'.\cylinder_full.nc');
+            hydro = Read_NEMOH(hydro,'..\..\NEMOH\Cylinder\');
+            hydro(end).body = {'cylinder_nemoh'};
+            hydro = Read_WAMIT(hydro,'..\..\WAMIT\Cylinder\cyl.out',[]);
+            hydro(end).body = {'cylinder_wamit'};
+            hydro = Combine_BEM(hydro);
```
H0R5E, line 64: It is better to skip this test as a known error, rather than leave it as a failure. The automated CI is always going to fail if we leave it like this. See https://github.com/akeeste/WEC-Sim/pull/8 for the approach to skip this test based on a known failure

tests/rotationTest.m
```diff
@@ -0,0 +1,71 @@
+classdef rotationTest < matlab.unittest.TestCase
+    
+    properties
+        testDir = '';
```
H0R5E, line 4: Again, no need for the semicolons here.

tests/runFromSimTest.m
```diff
@@ -0,0 +1,158 @@
+classdef runFromSimTest < matlab.unittest.TestCase
+    
+    properties
+        testDir = '';
```
H0R5E, line 4: Remove semicolons

wecSimTest.m
```diff
@@ -1,3 +1,51 @@
...
+    arguments
+        options.bemioTest = true
+        options.regressionTest = true
+        options.passiveYawTest = false
+        options.compilationTest = true
+        options.runFromSimTest = true
+        options.rotationTest = true
+    end
+    
+    import matlab.unittest.TestSuite;
```
H0R5E, line 12: No need for semicolons on import functions. See [here](https://uk.mathworks.com/help/matlab/ref/import.html)

wecSimTest.m
```diff
@@ -1,3 +1,51 @@
...
+    import matlab.unittest.Test;
+    import matlab.unittest.TestRunner;
+    import matlab.unittest.plugins.*;
```
H0R5E, line 15: This is a general programming comment, but it's normally bad style to use `import *` due to name space pollution (i.e. importing everything in the parent module). Now MATLAB is one huge namespace pollution, so it makes little difference, but if you were going to be programming in Python it's considered bad because you can't trace back the source of the functions / classes you are calling to the module you imported from. It would be better to add the plugins explicitly, e.g.

```
import matlab.unittest.plugins.DiagnosticsRecordingPlugin
import matlab.unittest.plugins.CodeCoveragePlugin
```

H0R5E (changes requested):
Hi @akeeste, overall I think this is a really good update to the testing framework. I've got a few minor style changes that I suggested in comments, but I also think there are three "blockers" to merging this right now. These are:

1. The test suite should not fail just because of the combine_bem test. I've added a solution to this in https://github.com/akeeste/WEC-Sim/pull/8
2. There is no documentation added to describe how to use the test suite. There should be something, even if its just a [docstring](https://uk.mathworks.com/help/matlab/matlab_prog/add-help-for-your-program.html) for the wecSimTest function.
3. I don't think we should skip the passive yaw tests by default. It was 200 seconds without and 260 seconds with on my machine and I think once you're in the minutes an extra one won't hurt. Explaining to users how to turn off the tests (in the doctring or whatever) is a better approach.

I'm happy to help out with any of these changes, so just give me a shout.

Mat

source/functions/wecSimMCR.m
```diff
@@ -96,7 +96,8 @@
             hydroData(ii) = body(ii).hydroData;
         end
     end
-end; clear imcr ans hydroData 
+end; clear imcr ans hydroData
+clear functions
```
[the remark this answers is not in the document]
akeeste: @H0R5E The purpose of this was to allow for copied files to be removed immediately after a run completed. It seems Simulink tries to hold on to certain files in the slprj folder, even after WEC-Sim is completed. This prevented cleaning up the copied files. 

The ``clear functions`` and the ``clear mex`` below closed the .mex file for acceleration and possibly an slprj file when paraview output is used. As we stopped the file copying altogether, I will clean up these extra clear statements.

source/functions/wecSimStopFunction.m
```diff
@@ -38,6 +39,7 @@
 clear ans table tout;
 toc
 diary off
+fclose('all');
```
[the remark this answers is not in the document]
akeeste: See comment above. I believe there was a file in the slprj folder that prevents a case directory from being deleted after the run completes.

tests/bemioTest.m
```diff
@@ -0,0 +1,109 @@
+classdef bemioTest < matlab.unittest.TestCase
+    
+    properties
+        wsDir = '';
```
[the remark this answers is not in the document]
akeeste: Thanks for that note! I'll make these changes.

wecSimTest.m
akeeste: Fair enough, I'll change this. As you say, its better practice to avoid the ``import *`` even if it doesn't make much difference in MATLAB.

tests/bemioTest.m
akeeste: I agree the test should be skipped if the fix is not already present. However since this PR is only merging into dev, the fix will be present and the test will pass as soon as the PR is merged. To simplify this I pulled dev into this branch so that all tests are passing before the merge.

H0R5E: Oh, I see. Shouldn't we be putting this into master as well?

akeeste: I assumed this wouldn't happen until the entire dev branch is merged into master and a new release is tagged. I believe v4.3 was planned for the near future.

H0R5E:
Offered an improvement to the docstring in https://github.com/akeeste/WEC-Sim/pull/9

H0R5E (approved):
Looks great. Please go ahead and merge it (I would recommend a squash and merge).

PR #627, opened 2021-06-25
Startup cleanup setup

salhus:
startup.m : This script initiates the WEC-Sim environment and adds necessary sub-folders to the MATLAB path.

cleanup.m : This script cleans up the path and removes all wec-sim folders from the path

Here is a text copy of the files,

STARTUP.M
% First written by : Gordon G. Parker, Michigan Technological University
% Modifed for WEC-Sim by: Sal Husain, sal.husain@nrel.gov
% Last modified : June 9th, 2021

% Root directory of this running .m file
projectRootDir = fileparts(mfilename('fullpath'));

% Add project directories to path
addpath(fullfile(projectRootDir,genpath('examples')),'-end');
addpath(fullfile(projectRootDir,genpath('source')),'-end');
addpath(fullfile(projectRootDir,genpath('tests')),'-end');
addpath(fullfile(projectRootDir,genpath('tutorials')),'-end');

CLEANUP.M
% First written by : Gordon G. Parker, Michigan Technological University
% Modifed for WEC-Sim by: Sal Husain, sal.husain@nrel.gov
% Last modified : June 9th, 2021

% Root directory of this running .m file
projectRootDir = fileparts(mfilename('fullpath'));

% Remove project directories from path
rmpath(fullfile(projectRootDir,genpath('examples')));
rmpath(fullfile(projectRootDir,genpath('source')));
rmpath(fullfile(projectRootDir,genpath('tests')));
rmpath(fullfile(projectRootDir,genpath('tutorials')));

% leave no trace...
clear projectRootDir

### DIFF
--- a/startup.m
+++ b/startup.m
@@ -1,0 +1,15 @@
+
+% First written by : Gordon G. Parker, Michigan Technological University
+% Modifed for WEC-Sim by: Sal Husain, sal.husain@nrel.gov
+% Last modified : June 9th, 2021
+
+
+% Root directory of this running .m file
+projectRootDir = fileparts(mfilename('fullpath'));
+
+% Add project directories to path
+addpath(fullfile(projectRootDir,genpath('examples')),'-end');
+addpath(fullfile(projectRootDir,genpath('source')),'-end');
+addpath(fullfile(projectRootDir,genpath('tests')),'-end');
+addpath(fullfile(projectRootDir,genpath('tutorials')),'-end');
+
--- a/cleanup.m
+++ b/cleanup.m
@@ -1,0 +1,17 @@
+
+% First written by : Gordon G. Parker, Michigan Technological University
+% Modifed for WEC-Sim by: Sal Husain, sal.husain@nrel.gov
+% Last modified : June 9th, 2021
+
+% Root directory of this running .m file
+projectRootDir = fileparts(mfilename('fullpath'));
+
+% Remove project directories from path
+rmpath(fullfile(projectRootDir,genpath('examples')));
+rmpath(fullfile(projectRootDir,genpath('source')));
+rmpath(fullfile(projectRootDir,genpath('tests')));
+rmpath(fullfile(projectRootDir,genpath('tutorials')));
+
+
+% leave no trace...
+clear projectRootDir


crobarcro:
I would **strongly** advise against distributing a file called startup.m that might clash with the users [matlab startup file](https://uk.mathworks.com/help/matlab/ref/startup.html)

startup.m
```diff
@@ -0,0 +1,15 @@
+
+% First written by : Gordon G. Parker, Michigan Technological University
+% Modifed for WEC-Sim by: Sal Husain, sal.husain@nrel.gov
+% Last modified : June 9th, 2021
+
+
+% Root directory of this running .m file
+projectRootDir = fileparts(mfilename('fullpath'));
```
nathanmtom, line 8: @salhus Thank you for cleaning up and resubmitting your pull request. 

In this file you create the `projectRootDir` variable which remains after the file executes while in the cleanup.m line 17 deletes this variable. I do not see any additional dependency on this path so can we also add the `clear` statement in the startup.m?

salhus: Yes I agree that can be cleared up. 
I usually use that variable in my scripts when I need to save variables programmatically but it is not needed per se.

nathanmtom:
@salhus Thank you for cleaning up and resubmitting your pull request. I agree with the above comment and you may want to relabel the file to not conflict with a users' individual `startup.m` which should be possible as I do not see see any other dependency based on the name of the file. 

I've noticed that the `startup.m` as written has different result if I use `Run` or `Run and Advance` options in the MATLAB Editor. When running the file with `Run` the right paths are set and I am able to move to the `examples\RM3` and run the example case. When running the file with `Run and Advance` some path appear but I am unable to run WECSim because it looks like the `\source\` folder is not a separate path item. I may not be up-to-date on MATLAB operations, but I would expect running both options should provide the same results. A similar effect happens, upon running the `cleanup.m` where not all paths are removed if you `Run and Advance` rather than `Run.` 

Appreciate your input and thoughts. 

Cheers, 
Nathan
National Renewable Energy Laboratory

salhus:
Hi @nathanmtom  
I could recreate the issue on my end. The MATLAB commands used here work on the global workspace so need the Run command. So according to the documentation of the addpath() command, even if the command is run inside a local function, the path gets added globally. I am suspecting this causes different behavior because the Run and Advance method does not execute the whole script..
I agree that either approach of running the script should not matter. I cannot find the exact reason why this is happening..

crobarcro:
The problem is this line probably

`projectRootDir = fileparts(mfilename('fullpath'));`

consider what this returns when you use 'Run and Advance' versus 'Run'

again, **don't call this file startup.m**

salhus:
Started a new PR with changed filenames and a way to package the installation process.

PR #632, opened 2021-07-02
Use the caller workspace when reading ptosim outputs

H0R5E:
This change allows the [PTOSim application examples](https://github.com/WEC-Sim/WEC-Sim_Applications/tree/master/PTO-Sim/) to run in function and class based tests.

I am not sure if I've covered all the use cases where this change might cause an issue, but it works for me when running wecSim in the application directory (i.e. in [OSWEC_Hydraulic_PTO](https://github.com/WEC-Sim/WEC-Sim_Applications/tree/master/PTO-Sim/OSWEC/OSWEC_Hydraulic_PTO)) and it now also works for the class based tests in https://github.com/WEC-Sim/WEC-Sim_Applications/pull/7

I've put this PR against master in order to get the WEC-Sim_Applications suite passing against the master branch. I am pretty sure that it will also merge into dev without issue.

### DIFF
--- a/source/objects/ptoSimClass.m
+++ b/source/objects/ptoSimClass.m
@@ -136,7 +136,7 @@ classdef ptoSimClass<handle
                 for jj = 1:length(obj.(names{ii}))
                     for kk = 1:length(signals.(names{ii}))
                         try
-                            tmp = evalin('base',[names{ii} num2str(jj) '_out.signals.values(:,' num2str(kk) ')']);
+                            tmp = evalin('caller',[names{ii} num2str(jj) '_out.signals.values(:,' num2str(kk) ')']);
                             ptosimOutput.(names{ii})(jj).(signals.(names{ii}){kk}) = tmp;
                         end
                     end


akeeste:
@H0R5E Sorry if this was holding up the other applications tests...
I just tested this in the applications cases too and don't have any issues.

This function is not set-up well; it shouldn't depend on the calling workspace. Instead the output items should be function parameters. However to get this update in before v4.3 and get the tests running, let's leave this for a future cleanup item. I added it to one of the current project board items.

My only request is in a PR in your [fork](<https://github.com/H0R5E/WEC-Sim/pull/24>). This will clear the old output variables from the calling workspace. I didn't intend to open a branch on your fork to do this, but that's what GitHub automatically did when I proposed changes...

hossein-mehdipour:
Hello @H0R5E I had a problem in #646 and you suggested me to use the file in this page to solve my problem. It in fact did solve my problem, but now I think I have another problem which I'm not sure what is the source of it. I wanted to reinstall and redownload MATLAB & WEC-Sim to see if my problem will gets solved but now that I have seen you deleted your file, I am a little confused and I wanted to ask you if you have any idea about my problem.
So basically when I use WEC-Sim directly or when in a function, I get different results (this happens both in the cases with HPTO or without HPTO). And most of the time the direct method gives the right answer, but sometimes it's the other way around. I was wondering if you could help me solve my problem?

Sorry if I shouldn't have commented here.

akeeste:
@hossein-mehdipour Thanks for following up on this problem. As this PR is merged, let's continue this discussion in issue #700 that you opened.

Thanks,
Adam

PR #633, opened 2021-07-02
Adding template for developer issues

Jeffalo1:


kmruehl:
@Jeffalo1 this looks great, thank you! The only revisions is that we would actually like to pull this into master and use it right away. Can you amend this PR do that it merges with master? (NOTE: the workflow you've used is correct for 99% of PRs, but this one is an exemption since we want it in master).

Jeffalo1:
I don't think I'm able to change the head branch so I created a new PR and closing this one.

PR #637, opened 2021-07-07
Activate deactivate wecsim

salhus:
Explanation of commits:
1. activate_wecsim.m : Add WEC-Sim folders and sub-folders to the path,
2. deactivate_wecsim.m: Remove WEC-Sim folders and sub-folders to the path,
3. WEC-Sim.prj : Open this file and hit Build (Green Tick), this will package the  whole WEC-Sim folder as a MATLAB package. A guide in the form of a live-script is also embedded in the package.


nathanmtom:
@salhus Thank you for submitting the new pull request; however, it looks like you have also committed in relation to capytaine development. If possible, can your open a new pull request so that the only changed files are the files you have described in the start of this conversation.

PR #638, opened 2021-07-08
Activate_Deactivate

salhus:
Explanation of commits:

1. activate_wecsim.m : Add WEC-Sim folders and sub-folders to the path,
2. deactivate_wecsim.m: Remove WEC-Sim folders and sub-folders to the path.

*Fixed the bug where run and advance did not work*


salhus:
These scripts also assign a dump folder for the simulink generated supported files.
User needs to create an empty folder title 'work' to contain the slprj and other support files created by Simulink.

salhus:
UPDATE : The updated installation documentation has been committed. The scripts have cleaned up as discussed.

docs/user/getting_started.rst
```diff
@@ -155,3 +149,6 @@ Both the RM3 and the OSWEC examples (``$WECSIM/examples/OSWEC``) come ready-to-r
     Storage <https://git-lfs.github.com/>`_ , or run 
     ``$WECSIM/examples/RM3/hydroData/bemio.m`` to generate the correct 
     ``rm3.h5`` file.
+
+.. Note:: You can remove the WEC-Sim Source folder from MATLAB path by running the "deactivate_wecsim.m" script in the ``$WECSIM folder. 
+However, it advisable to let the Source stay in MATLAB path.
```
nathanmtom, line 154: suggested grammatical change `However, it is advisable`

docs/user/getting_started.rst
```diff
@@ -91,15 +91,9 @@ referred to as ``$WECSIM`` (e.g. ``C:/User/Documents/GitHub/WEC-Sim``).
 Step 1. Add WEC-Sim to MATLAB Path
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Open the ``$WECSIM/source/wecSimStartup.m`` file.
+Open the ``$WECSIM folder and run the "activate_wecsim.m" script.
```
nathanmtom, line 94: There is no mention of creating a `work` folder in the current directory. Therefore, if this pull request is accepted then the folder will need to be added so that users do not have to create themselves. Or the documentation will need a line explaining that a `work` folder needs to be added.

activate_wecsim.m
```diff
@@ -0,0 +1,8 @@
+% Root directory of this running .m file
+projectRootDir = pwd;
+% Add project directories to path
+addpath(fullfile(projectRootDir,genpath('source')),'-end');
+addpath(fullfile(projectRootDir,'work'),'-end');
```
nathanmtom, line 5: Is it possible to add a line of code to check if a 'work` folder already exists? Thus, if it gets deleted accidentally or if the WEC-Sim development team decides to not include a `work` folder in the Master code this will be generated.

nathanmtom:
@salhus Thank you for making the suggested changes and cleaning up the folders. I have left some comments in the committed files for your review and would appreciate your thoughts and comments. I do appreciate being able to run this activation script when I switch branches rather than having to repeatedly open, edit, and close the `startup.m` file. 

Were you able to successfully build the documentation to see how your changes to the `getting_started.rst` would be seen on the website? If you are not familiar with this process then follow the guidelines [**here**](https://github.com/WEC-Sim/WEC-Sim/blob/master/docs/README.md). I was not able to build from your pull and you may need to update your branch to the latest WEC-Sim version release to utilize changes in the build documents process. 

The WEC-Sim team will discuss during our next development meeting about pulling this into the master code.

activate_wecsim.m
salhus: @nathanmtom  I resolved this by the most recent commit. Now the script checks if the work folder exists, and if it does not, the activate_wecsim.m creates it.

docs/user/getting_started.rst
salhus: I automated this by the most recent commit.

docs/user/getting_started.rst
salhus: @nathanmtom  sorry about that. Thanks for spotting it.

salhus:
Hi @nathanmtom I checked the index.html and it looks fine to me. Thanks for the link to the sphinx installation.
I did however, need to configure the bibtek there to make it work. I thought I'd mention that.

nathanmtom:
@salhus Thanks, the reconfiguration of the bibtek file might be because of not having the latest version of WEC-Sim pulled into your branch. I can successfully build the documentation from the Master Branch. Thanks for implementing the other changes, we can add this to the agenda of the next WEC-Sim development meeting to discuss if this should be pulled into the `dev` branch.

salhus:
@nathanmtom  Alright, thanks for checking that. I am glad it looks okay now !

docs/user/getting_started.rst
```diff
@@ -85,15 +85,9 @@ referred to as ``$WECSIM`` (e.g. ``C:/User/Documents/GitHub/WEC-Sim``).
...
```
jtgrasb, line 88:
```suggestion
Open the ``$WECSIM`` folder and run the "activate_wecsim.m" script.
```

jtgrasb (commented):
I only made one small comment on syntax to fix, but I have some further questions.

1. Does the activate_wecsim.m function need to be run each time MATLAB is opened? It seems like this would be the case without the startup.m file and this may be confusing for users. If the startup.m file is no longer needed, then I think it should be deleted from the source folder.

2. Is the work folder meant for the user/developer to put their own work in? I think this is a good idea, but it may be useful to include a short workflow description in the docs (also include rerunning activate_wecsim.m in workflow). 

Overall, I think this is a good change, just wanted to clear up confusion on the implementation.

docs/user/getting_started.rst
salhus: @jtgrasb Thanks for the comments and the fixes. To answer your questions,
1. Yes, the user needs to run activate script when they start using wecsim each time. The advantage of this is that. it ensures that the source folder is correctly added to the path, and no other files with name shadows are. There is always the risk that the user has similar filenames such as cylinder or sphere for instance. using the activate/deactivate workflow will prevent file path issues,
2. The work folder is actually a file dump folder. The slrpj files and other supporting files created by Simulink are dumped there. 

So a typical workflow would be,
1. User activates wecsim. This action populates the path with the source code,
2. Now user is free to work anywhere else without affecting the wecsim folder.

Let me know if that makes sense

jtgrasb: Okay, this makes sense, thanks!

In that case, I would recommend including something in the documentation about running the activate_wecsim file each time wec-sim is started and the reason for the work folder as this is a change from the previous methods. Also, I think it would be confusing to still have the wecsimStartup file in the source folder so I would recommend getting rid of that unless we have a reason to keep it. 

Let me know if that all makes sense and any thoughts you have.

salhus: 1. I agree and updated the getting started doc accordingly.
2. @nathanmtom  I can do that but do you see any downside to removing that?

nathanmtom:
@salhus Can you check to see if the latest changes in the `Dev` branch has been pulled into this PR? Also it looks like there are some comments provided by @jtgrasb that have not been answered. I'd like to add this to the WEC-Sim team meeting this Wednesday to decide if further changes need to be made or we can incorporate the pull request.

docs/user/getting_started.rst
salhus: @nathanmtom  
@jtgrasb marked this as resolved. I am not confident about removing the the wecsimStartup file..

salhus:
@nathanmtom 
I updated the branch with the latest changes, and the comments by @jtgrasb are resolved.
Thanks for bringing it on for Wednesday meeting.

jtgrasb:
@salhus The activate and deactivate functions are working well on my end. I also think the documentation updates look good! When looking at the changed files, it appears that you changed the RM3 example Simulink model, added an input file for the RM3 example, and made some changes to the test cases. Can you explain what the intentions are of these changes?

salhus:
@jtgrasb  Thanks for checking it out. I was making a new run before the meeting, to check examples ran okay.. I removed those additions.
Also I fetched the dev branch.. Is there any other way to check for differences?

salhus:
This has been reponed as PR #685

PR #595, opened 2021-05-28
Fixes #459. Update ParaView macros for Paraview 5.9

cmichelenstrofer:
Update the ParaView _pressureGlyphs.py_ macro for the latest ParaView version (5.9.1). The issue was that the current version of ParaView is not backwards compatible. 

@kmruehl To Do:

- [ ] Create new .pvsm files for the _WEC-Sim_Applications/Paraview_Visualization_ cases using Paraview 5.9.1
- [x]  Test new macro on _WEC-Sim_Applications/Paraview_Visualization/RM3_MoorDyn_Viz_. Currently only tested on _WEC-Sim_Applications/Paraview_Visualization/OSWEC_NonLinear_Viz_. 
-  [x] Mention the supported ParaView version somewhere in the Documentation or website.

PR #642, opened 2021-07-10
Updating ParaView Doc

kmruehl:
This PR relates to https://github.com/WEC-Sim/WEC-Sim/pull/595 and updates the documentation for using ParaView to visualize WEC-Sim data. Merging into master since it's related to a bug-fox.

### DIFF
--- a/docs/_include/viz.rst
+++ b/docs/_include/viz.rst
@@ -1,54 +1,55 @@
 
 
-This section describes how to setup, output, and use Paraview files for visualizing a WEC-Sim simulation. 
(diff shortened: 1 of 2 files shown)


akeeste:
@kmruehl 

Can you also copy these changes into dev? The docs there will not catch these updates when we merge dev and release a new version. I think this will be easier to do one change at a time than all at once.

kmruehl:
@akeeste yes! I was acually just thinking the same thing. I'll put these doc updates into dev as well.

kmruehl:
@akeeste added the docs to dev in https://github.com/WEC-Sim/WEC-Sim/commit/9d385e8d2e87ea2cc9177c60c88115aaccb2cd29

PR #649, opened 2021-07-19
Updated applications links in docs

jtgrasb:
The links in the applications section of the documentation were leading to errors so I updated them to refer and link to the correct sections.

### DIFF
--- a/docs/_include/bemio.rst
+++ b/docs/_include/bemio.rst
@@ -110,7 +110,7 @@ BEMIO Functions
 	*Plot_BEMIO(hydro)*
 		* *hydro* – data structure
 
-
+.. _user-advanced-features-bemio-h5:
 
 BEMIO *hydro* Data Structure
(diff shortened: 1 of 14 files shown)


akeeste (changes requested):
@Jeffalo1 This PR looks good! Thanks for updating these URLs. I made a few small in line comments. 

My main request is to also change the rest of the documentation URLs to the ``:ref:`` format For example if I CTRL+F in all rst files for "wec-sim.github.io/WEC-Sim" I see one outdated link in overview.rst and several in release_notes.rst

jtgrasb:
@akeeste I fixed all the links that contained "wec-sim.github.io/WEC-Sim" so it should be done now. I also changed some other links before I realized an issue on my end so that is the reason for the extra commits/reverts. Let me know if there's any other issues you find.

akeeste (approved):
Thanks Jeff, all these links work now. Merging!

PR #657, opened 2021-08-01
Viz marker

salhus:
UPDATED: Sept. 11th. Saturday.

- Updated wave class to accommodate N, (N \in Non-negative integers) from markers,
- The Global reference block programmatically initiates and adds/deletes viz. blocks based on the number of markers (0 to 
N) defined by the user,
Here are the steps to enable and define visualization blocks,

1. Define an array with the first column defining the X coordinates, and the second column defining the corresponding Y coordinates. The User should define waves.markLoc = []. If no array is defined, there would be no markers and vice versa,
2.  Define marker style by defining waves.markStyle = []; % 1: Sphere, 2: Cube, 3: Frame.
3. Define marker size in pixels by defining waves.markSize    = [];   % Marker Size in Pixels

Here is an example. In this example a mesh of points is described using the meshgrid command and then  making it an array of X and Y coordinates using reshape(). 
.. Note: The user only needs to initiate the visualization and define an array of visualization blocks.

examples:
1. A simple example with markers in a diagonal layout
X = 2*[1;2;3;4;5];
Y = 2*[1;2;3;4;5];
waves.markLoc = [X,Y];
waves.markStyle         = 2;  % 1: Sphere, 2: Cube, 3: Frame.
waves.markSize          = 15; % Marker Size in Pixels

diagViz

2. An example with a square mesh of visualization blocks:

mrk = 10;
dmrk = 5;
[X,Y] = meshgrid(-mrk:dmrk:mrk,-mrk:dmrk:mrk);
waves.markLoc = [reshape(X,[],1),reshape(Y,[],1)]; % Marker Locations [X,Y]
clear('mrk','dmrk','X','Y')

waves.markStyle   = 3;    % 1: Sphere, 2: Cube, 3: Frame.
waves.markSize    = 20;   % Marker Size in Pixels

meshViz


nathanmtom:
@salhus Apologies for taking so long to review this pull request. I think this is great progress and my personal opinion is the spherical buoys makes the most sense for wave elevation indicators. First, there seems to be a conflict with the WECSim_Lib.slx which I assume is because of the pull request moving several WEC-Sim functions into scripts, which will need to be resolved prior to incorporating this pull request. But, this change may be incorporated after/if the WEC-Sim team decides to move forward with the sub-libraries pull request. 

A few comments and questions to consider in the meantime:

1. Are the size of the sphere and cube fixed or does the user have the ability to define the characteristic length? In the commits this seems fixed which may be a problem for much smaller or larger devices than the OSWEC. If possible can the user input a value similar to how we define the thickness of the sea surface and the seabed `simu.domainSize`. 
2. Are the updates you have made flexible so a user can set greater than 9 wave probes? For example can the wave gauge be a matrix of x,y positions and the number of columns correspond to the number of wave probes the user wants? A user may want quite a few wave probes for some visualization purposes and would like to know if there is a limit that will need to be included in the WEC-Sim documentation if there is a limit. 
3. Does the calculation of the wave profile account for wave spreading? The calculation of the wave elevation does appear to account for the wave directionality, but not sure about spreading. If you can check that would be helpful as I know we need to update the Morison Element calculation to include spreading.

salhus:
Hi @nathanmtom ,
Thanks for the detailed review !
To answer your questions,
1. The size is defined by Simscape in pixels, so they relatively maintain their size upon zooming in/out. I can make that size parameter available as an option for the user, and would default to a reasonable size if not specified by the user.
2. I am using the wave gauges property to animate the markers. At the moment 9 is the upper limit. I was tempted to add a for loop around the method that calculates the wave gauge data.
3. I believe it does. I made changes in the wave class and added wave gauges in regular waves, and irregular waves methods for wave gauges. So if the wave gauges account for wave spreading, the markers should be able to track it. I will brush up on your more question further, and will update here...

examples/OSWEC/wecSimInputFile.m
```diff
@@ -11,21 +11,22 @@
...
+waves = waveClass('regular');           % Initialize Wave Class and Specify Type                                 
+waves.H = 2.5;                          % Wave Height [m]
+waves.T = 8;                            % Wave Period [s]
 
 % Irregular Waves using PM Spectrum with Directionality 
-waves = waveClass('irregular');         % Initialize Wave Class and Specify Type
-waves.H = 2.5;                          % Significant Wave Height [m]
-waves.T = 8;                            % Peak Period [s]
-waves.spectrumType = 'PM';              % Specify Spectrum Type
-waves.waveDir = [0,30,90];              % Wave Directionality [deg]
```
nathanmtom, line 27: @salhus If possible can you run the OSWEC example with lines 27 and 28 uncommented, or rather can you run with the directional spreading active? This would help to see how the balls would float with a more complex environment and answer our question if directional spreading is being used in the calculation of the wave gauge wave elevation.

salhus: irr_wave_with_spreading
@nathanmtom Thanks for pointing me there.. It seemed to work actually..

salhus: @nathanmtom Forgot to write in the last comment. I commented out the regular wave code and uncommented lines 23 - 29. Line 29 sets up the spreading, so seems like it did work..

salhus:
@nathanmtom
Hi Nathan,
I cleaned up the files so that only the input files are modified to enable visualization.   I could comment those out if the PR is ready to be adapted.
I am still trying to think a good way to programmatically set marker shape...

salhus:
@nathanmtom  
Hi Nathan,

I think I got all the desired things sorted.

The user can now,
1. Enable visualization by just defining the location matrix. If no locations defined, then no markers and vice versa,
2. User can define the marker shape and size from the input script,
3. The variable names will avoid confusion, 
such that, waves.markLoc : Marker Location matrix,
waves.markStyle : Choice of Marker Style between Sphere, Cube, and Frame,
waves.makeSize: Marker size in pixels.

nathanmtom:
@salhus This PR looks like it came in before we started doing automatic tests within WEC-Sim. I wanted to check to see if you have `runWECSimTest.m` on your pull request? Also, you will want to run the same test cases in the WEC-Sim Applications repository.

kmruehl:
@salhus please add to the docs, and we'll need to resolve the library conflict prior to merge

PR #664, opened 2021-08-06
Fix link tag address in gh-pages redirect html page

H0R5E:
This isn't super important because the link tag here is more for information than anything practical. This is just correcting an error in the given link address.


H0R5E:
Sorry, I have created a local branch by accident. I wish Github wouldn't do this...

PR #665, opened 2021-08-06
Fix link tag address in gh-pages redirect page

H0R5E:
Not super important as the link tag is just for SEO, but it's probably best if we fix it for correct Google results.


kmruehl (approved):
Thank you @H0R5E

PR #666, opened 2021-08-06
Update to the Morison Element Implementation to Allow Activation per Body Rather than Globally

nathanmtom:
Update on August 6, 2021: Updated based on the recent merge of #661 into the development branch.

--

Currently, WEC-Sim applies the implementation globally across all bodies regardless if a Morison Element is defined for a body or not. Discussions within the development team led to decision to adjust the code to allow a Morison Element to be active for one or more bodies while inactive for the remaining bodies. This is expected to improve computational time for irregular seas by bypassing the calculation of the local flow velocity for bodies that do not include a Morison Element.

The documentation has been updated to reflect these latest changes.

All cases passed when running wecSimTest.m.


source/functions/wecSimInitialize.m
```diff
@@ -273,40 +278,47 @@
...
-            %
-            [rgME,~] = size(body(ii).morisonElement.rgME);
-            %
-            for jj = 1:rgME
-                if true(isfinite(body(ii).morisonElement.z(jj,:))) == true
-                    warning(['"body.morisonElement.z" is not used for "simu.morisonElement = 1. Check body ',num2str(ii),' element ',num2str(jj)])
+% Check for morisonElement inputs for simu.morisonElement == 1 || simu.morisonElement == 2
+if length(simu.morisonElement) == 1 && simu.morisonElement == 0
+    warning(['"simu.morisonElement" needs to be a [n x 1] column vector where n is the number of bodies. WEC-Sim has assumed no bodies have active Morison Elements.'])
+    simu.morisonElement = zeros(length(body(1,:)),1);
```
kmruehl, line 284: Is there a reason why we need to initialize the ME in the simulation class? With this revision it seems to me that we should just remove this propoerty from the simulation class, and allow users to define the ME for each body using e.g. `body(#).morisonElement.option`, where the default is `0`, but users can specify `body(#).morisonElement.option = 1` or `2`. Thoughts?

kmruehl (commented):
@nathanmtom are you also planning to update the applications repo with an example using this new feature?

source/functions/wecSimInitialize.m
nathanmtom: @kmruehl If we move to the body class then this update will not be backwards compatible and my initial thought was to minimize this effect; however, per your pull request in #667, if we are planning to make other non backwards compatible changes then these should probably be all coordinated for a larger release and I can update to reflect these changes.

I'd also like to revisit the definitions between `body` and `simu` classes as if it is a simulation detail then perhaps we keep within `simu` rather than transfer to `body`. Because other simulations, such as nonlinear hydro or even the nonlinear yaw, should then also be updated to be in the `body` class to be consistent. 

We can use one of our next group meetings to discuss further.

nathanmtom:
@kmruehl I can make a new applications case; however, the only applications case that uses the morison element option is a single body so the changes made by this PR will not effect the simulation. A new applications case could be made with multiple bodies, but this could also be addressed by adding a commented line in the `wecSimInputFile.m` to indicate that with multiple bodies the chosen `simu` or `body` property would need to be a vector rather than a scalar.

nathanmtom:
@kmruehl I also wanted to push this PR to mostly put it on the radar of the WEC-Sim team. If we are satisfied with the implementation I would also like to implement this for the NL Hydro option as it will be a fairly easy addition.

source/functions/wecSimInitialize.m
kmruehl: I agree, let's discuss during our next team meeting. I think we should target a major release that includes some of these revisions that are not backward compatible.

kmruehl:
I think making this change for NL Hydro would be a great revision! Neither of those features should be universally defined in the simulationClass, it makes much more sense to have them in the bodyClass.

PR #667, opened 2021-08-06
Adding sub-libraries for each class

kmruehl:
This is the initial effort to break the library into sub-libraries. If approved, I can work on breaking the library into even more sub-libraries (e.g. for force calcs), is in progress here: https://github.com/kmruehl/WEC-Sim/tree/feature_subsublibraries

Changes include:
- Splitting library into sub-libraries for each class (e.g. body, pto, mooring, etc)
- Updating examples and test files to reflect the new library structure... aka the library change is **not** backwards compatible
- moved `spectralMoment` function to source directory so that test can be run in case directories


akeeste:
This will definitely make tracking library changes much easier. I like the implementation but think Simulink has some options that will maintain more backwards compatibility. Have you looked at converting the Body Elements, Constraints, etc into [referenced subsystems](https://www.mathworks.com/help/simulink/ug/referenced-subsystem-1.html)? This method will pull the Body Elements content directly from a different .slx file, instead of having the block open the file directly. I experimented with it a little bit and it may not require rebuilding all old models.

PR #672, opened 2021-08-10
Add MATLAB R2020a to CI tests

H0R5E:
Adds a new job to tests in master and dev branch (once merged) to test code against both latest and R2020a versions of MATLAB.

### DIFF
--- a/.github/workflows/run-tests-dev.yml
+++ b/.github/workflows/run-tests-dev.yml
@@ -10,7 +10,7 @@ on:
       - dev
 jobs:
   test_latest:
-    name: Run Latest MATLAB
+    name: MATLAB Latest 
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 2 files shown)


H0R5E:
There's certainly a better way to do this with variables rather than repeated code.

kmruehl:
@H0R5E I'm going to close this one

H0R5E:
@kmruehl why?

akeeste:
@kmruehl I don't think Mat meant this as a demo PR. This is a real update to the code.

kmruehl:
@H0R5E sorry, I thought this was from your demo yesterday

H0R5E:
@kmruehl no problem! Thought we might as well use it. Do you want to test the versions in between as well? I guess that's only 2020b right now.

PR #654, opened 2021-07-28
Moving MATLAB Functions from Simulink to m files

jtgrasb:
I went through and found all MATLAB function blocks within the WEC-Sim blocks and changed them to call an m file of the same name that completed the desired function. I located all of these functions within a new folder called "FunctionsCalledbySimBlocks" within the sources/functions directory. This will allow for easier access to the functions as well as better tracking of their changes.

PR #675, opened 2021-08-11
adding spherical constraint, cable block, and cable ex

dforbush2:
This PR consists of 3 new additions: 
- 1 DOF cable block
- 3 DOF cable block
- Spherical constraint

Significant modifications are to:
- wecSim.m
- wecSim_lib.slx (only the added blocks)
- responseClass
- simulationClass
- postprocess

Added files:
- cableClass

The 3 DOF cable uses a modification of the spherical constraint: the included block is distinct from the /constraint/spherical constraint.

To facilitate testing, I have also added ./examples/cable which runs a quick example, but this is intended to be deleted upon PR approval and re-uploaded to the applications repo.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ jobs:
   include:
   - stage: pages
     install:
-    - pip install "docutils<0.17" sphinxcontrib-matlabdomain sphinxext-remoteliteralinclude sphinx-multiversion sphinxcontrib-bibtex sphinx_rtd_theme
+    - pip install "docutils<0.17" sphinxcontrib-matlabdomain sphinxext-remoteliteralinclude sphinx-multiversion sphinxcontrib-bibtex "sphinx_rtd_theme<1"
     script:
     # Fix git repo so all branches can be seen by sphinx-multiversion
     - git config --replace-all remote.origin.fetch +refs/heads/*:refs/remotes/origin/*


akeeste:
@dforbush2 This PR is showing a lot of updates that are actually trying to change files back to a point in v4.2 (wecSim.m, simulationClass.m, responseClass.m line 56-59). Can you update your branch to avoid these extra changes? In the meantime I'll review the cable implementation.

source/objects/cableClass.m
```diff
@@ -0,0 +1,283 @@
...
+            linDisp = newCoord-rotloc1;
+            obj.rotinitDisp.initLinDisp1= linDisp + addLinDisp;
+            %
+            rotloc2 = obj.rotloc2;
+            relCoord = rotloc2 - x_rot;
+            rotatedRelCoord = rotateXYZ(relCoord,ax_rot,ang_rot);
+            newCoord = rotatedRelCoord + x_rot;
+            linDisp = newCoord-rotloc2;
+            obj.rotinitDisp.initLinDisp2= linDisp + addLinDisp;
+        end
```
akeeste, line 196: Use updated format of the setInitDisp() method in the body, pto, constraint, mooring classes.

source/objects/cableClass.m
```diff
@@ -0,0 +1,283 @@
...
+    % It is suggested that the mooringClass be used for connections between
+    % the bodies and global reference frame
+    %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+    properties (SetAccess = 'public', GetAccess = 'public')%input file
+        name                    = 'NOT DEFINED'
+        k_0                     = 0                                             % (`float`) Linear PTO stiffness coefficient. Default = `0`.
+        c_0                     = 0                                             % (`float`) Linear PTO damping coefficient. Default = `0`.                                             % (`float`) Rotational PTO cable-to-body coupling damping coefficient. Default = `0`.
+        K                       = 0
+        C                       = 0
+        L0                      = 0
```
akeeste, line 16: Remove equilibrium length, and have users define ``preTension`` as they would in the ``ptoClass`` or ``mooringClass``

source/objects/cableClass.m
```diff
@@ -0,0 +1,283 @@
...
+            'upperLimitTransitionRegionWidth', 1e-4, ... % (`float`) Define upper limit transition region, over which spring and damping values ramp up to full values. Increase for stability. m or deg. ``Default = 1e-4``
+            'lowerLimitSpecify', 'off',...        % Initialize lower stroke limit. ``  `on` or `off` `` Default = ``off``.
+            'lowerLimitBound', -1, ...            % (`float`) Define lower stroke limit in m or deg. Only active if `lowerLimitSpecify` is `on` ``   `` Default = ``-1``.
+            'lowerLimitStiffness', 1e6, ...       % (`float`) Define lower limit spring stiffness, N/m or N-m/deg.  `` Default = ``1e6``.
+            'lowerLimitDamping', 1e3, ...         % (`float`) Define lower limit damping, N/m/s or N-m/deg/s.  `` Default = ``1e3``.
+            'lowerLimitTransitionRegionWidth', 1e-4) % (`float`) Define lower limit transition region, over which spring and damping values ramp up to full values. Increase for stability. m or deg. ``Default = 1e-4``
+        geometryFile            ='geometry/sphere_out.stl'; % geometry file
+        bodyMass                =1;                                             % (`float`) mass in kg, default 1
```
akeeste, line 52: call `mass` similar to ``bodyClass``. Might make more sense to users to define the entire cable mass and then give each drag body mass/2 in Simulink behind the scenes.

source/objects/cableClass.m
```diff
@@ -0,0 +1,283 @@
...
+        viz               = struct(...                       %
+            'color', [1 0 0.5], ...                            %
+            'opacity', 1)                                    % Structure defining visualization properties in either SimScape or Paraview. ``color`` (`3x1 float vector`) is defined as the body visualization color, Default = [``1 1 0``]. ``opacity`` (`integer`) is defined as the body opacity, Default = ``1``.
+        bodyparaview      = 1;                               % (`integer`) Flag for visualisation in Paraview either 0 (no) or 1 (yes). Default = ``1`` since only called in paraview.
+        bodyInertia             = [ 1 1 1];                                     % (`1 x 3 float vector`) body inertia kg-m^2, default [1 1 1]
```
akeeste, line 57: --> ``momOfInertia`` see mass comment

source/objects/cableClass.m
```diff
@@ -0,0 +1,283 @@
...
+            obj.rotinitDisp.initLinDisp2= linDisp + addLinDisp;
+        end
+        
+        function checkFloat(obj, rho)
+            dispVol = obj.dispVol;
+            dispMass = obj.dispVol * rho;
+            if dispMass ~= obj.bodyMass;
+                warning('Cable dummy bodies are not neutrally buoyant. Check simu.rho and cable.bodyMass')
+            end
+        end
```
akeeste, line 204: Unless there's a good reason to allow the dummy bodies not to be neutrally buoyant, make dispVol private. Define the volume in dragprecalc as body.mass/simu.rho to enforce neutral buoyancy

akeeste (changes requested):
@dforbush2 I'm still going through some of the class and block details, but my main thought so far is on the end constraints. What is the reasoning for putting the end constraints (spherical or fixed) in the cable block? I suggest we make users define 2 extra constraints or ptos outside of the cableClass. The cable can be used in Simulink with the following connections:
Body -- Constraint/PTO -- Cable -- Constraint/PTO -- body

This will make the cableClass more like a body-pto combination, instead of body-ptox3, and will clean up the class properties. This would eliminate the need for separate 1DOF and 3DOF cables and allow users to create whatever end constraints they want. The top/bottom joint motion can be calculated with our motion sensor block. There will be a lot less class properties that users need to define.

sphere_out.stl -- since this is required in the cable's drag bodies and users can't change it, move to the source folder. However if users won't input an .stl file, we can turn the 'Body Properties' (mesh input) into a Simscape Inertia with a generic sphere shape. Then no .stl file is necessary.

WECSim_Lib.slx: 
- [ ] Add the cable on top of the v4.3 changes. Right now this can't run from Simulink. I can help set-up the callbacks for the cable once the implementation is finalized.
- [ ] move `calcCableTens()` to a .m file per #654  #661 
- [ ] set ``cable1_out`` decimation to ``round(simu.dtOut/simu.dt,0)`` per other to workspace blocks

RM3_cable.slx -- Link the 3dof Cable to the library. Maybe make this an application? We can likely move the RM3_runFromSimulink to an application case too.

source/objects/cableClass.m
dforbush2: Ok. I prefer keeping it somewhat clear, whether here or in the documentation, that this is the inertia/mass of the low-order lumped capacitance nodes rather than the cable itself, since the MOI of a flexible cable is complicated. That was my justification for not calling these cable.mass and cable.inertia, but this may not have been a better solution.

source/objects/cableClass.m
dforbush2: Treating this like a low-order lumped capacitance model of the cable, there is a good reason to retain this functionality. I think 90% of users will want it neutrally buoyant, so I've made this the default, but I prefer to not bury the property.

source/objects/cableClass.m
dforbush2: I can do that.

source/objects/cableClass.m
dforbush2: I strongly prefer having equilibrium length since this is the most physically meaningful quantity that one might have readily available to model a cable coupling two bodies, and is the critically important figure in determining whether or not the cable is slack. I am open to adding a preTension argument, but note that this will really complicate the mask functions for the dummy body positions. After the last group conversation I ended up chasing myself in circles thinking about which was the more intuitive option to include and kept landing on equilibrium length. Since this block is intended to couple two dynamic bodies, not body-to-seabed, equilibrium length is the parameter that requires the fewest number of assumptions.

akeeste: Yes good point, the set-up makes L0 an important parameter. I don't have an issue with using it, aside from trying to be consistent with how PTO and Mooring are initialized. Maybe keep the equilibrium length variable, but make it private and calculate based on the combination of constraint locations and the pretension. Then only let users define pretension for consistency with the other class parameters.

source/objects/cableClass.m
akeeste: Okay I see your reasoning. I guess if I was a user implementing a cable in WEC-Sim for the first time, I wouldn't know about the dummy bodies being used behind the scenes. I would probably approach this with a total mass or mass/length of the cable, but might be confused if the input file requires the mass or inertia of half the cable.

source/objects/cableClass.m
akeeste: Okay this makes sense, let's keep it this way

source/objects/cableClass.m
dforbush2: I'm good with that, basically calculating equilibrium length from user-specified preTension and constraint locations, for consistency with other classes.

akeeste:
@dforbush2 @kmruehl a few more comments below. I didn't have time to write out all my thoughts before I'm off work next week, but I didn't want to hold up your development. I think it would be easier to discuss this live when I'm back rather then trading comments over github.

1. I'm having trouble envisioning what the 1 DOF cable physically represents. If the cable ends are fixed to the body without rotation, then the cable becomes a type of telescoping pendulum. What kind of cable would this represent? I can't picture the kind of real body that results in these constraints.

2. The cableClass.setL0() method uses rotloc1 and rotloc2 (spherical joints) to define the equilibrium length, but the simulink model uses the dummy drag body positions for the cable length. This is going to result in strange behavior if rotloc1 ~= cg1 or rotloc2 ~= cg2. Also with regards to the dummy bodies: I feel that the cable is currently halfway between two models:
    - simplified model that only takes the position of the spherical joints and applies stiffness/damping when in tension
    - a complex lumped capacitance model with a series of nodes (dummy drag bodies with drag, buoyancy, mass)

    This combination of the two above is making this difficult to understand. I think if we committed to one or the other (or had a separate cable block for each), this would be more clear.

3. Why does the translational PTO allow for bidirectional stiffness/damping? Seems counterintuitive.

akeeste:
@dforbush2 @kmruehl Just looked back on this in preparation for v4.4 release--
In addition to an advanced features section, we need to:
- [ ] merge with other dev library changes
- [ ] move the example to the applications repo and set-up a CI test there
- [ ] complete the Run from Simulink capability
- [ ] finish discussing naming/implementation
- [ ] any items from Kelley's review

I can do the run from Simulink capability, but need to completely hammer out the implementation before doing that (see items in previous comment). I think there's some renaming we can do to clarify things and that we can probably reduce the number of user-defined options. IMO the number of options makes the implementation a bit overwhelming.

Issue #336, opened 2020-02-06
MoorDyn on Linux


ttosdevin:
Hi,

I'm trying to run WEC-Sim with MoorDyn on a Linux machine and have followed the step suggested in issue #269 of changing line 88 in mooringClass.m to:
obj.moorDynInputRaw = textread('./Mooring/lines.txt', '%s', 'delimiter', '\n');

But I get the following error message:

 Running MoorDyn (v1.01.00C, 2016-04-20)
   Copyright (c) Matt Hall, licensed under GPL v3.
   Creating mooring system.  3 fairleads, 3 anchors, 3 connections.
   Error: unable to write file Line1.out
   Error: unable to write file Line2.out
   Error: unable to write file Line3.out
   Error: unable to write file Line4.out
   Warning from Catenary: Unstretched line length too large.
   Error: unable to write file Line5.out
   Warning from Catenary: Unstretched line length too large.
   Error: unable to write file Line6.out
   Warning from Catenary: Unstretched line length too large.
   Finalizing ICs using dynamic relaxation (5X normal drag)
    t = 1 s, tension at first fairlead is 77716.5 N    
    t = 2 s, tension at first fairlead is 1.01718e+06 N    
    t = 3 s, tension at first fairlead is 199041 N    
    t = 4 s, tension at first fairlead is 186692 N    
    t = 5 s, tension at first fairlead is 175286 N    
    t = 6 s, tension at first fairlead is 175517 N    
    t = 7 s, tension at first fairlead is 174200 N    
    t = 8 s, tension at first fairlead is 174005 N    
    t = 9 s, tension at first fairlead is 174790 N    
    t = 10 s, tension at first fairlead is 174934 N    
    t = 11 s, tension at first fairlead is 174871 N    
   Fairlead tensions converged to 0.1% after 11 seconds.        
Unable to write to output file 
Unable to write to output file 
Unable to write to output file 
Unable to write to output file 
Unable to write to output file 
Unable to write to output file   

I have tried this for 2 different WEC devices and on 2 separate Linux machines and I always get the above error (one of the cases is the RMS MoorDyn tutorial). The error appears to be from line 235 of the line.cpp MoorDyn file;
https://github.com/mattEhall/MoorDyn/blob/master/Line.cpp

Any help with this would be much appreciated.
Thanks,
Tom

yuyihsiang:
I noticed that you have "Mooring" instead if "mooring"

Not sure if that is the issue

[...]

Issue #269, opened 2018-09-10
Run WEC-Sim with MoorDyn in Linux


nbcm:
I have tried to run examples RM3_MoorDyn but it did not works well in my machine (CentOS 7). I use the rm3.h5 file which built from RM3 case in tutorials folder. 

In the first try, I got following error information:
 Running MoorDyn (v1.01.00C, 2016-04-20)
   Copyright (c) Matt Hall, licensed under GPL v3.
Error: unable to open lines.txt file

Then I back to check the source code of MoorDyn and it seems the input file in the latest version should be in folder named 'Mooring' instead of 'mooring' (MoorDyn.cpp, line 287).

Then I change the folder's name to 'Mooring' in 'RM3_MoorDyn' and run again. The new issue comes and the error information shows:

Simulating the WEC device defined in the SimMechanics model RM3MoorDyn.slx...   

 Running MoorDyn (v1.01.00C, 2016-04-20)
   Copyright (c) Matt Hall, licensed under GPL v3.
   Creating mooring system.  3 fairleads, 3 anchors, 3 connections.
   Warning from Catenary: Unstretched line length too large.
   Warning from Catenary: Unstretched line length too large.
   Warning from Catenary: Unstretched line length too large.
   Finalizing ICs using dynamic relaxation (5X normal drag)
    t = 1 s, tension at first fairlead is 77716.5 N    
    t = 2 s, tension at first fairlead is 1.01718e+06 N    
    t = 3 s, tension at first fairlead is 199041 N    
    t = 4 s, tension at first fairlead is 186692 N    
    t = 5 s, tension at first fairlead is 175286 N    
    t = 6 s, tension at first fairlead is 175517 N    
    t = 7 s, tension at first fairlead is 174200 N    
    t = 8 s, tension at first fairlead is 174005 N    
    t = 9 s, tension at first fairlead is 174790 N    
    t = 10 s, tension at first fairlead is 174934 N    
    t = 11 s, tension at first fairlead is 174871 N    
   Fairlead tensions converged to 0.1% after 11 seconds.        

Error using wecSim (line 257)
Error in 'RM3MoorDyn/MoorDyn': Initialization commands cannot be evaluated.
Caused by:
    Error using textread (line 159)
    File not found.

Is there any path setting incorrect in WEC-Sim? Anyone has the same issue and can help?

yuyihsiang:
Hi,

[...]

PR #677, opened 2021-08-12
Make mooringClass.moorDynInput case insensitive

H0R5E:
This PR makes the `mooringClass.moorDynInput` method insensitive to the case of folders and files required by MoorDyn.

Currently, MoorDyn requires that the `Mooring` folder (which contains the `lines.txt` file) should have a capital Em as shown by [this line](https://github.com/mattEhall/MoorDyn/blob/ea5ccf8c050f359cdd13a0a8b6f5cc5f3158405c/MoorDyn.cpp#L287) in the [MoorDyn](https://github.com/mattEhall/MoorDyn) source code. Because Linux is case sensitive, changing the folder name to have a capital em brakes the `moorDynInput` method of the `mooringClass` class and, therefore, this change is required for the [WEC-Sim_Applications](https://github.com/WEC-Sim/WEC-Sim_Applications) tests to run correctly on GitHub Actions.

This PR makes the following changes:

* Add `caseInsenstiveDir` function to match file objects regardless of case.
* Changes `mooringClass.moorDynInput` to use `caseInsenstiveDir` to locate the `Mooring` folder and `lines.txt`.
* Add tests of both `caseInsenstiveDir` and `mooringClass.moorDynInput` in new `objects` and `functions` sub-directories of the `tests` directory.
* Fixes bad line endings (`^M`) in the `mooringClass.m` file


akeeste:
@H0R5E The linked test result doesn't show an error with MoorDyn that I can see, but I don't see an issue with updating this to match MoorDyn's syntax. I can confirm the MoorDyn application works for me on Windows with the folder name capitalized.

H0R5E:
Hi @akeeste, sorry my explanation was a bit brief. The problem occurs on Linux, because Linux uses case sensitive file names. For the [Moordyn application example](https://github.com/WEC-Sim/WEC-Sim_Applications/tree/master/Mooring/MoorDyn), MoorDyn does not like that the `mooring` folder is not capitalized, resulting in:

```
Running MoorDyn (v1.01.03C, 2021-03-09)
     Copyright (c) Matt Hall and NREL, licensed under GPL v3.
  Error: unable to open lines.txt file
```

which breaks the test case. Renaming this folder to `Mooring` fixes MoorDyn, but then WEC-Sim complains:

```
  ================================================================================
  Error occurred in TestMooring/testMoorDyn and it did not run to completion.
      ---------
      Error ID:
      ---------
      'Simulink:Masking:Bad_Init_Commands'
      --------------
      Error Details:
      --------------
      Error using wecSim (line 45)
      Error in 'RM3MoorDyn/MoorDyn': Failed to evaluate mask initialization
      commands.
      
      Error in TestMooring/testMoorDyn (line 76)
                  wecSim
      
      Caused by:
          Error using wecSim (line 45)
          File not found.
  ================================================================================
```

So a fix needs added to WEC-Sim as well. This is probably related to issue #336 & #269. A more robust solution on the WEC-Sim side might be to read the file name in a case-insensitive manner, to avoid issues if MoorDyn changes again.

H0R5E:
I'm trying something else with this.

H0R5E:
Closing in favour of #682

Issue #656, opened 2021-07-29
[Developer Issue] Output time step with variable step solvers


akeeste:
**Note: italicized text below is include as an example and should be updated before submission. If you feel any section is not applicable to your request, please replace with 'N/A' rather than delete a section.**

**Project**
WEC-Sim development

**Describe the issue**
The output time step ``simu.dtOut`` is not robust in variable step WEC-Sim simulations. Library ``To Workspace`` blocks currently use a [decimation ](https://www.mathworks.com/help/simulink/slref/toworkspace.html) of ``round(simu.dtOut/simu.dt,0)``. For fixed step solvers this is fine, but with variable step solvers this becomes inaccurate. For example, if simu.dt = 1, simu.dtOut=5, WEC-Sim outputs every 5th data point. However when using a variable step solver, the actual time step can become 0.01. Then the output time step is effectively 0.05, not 5 seconds.

This can be fixed by changing all ``To Workspace`` blocks in the library to use the ``SampleTime = simu.dtOut``

dforbush2:
see also global model settings: 
![image](https://user-images.githubusercontent.com/30122876/128199774-0cba0bf7-0d03-43fb-8c14-fd8ee847d049.png)

not sure which is the easier/better way

akeeste:
@dforbush2 Thank for pointing this out. I'll look at loading the simulink file with the global output time step instead of changing each individual ``To Workspace`` block.

PR #678, opened 2021-08-12
Wave visualization as a post-processing function

jtgrasb:
I created a function in the response class which creates a visualization of the waves and body reactions and saves it as either a video file or gif. I've attached examples below of the RM3 with regular waves and the OSWEC with irregular, directional waves.

wave_visualization_rm3

wave_visualization_oswec

The function can be run by adding "output.plotWaves(simu.dt,1,simu.g,[-100 100 -100 100 -40 15],body,waves,1);" to the userDefinedFunctions.


source/objects/responseClass.m
```diff
@@ -318,6 +318,141 @@ function plotForces(obj,bodyNum,comp)
...
+            %         (time step factor) number of simulation time steps
+            %         per video frame (higher number decreases computation
+            %         time)
+            %
+            %     g : float
+            %         acceleration due to gravity (simu.g)
+            %
+            %     axis_lims : 1x6 float matrix
+            %         upper and lower limits to the figure
+            %         [xmin xmax ymin ymax zmin zmax]
```
akeeste, line 340: If we require users to input the simu object, then we can avoid taking it, g and axis_lims as parameters. Instead simu.g, simu.domain and simu.dt can be used directly. Also I would recommend renaming some variables as I'll note below. If possible keep variable names the same as the parameter that will likely be input to keep things clear. Also use camel case per WEC-Sim's convention.

source/objects/responseClass.m
```diff
@@ -318,6 +318,141 @@ function plotForces(obj,bodyNum,comp)
             clear t FT FE FRD FR FV FM i
         end
         
+        function plotWaves(obj,it,tsf,g,axis_lims,body,waves,save)
+            % This method plots the wave elevation and body geometry over
+            % time to visualize the waves and response
+            %
+            % Parameters
+            % ------------
+            %     it : float
```
akeeste, line 327: Replace with ``simu.dtOut`` throughout the function. Note that the wave elevation is a pre-processing step so simu.dtOut is not used, it is instead sampled at simu.dt. However, the body's position response will have a sample time of simu.dtOut.  If simu.dtOut~=simu.dt, the function will break.

source/objects/responseClass.m
```diff
@@ -318,6 +318,141 @@ function plotForces(obj,bodyNum,comp)
...
+            %     save : integer
+            %         0 = avi video file, 1 = gif file
+            %     
+            
+            % Set time vector
+            t = obj.wave.time(1:tsf:end,1);
+            
+            % Create grid using provided x and y coordinates
+            x = linspace(axis_lims(1),axis_lims(2),200);
+            y = linspace(axis_lims(3),axis_lims(4),200);
```
akeeste, line 357: Consider using:
``x = linspace(-simu.domain/2, simu.domain/2, 200);``
``y = linspace(-simu.domain/2, simu.domain/2, 200);``

source/objects/responseClass.m
```diff
@@ -318,6 +318,141 @@ function plotForces(obj,bodyNum,comp)
...
+            [X,Y] = meshgrid(x,y);
+
+            % Read in data for each body
+            for ibod=1:length(obj.bodies)
+                % Read and assign geometry data
+                read_bod = stlread(body(ibod).geometryFile);
```
akeeste, line 363: Camel case --> ``readBod``

source/objects/responseClass.m
akeeste, line 340: Some of these parameters are already defined in the simulation class for visualization, so you might consider requiring ``simu`` as an input instead of ``it``, `g`, `axis_lims`. Also I recommended changing your variable naming to be more informative and in camel case. I.e. ``tsf`` --> ``timesPerFrame``, ``axis_lims`` --> ``axisLimits`` (or use ``simu.domainSize`` for this)

source/objects/responseClass.m
akeeste, line 363: camel case --> ``readBodyMesh``

source/objects/responseClass.m
```diff
@@ -318,6 +318,141 @@ function plotForces(obj,bodyNum,comp)
...
+                bod(ibod).Points = read_bod.Points;
```
akeeste, line 364: use a more informative name. maybe ``bodyMesh``

source/objects/responseClass.m
akeeste: I'll make a PR with some of these changes directly into this file so that if you agree they are easy to pull in

akeeste (changes requested):
See inline comments for suggested changes. I think this works well but there are a few changes that can make it more robust and easier to read. Also, I would consider including a colored plane for the seafloor and force the axes to have the same aspect ratio (see ``daspect([])`` or ``axis equal`` commands). If not, then the device can appear stretched depending on the domain size input

jtgrasb:
@akeeste Thanks for the suggestions! I made the requested changes and I think it is a lot simpler and easier to read now. One small issue I ran into is that because I used simu.domainSize/2 for the axes, the seafloor is not on the figure for the RM3 because it's at -200 m. Should I just make the axes go to the full simu.domainSize?

akeeste:
@jtgrasb For the lower z-axis limit, you could instead use ``-waves.waterDepth``. This is the parameter that currently determines the depth  shown in the Mechanics Explorer. 

The upper z-axis limit doesn't need to be very high. A couple thoughts:
- Easy solution - choose something relatively arbitrary (15m? 20m?) and let users optionally input the z-limits to overwrite the [default value](https://www.mathworks.com/help/matlab/ref/arguments.html). 
- Alternative - use the maximum z position of the bodies and calculate the highest point that any mesh reaches. The maximum z-axis could then be that point + ~5m.

It might be worthwhile letting users override the default limits regardless of method chosen.

jtgrasb:
@akeeste I like the idea of having options instead of needing to specify some of the inputs. I made the default z maximum 10 and allow the user to input the axis limits, timesPerFrame, and saveSetting as optional inputs.

akeeste:
@jtgrasb Great work! I made a few more minor changes:
- use ``options.axisLimits`` for ``X,Y`` grids
- allow different ``simu.dtOut`` and ``simu.dt``
- Change default z axis maximum to -999. By default this is overwritten with the ``maxHeight`` of the body if the user does not input another option
- move time annotation to the 2nd title line so that it cannot lie on top of the title box (this happened to me depending on the figure size upon opening)

I wasn't sure if you will have time to review before your semester starts so I went ahead and committed directly to your branch. Let me know if you disagree with anything

Adam

akeeste:
@kmruehl @jtgrasb My only other concern is that this function will not work with the variable step solver. However this relates to a larger problem of how we use ``simu.maxIt`` to calculate the ``waves.waveAmpTime`` before the simulation, instead of using the actual simulation sample time or the output sample time. I am okay merging this PR in and addressing the variable step solver problems later on when I solve issue #656

kmruehl:
> 
> 
> @akeeste I like the idea of having options instead of needing to specify some of the inputs. I made the default z maximum 10 and allow the user to input the axis limits, timesPerFrame, and saveSetting as optional inputs.

@jtgrasb another option would be to make it a multiplier of `waves.H`, so it could be `4*waves.H` by default (or similar).

kmruehl:
> 
> 
> @kmruehl @jtgrasb My only other concern is that this function will not work with the variable step solver. However this relates to a larger problem of how we use `simu.maxIt` to calculate the `waves.waveAmpTime` before the simulation, instead of using the actual simulation sample time or the output sample time. I am okay merging this PR in and addressing the variable step solver problems later on when I solve issue #656

@akeeste that sounds good to me. I think we should go ahead and merge it and then open an issue or project card related to the variable time-step issue with this feature.

kmruehl (commented):
@jtgrasb and @akeeste I made some minor revisions in this PR, feel free to merge or resolve them elsewhere: https://github.com/jtgrasb/WEC-Sim/pull/1
- added output.plotWaves to examples, we could comment this but provide it as an example
- resolved issues with examples files to use the wecSimInputFile instead of custom parameters

To Do
- [ ] add documentation

kmruehl (changes requested):
Please add documentation!

akeeste:
> Please add documentation!

Where are you thinking to add this? I don't see anything in the documentation about ``plotForces()`` or ``plotReponse()`` outside of the autogenerated API. Maybe we should start an advanced features section on the response class?

kmruehl:
@akeeste I think that's a good suggestion. The only reason we don't have docs for functions it because they were merged prior to our enforcement of the docs requirement. So if you'd like to add docs for them too... :)

akeeste (approved):

PR #682, opened 2021-08-16
Disable moorDynInput method

H0R5E:
This is a follow-up to #677, which simply disables the `mooringClass.moorDynInput` method. As far as I can see, `mooringClass.moorDynInput` and the `moorDynInputRaw` property have no functional usage. My testing also indicated that `mooringClass.moorDynInput` can end up racing for access of the `lines.txt` file with `MoorDyn` which is another source of failure. So the easiest thing to do, to me, is to stop it running. This also fixes the case sensitivity problem detailed below for Linux.

Currently, MoorDyn requires that the `Mooring` folder (which contains the `lines.txt` file) should have a capital Em as shown by [this line](https://github.com/mattEhall/MoorDyn/blob/ea5ccf8c050f359cdd13a0a8b6f5cc5f3158405c/MoorDyn.cpp#L287) in the [MoorDyn](https://github.com/mattEhall/MoorDyn) source code. Because Linux is case sensitive, changing the folder name to have a capital em brakes the `moorDynInput` method of the `mooringClass` class and, therefore, this change is required for the [WEC-Sim_Applications](https://github.com/WEC-Sim/WEC-Sim_Applications) tests to run correctly on GitHub Actions.

### DIFF
--- a/source/objects/mooringClass.m
+++ b/source/objects/mooringClass.m
@@ -104,7 +104,6 @@ classdef mooringClass<handle
 
         function obj = moorDynInput(obj)
             % Method to read MoorDyn input file
-            obj.moorDynInputRaw = textread('./mooring/lines.txt', '%s', 'delimiter', '\n');
         end
 
         function listInfo(obj)


akeeste:
I can confirm that I don't see ``mooring.moorDynInputRaw`` used anywhere in the code. The ``responseClass.loadMoorDyn()`` method is what actually reads and uses MoorDyn output, not ``mooring.moorDynInput()``. The applications repo works with this change.

To clean this up fully, we should remove ``mooring.moorDynInput()`` from the mooringClass definition and from the library mask initialization altogether. However since there are already 4 library PRs that will need merging together, I think removing this line is sufficient right now.

akeeste (approved):
PR approved. Only comment--
Why don't we actually just clean up this parameter completely? I have a commit that removes ``mooring.moorDynInputRaw`` and ``mooring.moorDynInput()`` from the class and library entirely. The other library PRs might not even have to deal with this since Jeff has moved mask initializations to .m files in dev. Even if they do, it will make a good demo of the matlab merge tool for the team.

If you agree, I can push the commit and merge this PR.

PR #696, opened 2021-09-06
Fix .gitignore and .gitattributes

srflp:
This PR removes bash scripts from the top of .gitignore and .gitattributes files. Those scripts were placed there probably by a mistake.
Syntax of .gitignore and .gitattributes does not support embedding bash scripts.

This PR is ready to merge


kmruehl (approved):
@srflp thanks

Issue #693, opened 2021-08-31
Biased Response for Free Decay Tests for OSWECs with an initial angle


salhus:
**Project**
OSWEC

**Describe the issue**

When conducting free-decay tests for OSWEC type WECs, if I give an initial angular displacement using setInitDisp method, the pitch response starts from 0 and not from the initial angle set. On closer inspection, the pitch response has a bias, and the crest to trough difference corresponds to the initial angular displacement.

NOTE: The output data object has the right data but the User Defined Functions script plots the data for pitch mode from 0, regardless of initial conditions

![free_decay_with_init_ang](https://user-images.githubusercontent.com/84348506/131556462-675c9087-6eb9-4bb8-91ed-880b62026cc6.jpg)
![free_decay_with_init_ang_zoomed](https://user-images.githubusercontent.com/84348506/131556465-3f7fafc1-7b1c-4bbe-87de-3133d49baef2.jpg)
![oswec_with_initial_ang](https://user-images.githubusercontent.com/84348506/131556467-2d482739-9b05-471c-9337-c1711c19d86a.gif)
[wecsim_input_file.txt](https://github.com/WEC-Sim/WEC-Sim/files/7086066/wecsim_input_file.txt)

nathanmtom:
@salhus Based on our discussion last week, I just wanted to clarify that this bias is from the `output.plotResponse(1,5)` function, correct? You mentioned that when looking at the `output.bodies(1).position(1,5)` the initial displacement is accurate. If you can confirm then we can look at this function to see if it needs to be updated.

salhus:
@nathanmtom  Yes, the output class has the accurate values, but the plotResponse(1,5) method plots biased results

jtgrasb:
@salhus I am having trouble replicating the issue. What exactly did you put in the input file to produce the initial rotation?

salhus:
Hi @jtgrasb  Sorry my bad,
here is the code for initial angular position I used,
intAng = deg2rad( #your angle here in deg  );
body(1).setInitDisp([0,0,-0.2352]-[0,0,-0.5], [0 1 0 intAng], [0 0 0]);

jtgrasb:
@salhus Thanks for the response. I was able to replicate your issue and find the issue within the plotResponse function. The function is subtracting the initial position from the position throughout the motion, producing the bias. 

pos=obj.bodies(bodyNum).position(:,comp) - obj.bodies(bodyNum).position(1,comp);

[...]

PR #698, opened 2021-09-10
Remove Bias from Angular mode reponses resolves issue #693

salhus:
This PR resolves the issue #693 .

This is done by adding an if statement that ensures the initial angular position is NOT subtracted from the angular mode responses.

### DIFF
--- a/source/objects/responseClass.m
+++ b/source/objects/responseClass.m
@@ -139,6 +139,7 @@ classdef responseClass<handle
             for ii = 1:length(bodiesOutput)
                 obj.bodies(ii).name = bodiesOutput(ii).name;
                 obj.bodies(ii).time = bodiesOutput(ii).time;
+%                 obj.bodies(ii).cg   = bodiesOutput(ii).cg;
                 for jj = 1:length(signals)
                     obj.bodies(ii).(signals{jj}) = bodiesOutput(ii).signals.values(:, (jj-1)*6+1:(jj-1)*6+6);
                 end
@@ -255,7 +256,7 @@ classdef responseClass<handle
             end
         end
 
-        function plotResponse(obj,bodyNum,comp,bodiesOutput)
+        function plotResponse(obj,bodyNum,comp)
             % This method plots the response of a body for a given DOF.
             %
             % Parameters
@@ -268,15 +269,9 @@ classdef responseClass<handle
             %     
             DOF = {'Surge','Sway','Heave','Roll','Pitch','Yaw'};
             t=obj.bodies(bodyNum).time;
-            if comp<4
-            pos=obj.bodies(bodyNum).position(:,comp) - [bodiesOutput(bodyNum).cg,0,0,0];
-            vel=obj.bodies(bodyNum).velocity(:,comp);
-            acc=obj.bodies(bodyNum).acceleration(:,comp);
-            else
             pos=obj.bodies(bodyNum).position(:,comp);
             vel=obj.bodies(bodyNum).velocity(:,comp);
             acc=obj.bodies(bodyNum).acceleration(:,comp);
-            end
             figure()
             plot(t,pos,'k-',...
                 t,vel,'b-',...
--- a/source/objects/responseClass.m
+++ b/source/objects/responseClass.m
@@ -268,9 +268,15 @@ classdef responseClass<handle
             %     
             DOF = {'Surge','Sway','Heave','Roll','Pitch','Yaw'};
             t=obj.bodies(bodyNum).time;
+            if comp<4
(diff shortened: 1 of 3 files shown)


source/objects/responseClass.m
```diff
@@ -268,9 +268,15 @@ function plotResponse(obj,bodyNum,comp)
             %     
             DOF = {'Surge','Sway','Heave','Roll','Pitch','Yaw'};
             t=obj.bodies(bodyNum).time;
+            if comp<4
             pos=obj.bodies(bodyNum).position(:,comp) - obj.bodies(bodyNum).position(1,comp);
```
nathanmtom, line 272: @salhus If we apply an initial linear displacement in the translational modes this will still cause the `pos` to be 0 at time = 0 and oscillate about the displacement value. Therefore, my suggestion is to subtract off the location of the center of gravity.

nathanmtom (changes requested):
Appreciate starting to address the issues with this function, but suggesting other changes to improve the ability to handle initial linear displacements.

source/objects/responseClass.m
salhus: @nathanmtom  Thanks for catching that.

I noticed that the bodies property, only stores the motion and forces. It doesn't have the cg information at  the moment.
Do you think I can modify that to have the cg information?

nathanmtom: Yes, you should be able to amend the function to pull in the body object to pull the center of gravity information.

salhus: @nathanmtom  Thanks for the recommendation. I implemented it that way in the latest commit.

salhus: @nathanmtom Sorry I deleted that comment when I realized I could do that, and didn't refresh the page to see your comment. But now i think it is doing just that.

salhus:
@nathanmtom  

I noticed the same behavior for linear modes as well. I did a free decay test, in the applications repo, and also tried it on rm3,
I gave an initial displacement,
body(1).initDisp.initLinDisp = [ 0 0 .5];    % Initial Displacement [m]
The cg of the body -.72

So plotResponse plots this,
image

While the ouput object has the correct plot,
image

However, when I remove the  term being subtracted, I get the correct plot. 
image

Which means, no term should be subtracted in any mode...

P.S.

If possible could you tell me how to bring the cg in the responseClass...
When I put a debug statement in reponseClass, I have all the classes as bodiesOut, simuOut.. I can't figure out,
1.  Why can't I access the objects that populated by the wecSimInput file. Seems like the wecSim.m clears those up..
2. If the bodiesOut etc. are in fact body Class, how do populate the other properties. Since bodiesOut only has the simulink outs, body names and none of the other properties like cg etc...
I will try to figure this out meanwhile...

nathanmtom:
@salhus Thank you for the update. My major comment is that even the second on third plots are not accurate based on why we would were subtracting off the initial (t=0s) index to show the displacement from equilibrium. If in plots 2 and 3 you subtracted off the body.cg(3) = -0.72 m, then at (t=0s) your position would be +0.5 m as that is the displacement from the equilibrium position. If the body had no initial displacement that would be equivalent to subtracting off the body position at t=0s. Therefore, if you take plot 3 and subtract off the static value of body.cg(3) = -0.72 m, I'm not sure why in the first plot that the plot wouldn't start at +0.5 m. Perhaps I've missed what bodiesOutput was passing into the function and we should be able to pass in the `body` object to pull out the center of gravity. 

I've made some edits from your branch, but do not have write access, so if you can make me a collaborator I can push the changes for your review.

salhus:
@nathanmtom 
That makes sense. I was interpreting that the plots were showing the position of the cg wrt global origin. But, if they are displacement  from the eqbm, I agree plots 2 and 3 would not be accurate

nathanmtom:
@salhus Fully agree that if we want to always output the global center of gravity with respect to the origin then no additional edits are needed to resolve the bias issue; however, since `plotResponse` was coded to output the location of the center of gravity relative to the equilibrium position then we should try to make edits to `plotResponse` consistent with the previous implementation. We can discuss more at the next WEC-Sim development meeting to determine if we'd like to move to one output coordinate system or another.

I just pushed my minor edits and you can review to check if there are any issues from your perspective.

salhus:
@nathanmtom 
Thanks a lot ! I will go through them now. Appreciate the help

jtgrasb:
@salhus @nathanmtom Just ran it with the updates and it looks like subtracting the cg for the position and plotting the angle as is works well.

nathanmtom (approved):

kmruehl:
I'll go ahead and merge this PR

PR #705, opened 2021-09-15
Add start and end time to output.plotWaves()

akeeste:
This PR adds an optional parameter for the start and end time of the visualization created by the output.plotWaves() method


kmruehl (approved):
@akeeste thanks!

Issue #724, opened 2021-10-04
Wave Visualization not creating GIF [BUG]


jtgrasb:
**Note: italicized text below is include as an example and should be updated before submission. If you feel any section is not applicable to your request, please replace with 'N/A' rather than delete a section.**

**Describe the bug**
When the wave visualization is attempted to be output to a gif file, the error "Can only append to GIF89a format GIFs." occurs. When using the imwrite() function, the first image is set using slightly different notation than the rest. This first step is not completed when the gif animation is not started at time = 0.

**WEC-Sim file(s)**
responseClass.m

**To Reproduce**
_Steps to reproduce the behavior:_
1. _Go to examples/RM3
2. _Open userDefinedFunctions.m and uncomment the wave visualization section
3. _Input the option for 'saveSetting' as 1.
4. _Run wecSim and see error.

**Expected behavior**
_Function should write images to gif.

**Screenshots**
![image](https://user-images.githubusercontent.com/87095491/135887227-e3031741-1a88-4f24-9f84-5b03930aa9e6.png)

**Workstation Information (this section MUST BE completed):**
 - OS: Windows 10
 - MATLAB 2021a
 - v4.3 (Dev)

**Additional context**
This should be a pretty simple bug to fix, so I will work on it and plan to submit a PR today.

PR #725, opened 2021-10-04
Add image counter to fix gif initialization

jtgrasb:
Fix for issue #724. Add a counter which is used to check whether the image is the first in order to initialize the gif file. This was originally done by check if the iteration was the first element of the time vector, but this gives an error when the wave visualization is not started at t=0.


kmruehl:
@jtgrasb thank you! I was able to reproduce this error with the gif export and resolve it with this PR, Thanks!

PR #727, opened 2021-10-05
Bugfix for Read Capytaine

akeeste:
This fixes a bug in the BEMIO Read_Capytaine.m function. Part of that function parses dofs to ensure they are in the correct order. Capytaine always uses lower case body names when naming dofs. This PR updates Read_Capytaine to do the same and ensures the dof order is correct.


kmruehl:
@akeeste thanks, merging now

PR #729, opened 2021-10-06
Add nlHydro option to body masks

akeeste:
This PR adds the nlHydro flag to body masks, so that it can be used in the Run From Simulink feature.

Only rigid and flexible body masks changed. A single mask parameter (``nlHydro``) was added to each. Visibility callbacks updated to hide the nlHydro flag when a body is drag or nonhydro.


kmruehl:
@akeeste after a lot of (frustrating) troubleshooting I determined the issue. While the `source/lib/WEC-Sim/WECSim_Lib_Body_Elements.slx` sub-library block was in the source directory, and had been updated correctly to include nonlinear hydro, for some reason none of the sub-library blocks were being called on from `WECSim_Lib.slx`. This resulted in the strange error saying there was no nonlinear hydro from the mask, when `WECSim_Lib_Body_Elements.slx` clearly included it...  because the `WECSim_Lib_Body_Elements.slx` sub-library was never being used. 

Turns out it was an easy fix was I was able to identify the issue. This must have been the result of a failed merged, possibly one that was done automatically. Anyway, we should be good now. Thank you!

PR #730, opened 2021-10-06
Remove passive yaw tests from WEC-Sim repo

akeeste:
This PR moves the passive yaw test from the WEC-Sim repo to the applications repo, where it already exists.
It also makes a minor fix to the BEMIO test so that plot visibility is restored properly after the test runs.


kmruehl:
Passive yaw test migrated to applications repo with https://github.com/WEC-Sim/WEC-Sim_Applications/pull/15

kmruehl:
This PR does the following:
- removes passive yaw tests
- cleans up bemio test
- moves `spectralMoment` to source/functions

kmruehl:
@akeeste this pr is no longer current with dev so I'm going to close and re-open a new one

PR #743, opened 2021-10-13
DRAFT: Updated blocks: Gas Hydraulic Accumulator, compressible fluid piston,…

jleonqu:
This is a pull request with some PTO-Sim blocks updated.


jtgrasb:
The PTO-sim updates look pretty good so far! I am not very familiar with the details of PTO systems so I have a few questions:
First, is the gas-charged hydraulic accumulator made to replace the normal accumulator or is it a different type of accumulator? 
What does the electric machine correspond to in terms of a real-life PTO system?
Will all of the PTO-Sim blocks be updated to match the formatting of these blocks, such as the rest of the hydraulic components?
Thanks! -Jeff

dav-og (commented):
nice work Jorge! - the new additions look good - although I haven't tested them in an example case. I did check the PR with another model that features a heavily modified incompressible fluid piston block, accumulators, motor, valves, etc and everything is working fine 👍

if you have a working example using the new blocks that could be a nice test to include - lets discuss with the others, re: https://github.com/WEC-Sim/WEC-Sim/projects/58#card-64155595

jleonqu:
Thanks David and Jeff for your comments. I have included more hydraulic blocks and there is an application example in examples/RM3NewPTOSim.

Jeff, the electric machine block can be used to simulated an electric generator. It could be used to simulate an electric motor with some minor modifications. There some options for hydraulic accumulators such as a spring load accumulator or a gas charged accumulator, in this case I have decided to start with the gas charged accumulator.

dav-og:
Looks good Jorge - seems to be working fine 👍
Lets check with @kmruehl and the others about whether it makes sense to keep your example in WEC-Sim/examples or to put it in the WEC-Sim Applications repo

jtgrasb:
I am getting a few errors when trying to run the example case. First, the PTOSimBlock variable is spelled incorrectly in the wecSimInputFile. Once I fixed this, I was still getting the following errors, which I couldn't figure out how to fix.

errors

I also noticed that some of the variables are not in camel case, such as some variables in PTOSimClassUpdated.m (Ex: BulkModulus should be changed to bulkModulus to be consistent with WEC-Sim convention).

docs/user/code_structure.rst
```diff
@@ -736,6 +736,65 @@ restrictions on the number of MooringMatrix blocks.
...
+
+The PTO-Sim blocks receive the linear or angular response from the PTO blocks and give either the torque or the force depending on the PTO dynamics.
+
+PTO-Sim Class Initialization
+""""""""""""""""""""""""""""""""
+The properties of the PTO-Sim class (``PTOSimClass``) are defined in the ``PTOSimBlock`` object. The PTO-Sim class must be
+initialized in the ``wecSimInputFile.m`` script. There are three properties that must be initialized for all the PTO-Sim blocks,
+those are the name, the block number, and the type:: 
+
+    PTOSimBlock(i) = cableClass('PTOsimName');
```
nathanmtom, line 762: @jleonqu I believe `cableClass` should be `ptoSimClass`, but can you confirm and update as required.

examples/RM3NewPTOSim/RM3Files/README.md
```diff
@@ -0,0 +1,11 @@
+# PTO-Sim Compressible Hydraulic PTO
+
+**Author:**  	Ratanak So
```
nathanmtom, line 3: @jleonqu Minor comment, but I would expect you should be listed as an additional Author of this example or repository.

examples/RM3NewPTOSim/RM3Files/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+**Version:** 	WEC-Sim v2.0
```
nathanmtom, line 5: @jleonqu You will need to update the WEC-Sim version number that this updated example will be compatible with.

examples/RM3NewPTOSim/RM3Files/wecSimInputFile.m
```diff
@@ -0,0 +1,101 @@
...
+pto(1) = ptoClass('PTO1');           	% Initialize PTO Class for PTO1
+pto(1).k = 0;                           % PTO Stiffness [N/m]
+pto(1).c = 0;                           % PTO Damping [N/(m/s)]
+pto(1).loc = [0 0 0];                   % PTO Location [m]       
+
+%% PTO new blocks
+
+%Hydraulic Cylinder
+%PTOSimBlcock(1) = PTOSimClassUpdated('HydraulicPiston');
+PTOSimBlock(1) = PTOSimClassUpdated('PTOSim1');
```
nathanmtom, line 46: @jleonqu can you double check if other changes need to be made to the code now that `PTOSimClassUpdated` was removed from this PR.

examples/RM3NewPTOSim/RM3Files/powerPlotCode.m
```diff
@@ -0,0 +1,26 @@
+TimeV = PTOSimBlock5_out.time;
+shaftSpeed = PTOSimBlock5_out.signals.values(:,1);
+TorqueHM = PTOSimBlock5_out.signals.values(:,2);
+Current = PTOSimBlock6_out.i.Data;
+Voltage = PTOSimBlock6_out.Voltage.Data;
+
```
nathanmtom, line 6: @jleonqu Just for clarification, but the numeric values after PTOSimBlock_#_ will not be automatically updated and in this example these PTOSimBlocks are the corresponding blocks for the power conversion process. If so you may want to clarify as comments that if changes are made to this `powerPlotCode.m` the user will need to update these manually.

nathanmtom:
@jleonqu In the `RM3NewPTOSim' example folder, the `wecSimInputFile.m` is looking for a function called PTOSimClassUpdated.m which seems to have been deleted during one of your last commits. I have tried to update to use 'ptoSimClass', but the function was incorrectly capitalized in the function. After completing this, there still seems to be an issues such that if you try `PTOSimBlock(1) = ptoSimClass('PTO1')` WEC-Sim will error and only proceed once you remove `PTO1` to be `PTOSimBlock(1) = ptoSimClass('PTO1')`.

Some other minor notes:

- There appears to be a `untitled.slx` which I would expect you should remove to clean up this PR.
- There appears to be a PTOSim_LibOld.slx` which I would also expect you should remove to clean up this PR.
- There also appear to be conflicting versions of some WEC-Sim files, not sure if you need to pull the latest versions from `Master` and `Dev` into your branch. 
- Your MATLAB `.slx` files are saved in version 2021b. This is my issue, but right now I'm only at 2021a so if you can save the `slx` files as a 2020b version then we can ensure all users can open the `.slx` file.

@jleonqu If you are able to review and make some of these requested changes, I should be able to continue my review.

akeeste:
I have received a response from Mathworks support on this issue:

    "In short, we believe this is a bug and the root cause is due to the multitasking executed in normal making the "if-else" and "merge" block cannot correctly execute during the minor steps. 
    
    Please refer to the attached "pic1" for the problematic "if-else" block. 
    
    As for a workaround, there are several ways:
    1. Keep the "if-else" block and disable the multitasking (RM3 Simulink -> Model Configuration -> Solver -> Solver details -> Uncheck treat each discrete rate as a separate task)
    
    2. Replace the "If-else" & "Merge" with "Switch" as shown in "pic2". 
    
    3. Keep the "if-else" block, simulate the model in Accelerator mode. 
    
    I believe this solves your issue and hope the workaround works for you. I have submitted a bug report on behalf of you and our development team will consider fixing it in the future release.
    "
pic1
pic2


In response the proposed solutions, # 1 seems like it might effect wecSimPCT though I not checked, # 2 is possible, # 3 is undesirable as we don't want to confine users to Accelerator mode.

As a 4th option, we can move the ramp to a *.m function. I confirmed that this works around the solver bug. The ramp is a simple enough function, but it is defined in many different places in WEC-Sim. It would be better to use a single function for all of these ramp calls, which could later be implemented with a *.m function.

Adam

jleonqu:
Closing this PR. The updates are now in PR #821

Issue #748, opened 2021-10-15
[BUG] Error when generating the vtk directory for Paraview 


vibishana:
**Note: italicized text below is include as an example and should be updated before submission. If you feel any section is not applicable to your request, please replace with 'N/A' rather than delete a section.**

**Describe the bug**
Hello! I am modeling a point absorber in 3DOF. I have placed a translational PTO between two rotational joints to represent the tether. However when I tried to generate vtk directory for paraview visualization following error occurred. 

Moreover, I have scaled down the real dimensions by 1/1000 times in the .stl file so that it fits the Simulink explorer. Does this affect my calculations?

Thank you in advance!

**WEC-Sim file(s)**
N/A

**To Reproduce**
Run wecSim

**Expected behavior**
For WEC-Sim to generate vtk directory.

**Screenshots**
![err](https://user-images.githubusercontent.com/92297649/137538067-56fb487a-745a-4fad-ad7d-59dabbcc5458.png)

**Workstation Information (this section MUST BE completed):**
 - OS: [Windows 10]
 - [MATLAB/Simulink Release][2020a]
 - [WEC-Sim Release][v4.3]

**Additional context**
I have attached my files below
[BuoyFour.zip](https://github.com/WEC-Sim/WEC-Sim/files/7355440/BuoyFour.zip)

PR #760, opened 2021-11-01
Allow binary STL files

akeeste:
This PR updates WEC-Sim to allow users to use binary stl files. It removes the custom ``import_stl_fast`` function and instead uses the MATLAB supported ``stlread`` function which supports binary and ASCII STL files. ``stlread`` runs as or more quickly than ``import_stl_fast``.

Resolves issue #748.


kmruehl (commented):
@akeeste this all looks good to me, but we need to revisit the discussion with the team about whether this is the direction we want to go. Also, we need to revise the [NOTICE](https://github.com/WEC-Sim/WEC-Sim/blob/master/NOTICE) file to remove the `import_stl_fast` dependency.

Issue #683, opened 2021-08-20
ERROR during import data from ANSYS AQWA


Dadidal:
**Cite the [WEC-Sim applications](https://github.com/WEC-Sim/WEC-Sim_Applications) case in question**
BEMIO import from Ansys Aqwa

**Describe the bug or issue**
I calculated the Hydrodynamic Response of structure using ANSYS Aqwa (Hydrodynamic Diffraction module in ANSYS Workbench). During the import of hydrodata from Ansys Aqwa and creating the .h5 file I got the following ERROR:

Unable to perform assignment because the size of the left side is 6-by-1 and the size of the right
side is 1-by-5.

Error in Read_AQWA (line 232)
                hydro(F).sc_ph((kdiff-1)*6+1:kdiff*6,j,i) = tmp2;  % phase of scattering force

**To Reproduce**
1. Create bemio.m file for import from Ansys Aqwa
2. Run bemio.m
3. See error

**Expected behavior**
Generated .h5 file

**Screenshots**
![Wec-Sim error](https://user-images.githubusercontent.com/32517441/130300243-b8e6d7a6-b98d-40da-b4fa-1f63ec7d2d04.png)

**Workstation Information (this section MUST BE completed):**
 - OS: Windows 10
 - MATLAB/Simulink 2021a
 - WEC-Sim Release v4.3
 - **ANSYS 2021R1**

**Additional context**
How I can solve this problem and import the hydrodata from .AH1 and .LIS files generated using Aqwa?

Dadidal:
[hydroData.zip](https://github.com/WEC-Sim/WEC-Sim/files/7027285/hydroData.zip)

dforbush2:
Hi @Dadidal 

We're not AQWA experts, so the support we can provide may not be comprehensive. 

AQWA (very unfortunately) changes the content and structure of their output files with every new edition. So we often encounter versioning issues, and this may be the case here since you are using a new edition, but I do not see anything that suggests this is the case. 

What I am seeing when I try to run your files is that, for some reason, the variable `raw2`  which should read in the .LIS file is only read in part of the way, and thus is unable to return meaningful results, resulting in the error you are seeing. This suggests that maybe the EOL characters in the textscan function are not, for some reason, working. Will follow up.

nathanmtom:
@Dadidal The WEC-Sim team wanted to check-in to see if @dforbush2 response above has answered your initial questions? 

Cheers, 
Nathan 
[...]

PR #761, opened 2021-11-03
Update Read_AQWA

jtgrasb:
I altered the Read_AQWA function so that it can account for headers in the LIS file (resolves issue #683). AQWA places the headers in different places for the Froude-Krylov and Diffraction Forces depending on the amount of frequencies being run. The updated Read_AQWA function keeps a count of when a header is placed in order to skip the header and only read in the relevant data.


jtgrasb:
I also updated the script to fix a bug that derives from my last PR for Read_AQWA. The added_mass was being overwritten with zeros the way I initially wrote the code so this commit fixes the added_mass readings.

Devsh93:
Can you please upload updated github zip

Devsh93:
Error using wecSim (line 46)
['RM3/Global Reference Frame/Solver Configuration']: Error evaluating equations at time
0.0045000000000000005. Stopping simulation.  There may be a singularity in the solution.  If not, try
reducing the step size (either by reducing the fixed step size or by tightening the error tolerances)
Caused by:
    Error using wecSim (line 46)
    ['RM3/Floating (3DOF)/CONSTRAINT (Planar Joint)']: 'RM3/Floating (3DOF)/CONSTRAINT (Planar Joint)' has a degenerate mass distribution on its follower side.

jtgrasb:
@Devsh93 It seems like you had some issues with the RM3 AQWA example? I updated the AH1 and LIS files as I realized they weren't from my latest run -- thanks for catching it! For some reason, it won't let me upload a .zip file here, but let me know if you still need it and I can find a different way to upload.

kmruehl:
@jtgrasb this PR resolves issues with the READ_AQWA script, so I'd like to merge that. However, the RM3 AQWA BEM solution for looks like it has an irregular frequency issue that's causing IRF instabilities. Have you compared this AQWA solution to the other RM3 runs? Did you run this case with an internal lid? 


image
image

kmruehl:
@jtgrasb I'm going to merge this, but will reference this AQWA RM3 comment in another issue/pr

kmruehl:
@jtgrasb would you like to delete this branch?

jtgrasb:
@kmruehl Thanks! I deleted it and I will address the RM3 issue in the other PR.

PR #765, opened 2021-11-10
Rename plotWaves

jtgrasb:
I renamed the plotWaves function to plotWaveBodyVisualization to describe the function more accurately. I also updated the documentation to account for the new name. @kmruehl Did you also want this function moved to a different class?


kmruehl:
@jtgrasb do you have any ideas on what we could call this? It does more than `plotWaves`, but `plotWaveBodyVisualization` is very long. What about something like `plotViz`, `saveViz` or `saveAnimation`? I somewhat prefer using "save" since this method automatically saves a `gif` or `avi` file, unlike the plot methods. Thoughts?

For reference, are the places it shows up in the docs:
http://wec-sim.github.io/WEC-Sim/master/user/advanced_features.html#wec-sim-visualization
http://wec-sim.github.io/WEC-Sim/master/user/api.html#objects.responseClass.plotWaves

jtgrasb:
@kmruehl thanks for the suggestions! I think saveViz is nice and short but gets the point across, so I'll change it to that and update the relevant documentation.

kmruehl (approved):
This looks great, thanks! I'll merge it now

Issue #772, opened 2021-11-20
[BUG] unresolved Link Goto1  in Cable/Base or Follower Drag Body  , MATLAB v2021a


maoyanjun:
**Describe the bug**
 when I explored the calbe block in the Model Explorer， It showed Error information for Cabel/Base Drag Body/ Goto1 block: 
```
Failed to find 'WECSim_Lib/Body Elements/Rigid Body/Hydrodynamic Body/Goto1' in library 'WECSim_Lib'.
```
Hence It got multiple errors when running a case of two bodies connecting with calbe like the snipshot below.
**WEC-Sim file(s)**
N/A

**To Reproduce**
N/A

_State method of running WEC-Sim : from command line

**Expected behavior**
I tested several examples, They all worked fine. For now, I just found  the cable block has such problems on my PC. I guess it maybe cause by incompatiblity of MATLAB version.

Besides I want to know how to delete the Base Drag Body and Follower Drag Body in the Cable mask and creat another cable2 mask based on current Cable mask. They are not necessary in my model. 

**Screenshots**
here is the simulink model:
![image](https://user-images.githubusercontent.com/34297878/142712483-23450a69-502a-44b6-937c-60cf2dcee6f3.png)

here is the error messages:
![image](https://user-images.githubusercontent.com/34297878/142712199-992ed965-7dd9-4430-899a-681b3895ca16.png)

**Workstation Information (this section MUST BE completed):**
 - OS: [ Windows 11]
 - [MATLAB/Simulink Release](https://www.mathworks.com/help/matlab/release-notes.html) [ 2021a]
 - [WEC-Sim Release](http://wec-sim.github.io/WEC-Sim/man/release_notes.html) [ v4.4]

**Additional context**
N/A

akeeste:
@maoyanjun 

Thanks for your question and using the template. This bug is tied to issue #770. There are a few library blocks that were recently renamed and the links in the cable block were broken. This should be a quick fix that I will create a PR for shortly.

On the second half of your question, can you explain what problem the cable drag bodies cause? You can effectively remove these drag bodies by leaving the following variables as the default values:

    cable.bodyMass = 1;
    cable.bodyInertia = [1 1 1];
    cable.viscDrag.characteristicArea = [0 0 0 0 0 0];
    cable.viscDrag.Drag = zeros(6);
    cable.viscDrag.cd = [0 0 0 0 0 0];

[...]

PR #774, opened 2021-11-24
Fix cable library links. Resolves #770

### LINKED ISSUE #770: [BUG] Moving the library source files has changed install procedure

**Describe the bug**
Splitting the library source files has changed the startup procedure. Following the procedure in the documentation results in several broken links

**WEC-Sim file(s)**
All library source files combined with startup.m. It may in particular be cableClass or cable blocks.

**To Reproduce**
Fresh clone the library and follow the install instructions as presented in documentation. Run a model with a cable block.

_State method of running WEC-Sim (command line, from simulink with input file, from simulink with custom parameters)_
Command line from wecSimInputFile

**Expected behavior**
The body link in the cable class should initialize correctly from the startup procedure.

akeeste:
This PR fixes the broken cable library links identified in issues #770 #772


dforbush2 (approved):
Tested on MBARI model. All works as expected, links fixed. Works successfully following standard git clone install procedure.

PR #779, opened 2021-12-06
AQWA Examples

jtgrasb:
This PR adds Ansys AQWA example cases to be consistent with the example cases for WAMIT/NEMOH/Capytaine.

PR #790, opened 2022-01-26
BEMIO Cleanup

akeeste:
This scripts completes the BEMIO cleanup project board task:
- [x] rename BEMIO variables and internal functions to camel case
- [x] rename BEMIO scripts to camel case
- [x] removed redundant writeh5 functions

updates after review:
- [x] changed ``writeH5`` to ``writeBEMIOH5`` to prevent confusion with matlab functions
- [x] updated ``H5CreateWriteAtt`` to ``writeH5Parameter`` since it creates a variable, writes values, and defines attributes
- [x] removed ``bemio.m`` and replaced it with a short ``README.md`` pointing to the examples/BEMIO cases
- [x] removed specific BEM names in ``normalizeBEM.m``, it is general now
- [x] removed the ``loadH5`` function since it replicates a lot of what the built in ``h5read`` does. The only purpose it served for us was to reverse the order of the dimensions upon reading from the h5 file. I'm not sure why this is necessary, but I created a short utility function that will does this same thing after using ``h5read``

updates building on Kelley's compareBEMIO PR into this fork:
- [x] test plotting for different number of WEC bodies, single/double/triple, e.g. sphere, rm3, wec3
- [x] test plotting for single, double and triple BEMIO results, e.g. WAMIT, Capytaine, AQWA
- [x] test plotting when nested in `plotBEMIO`, resolve issues with nested varargin functions
- [x] remove *.mat and AQWA files
- [x] add examples/**/*.png to gitignore so that images of BEMIO plots are not included on the repo 
- [x] Point bodyClass to the readBEMIOH5 function
- [x] refactor plotting scripts to remove the hydro parameter entirely. Also functions use ``varargin`` for simplicity
- [x] create new test for the compareBEMIO functionality by running `plotBEMIO` for two different Sphere BEM runs

Future Task --> project board
- [ ] Compare BEM data by loading h5 data into hydro data structure. Right now, this comparison is based on a temporary hydro structure from the BEM output files

PR #797, opened 2022-02-07
Fix AQWA data for sphere, ellipsoid, and WEC3

jtgrasb:
This PR addresses comments from https://github.com/WEC-Sim/WEC-Sim/pull/779


kmruehl:
This PR updates the AQWA runs for the following cases:
- [ ] Ellipsoid
- [ ] Sphere
- [ ] WEC3

kmruehl (changes requested):
Refer to comments about Sphere and WEC3

jtgrasb:
Thanks for the input @kmruehl. I was looking mostly at the shapes rather than values so didn't really notice the differences with the sphere (I guess that's why we need the compareBEM function haha) -- thanks for pointing those out. I will try a few things and should be able to figure it out. As for the WEC3, I'm not quite sure what the issue may be, but I'll try adjusting the mesh a bit and see how that changes it.

kmruehl:
Thanks! And hopefully the updates to plotBEMIO will make this comparison much easier :)

kmruehl:
@jtgrasb I found another run that needs to be updated, the RM3, especially the radiation damping. There is a frequency spike around 2 rad/s that's making the IRF unstable, especially in surge and pitch. This plot is comparing the WAMIT to the AQWA solution for the RM3.

image
image

kmruehl:
@jtgrasb https://github.com/WEC-Sim/WEC-Sim/pull/790 has been merged into dev, so please pull this into this PR

jtgrasb:
@kmruehl Okay, I've pulled the CompareBEMIO updates into this PR and I will check out the RM3 as well. Thanks!

kmruehl:
Great, thank you!

jtgrasb:
I have fixed the example cases for the sphere and ellipsoid and also improved the example case for the RM3. Unfortunately, the RM3 and WEC3 still have unstable IRF's in surge and pitch that I am working on fixing.

akeeste:
@jtgrasb Have you tried changing the frequency limit for the IRF? If you can't get rid of the IRR problems in Aqwa itself, you can adjust this limit to try and tune the IRF. Not sure what you're using now, but if you set the frequency limit in the IRF to 2rad/s then most IRR spikes are avoided and the IRF may decay better. This worked somewhat in OMT, might be useful here.

kmruehl:
@jtgrasb is this ready for a review?

jtgrasb:
Although the WAMIT results seem to avoid the spikes for the RM3, the Capytaine and Nemoh results have similar spikes and just use a frequency limit for the IRF. I will keep looking into potential solutions for the RM3 and WEC3, but I think it is okay to merge for v5.0 for now.

kmruehl (approved):
Thanks @jtgrasb I'm merging this now

Issue #793, opened 2022-01-31
[BUG] Utilizing a fixed-step time solver when simu.dt and simu.dtOut are not equal


nathanmtom:
**Describe the bug**
There have been several recorded instances where the error show in the screenshots section occurs when using a fixed step time solver `ode4` while setting `simu.dt` and `simu.dtOut` to different values. This has so far not been an issue when using the variable time step solver `ode45`.

Anticipate that changes to the rate transition block made to ensure consistent time stamps for the variable time step has caused the issue. Past models with a fixed time step and different `simu.dt` and `simu.dtOut` were working as expected.

**WEC-Sim file(s)**
Currently, the issue is likely to be caused by the rate transition block in the `to workspace` block that outputs recorded quantities of interest at a different sampling rate than the simulation time step.

**To Reproduce**
_Steps to reproduce the behavior:_

1. Open example RM3 model.
2. Add simu.dtOut = 0.2.
3. Type WEC-Sim in the command line.
4. See the following error shown below in screenshots.

**Expected behavior**
Regardless if a fixed or variable time step solver is used, the simulation time step and the output time step are different the simulation should execute successfully.

**Screenshots**

![image](https://user-images.githubusercontent.com/7635064/151838615-af28b7c8-f617-47f8-b09e-f89f52f5ab3c.png)

**Workstation Information (this section MUST BE completed):**
 - OS: Windows 10
 - MATLAB/Simulink Release:  2021a
 - WEC-Sim Release: v4.4

**Additional context**
N/A

akeeste:
Hi @nathanmtom 

I looked into this issue a bit today. There are two solutions that I found:
- explicitly add a rate transition block immediately before the ``To Workspace`` block
![image](https://user-images.githubusercontent.com/72414466/152017658-7652d4d4-470b-4fc1-bf2d-2cf5d62c6c6f.png)

- change the ``To Workspace`` signal from being a `bus` to a vector (`mux` block)
![image](https://user-images.githubusercontent.com/72414466/152017536-3e7cd851-2ec3-4393-b8dd-d72de92b74da.png)

[...]

PR #799, opened 2022-02-08
Fix rate transition error

akeeste:
This PR resolves the issue in #793 by changing the ``bus`` to ``mux`` immediately before each body's ``ToWorkspace`` block. This is the same method used in the constraint and PTO blocks.

@nathanmtom  Can you see if this fully resolves your issue?

@dforbush2 I couldn't quite figure out how to get the [model level sample time](https://www.mathworks.com/help/simulink/gui/output-options.html) to work correctly. I tested with the RM3 but couldn't consistently change the specified times to output. If you have any insight here I think that could be a cleaner solution.


nathanmtom:
Thanks @akeeste for suggesting these changes and I can confirm that WECSim will run with a fixed-step solver when `simu.dt` is not equal to `simu.dtOut`; however, what is strange is that when running with `simu.dtOut` is double `simu.dt` in our RM3 model we get different results than when both variables are the same. The first image below shows the base example case with the same sampling rate (for both variables) while the lower image is when `simu.dtOut` is double `simu.dt` where we pick up a low frequency oscillation in the output. I double checked the WEC-Sim master branch and when using `ode45` rather than `ode4` this behavior is not reproduced. The `ode45` implementation provides the same output time histories. So this somehow still sees to be an issue only for `ode4` with different sampling rates between simulation and output.

simudtout1
simudtout2

This is strange as you have only changed a `bus` to a `mux` and the `To Workspace` block should not have an effect on the actual system dynamics since this is just logging data.

akeeste:
@nathanmtom very strange! I'll take a look and see what's going on. If needed I will revert back to master and make the change again to ensure nothing else happened while I was working.

kmruehl:
@dforbush2 and @akeeste what's the status on this PR?

dforbush2:
@kmruehl Still not clear why this happens...checking if it is something related to MATLAB version, since this has not been noticed till recently.

dforbush2:
To update: Error replicated in 2020a (oldest library version), 2020b, and 2021b. 

Error does not occur with all fixed-step solvers: using ODE1 shows the expected results, while ode2 and ode4 show the erroneous low-frequency component. This suggests that it is a Simulink solver implementation issue and is not a product of our code.

Changing the output generation at the top level as per https://www.mathworks.com/help/simulink/gui/output-times.html is NOT a robust solution because this parameter is only read for variable-step solvers and, to my investigation, cannot easily be made active for fixed-step solvers.

To pursue a solution, @akeeste has contacted MathWorks. In the short term, it is recommended that any re-sampling of output data takes place in post-processing.

akeeste:
I have received a response from Mathworks support on this issue:

    "In short, we believe this is a bug and the root cause is due to the multitasking executed in normal making the "if-else" and "merge" block cannot correctly execute during the minor steps. 
    
    Please refer to the attached "pic1" for the problematic "if-else" block. 
    
    As for a workaround, there are several ways:
    1. Keep the "if-else" block and disable the multitasking (RM3 Simulink -> Model Configuration -> Solver -> Solver details -> Uncheck treat each discrete rate as a separate task)
    
    2. Replace the "If-else" & "Merge" with "Switch" as shown in "pic2". 
    
    3. Keep the "if-else" block, simulate the model in Accelerator mode. 
    
    I believe this solves your issue and hope the workaround works for you. I have submitted a bug report on behalf of you and our development team will consider fixing it in the future release.
    "
pic1
pic2

I only see two If blocks in our model:
WECSim_Lib_Body_Elements/Rigid Body/Hydrodynamic Body/Wave Diffraction and Excitation Force Calculation/If2
WECSim_Lib_Body_Elements/Flex Body/Wave Diffraction and Excitation Force Calculation/Wave Diffraction and Excitation Force Calculation/If2

In response to the proposed solutions, # 1 seems like it might effect wecSimPCT though I not checked, # 2 is possible, # 3 is undesirable as we don't want to confine users to Accelerator mode.

As a 4th option, we can move the if statement and ramp to a *.m function. I confirmed that this successfully works around the solver bug. The ramp is a simple enough function, but it is defined in many different places in WEC-Sim. It would be better to use a single function for all of these ramp calls, which is easiest to implement with a *.m function.

Adam

akeeste:
I updated this branch to
1. change the bus to mux for body ``To Workspace`` blocks (1 in rigid body, 1 in nonhydro, 1 in drag, 2 in flex)
2. change the bus to mux for mooring ``To Workspace`` blocks (1 in mooring, 1 in MoorDyn)
3. change the bus to mux for PTO-Sim ``To Workspace`` blocks (11 total, 1 in each block)
4. Replace the ``If`` blocks in the rigid and flex body excitation forces to use a switch instead

kmruehl:
@akeeste and @nathanmtom is this ready to be merged into master?

akeeste:
@kmruehl @nathanmtom @dforbush2 This should be reviewed one more time since I implemented new changes to the library.

nathanmtom:
@akeeste @kmruehl I've just reviewed these changes and can confirm that the low frequency oscillation is no longer present when running `ode4` when `simu.dt` and `simu.dtOut` are not equal. Therefore, this bug fix should address the initial issue with WEC-Sim erroring out when have the simulation and output time steps at different rates and the later issue of a low frequency signal in the output time history. 

I do want to acknowledge @akeeste comment about transitioning the ramp function to a `*.m` file. In the initial development of WEC-Sim, if I remember correctly, where trying to do all computations with Simulink blocks as at the time it proved computationally faster. Given our recent conversion from Simulink blocks to `*.m` files this seems like a reasonable script to make as it is applied for all instants where we are calculating the wave elevation or wave-excitation forces. This is not needed for this PR, but do not want to lose discussion on that topic.

However, these does seem to be a conflict with the mooring library that needs to be resolved before merge.

akeeste:
@nathanmtom In a different branch I have been working on combining some duplicate functions. I think I can finish implementing a ramp *.m there, but that will be after v5.0

and thanks for pointing out that conflict. The other PR causing this was just merged. I will pull those changes in next week so that we can merge this.

kmruehl:
@akeeste LMK when you've updated the Mooring library block and this is ready for a merge

akeeste:
> @akeeste LMK when you've updated the Mooring library block and this is ready for a merge

@kmruehl The mooring library is merged and this PR is ready

PR #800, opened 2022-02-08
Update documentation on mean drift and current

akeeste:
This PR addresses project card https://github.com/WEC-Sim/WEC-Sim/projects/58#card-72526916
It adds minor doc updates to clarify how/when the mean drift force and waves.current options are used.

To my understanding mean drift and current each have one scenario they affect:
- the mean drift is combined with the excitation force, if coefficients are included in the BEM data
- the current speed is used to calculate a more accurate fluid velocity and acceleration in the Morison Element calculations

### DIFF
--- a/docs/theory/theory.rst
+++ b/docs/theory/theory.rst
@@ -101,12 +101,13 @@ as:
 
 .. math::
 
-    m\ddot{X}=F_{exc}(t)+F_{rad}(t)+F_{pto}(t)+F_{v}(t)+F_{me}(t)+F_{B}(t)+F_{m}(t)
+    m\ddot{X}=F_{exc}(t)+F_{md}(t)+F_{rad}(t)+F_{pto}(t)+F_{v}(t)+F_{me}(t)+F_{B}(t)+F_{m}(t)
 
 
 where :math:`\ddot{X}` is the (translational and rotational) acceleration 
 vector of the device, :math:`m` is the mass matrix, :math:`F_{exc}(t)` is the 
-wave excitation force and torque (6-element) vector, :math:`F_{rad}(t)` is the 
+wave excitation force and torque (6-element) vector,  :math:`F_{md}(t)` is the 
+mean drift force and torque vector, :math:`F_{rad}(t)` is the 
 force and torque vector resulting from wave radiation, :math:`F_{pto}(t)` is 
 the PTO force and torque vector, :math:`F_{v}(t)` is the damping force and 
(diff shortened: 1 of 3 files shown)


kmruehl:
@nathanmtom offered to review the current documentation. @yuyihsiang can you reviewthe mean drift documentation, or recommend someone else to review it?

docs/theory/theory.rst
```diff
@@ -226,14 +235,17 @@ where :math:`\phi` is the randomized phase angle and :math:`N` is the number of
...
 specification of :math:`\phi`, refer to the :ref:`user-advanced-features-seeded-phase` 
-section. Additionally, an excitation force impulse response function is defined 
+section. 
+
+Additionally, an excitation force impulse response function is defined 
 as 
 
 .. math::
 
     K_{e}(t) = \frac{1}{2\pi} \intop_{0-\infty}^{\infty}
```
nathanmtom, line 245: @akeeste In the lower bound of the integral it should just be -\infty

docs/theory/theory.rst
```diff
@@ -186,6 +187,14 @@ ramp function, :math:`H` is the wave height, :math:`F_{exc}` is the frequency
 dependent complex wave-excitation amplitude vector, and :math:`\theta` is the 
 wave direction. 
 
+The mean drift force can optionally be included if coefficients are defined in the BEM data.
```
nathanmtom, line 190: @akeeste Minor comment, but the text should read "mean drift force, and moment, can" correct?

nathanmtom (commented):
@akeeste Thanks for updating the documentation. Other than my two minor edits, this looks good from my end @yuyihsiang might have more comments on the mean drift description. 

Also, we probably need to add cards to add to our documentation on this realization of the second order drift forces. @yuyihsiang can comment but not sure if we are solving the second order problem or using an approximation based on first order diffraction coefficients.

docs/theory/theory.rst
akeeste: good catch, I will update this

nathanmtom:
@akeeste I got some initial feedback from Toan and his comment about the documentation was to change the mean drift force description as follows:

"The mean drift term can optionally be included if mean drift force coefficients are defined in the BEM data."

I think next week it would be worth having a short call with Toan to double check our implementation and discuss how we can expand our documentation on the mean drift implementation.

akeeste:
- [ ] add note that mean drift is only compatible with WAMIT right now
- [ ] resolve conflicts

PR #802, opened 2022-02-15
Update response class API

akeeste:
This PR primarily updates the response class API to
- [x] Better describe the variables contained there and what each refers to.
- [x] Clarify the orientation of the body forces, e.g. that the output.bodies(#).forceRadiationDamping is actually the negative of the radiation damping force applied to the WEC
- [x] cables
- [x] moordyn
- [x] ptosim

### DIFF
--- a/source/functions/postProcessWecSim.m
+++ b/source/functions/postProcessWecSim.m
@@ -87,10 +87,11 @@ end
 
 % PTO-Sim
 if exist('ptosim','var')
-    ptosimOutput = ptosim.response;
+    ptosimOutput = ptosim.response();
 else
     ptosimOutput = 0;
 end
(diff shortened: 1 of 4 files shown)


akeeste:
@kmruehl This is ready for review. Can you confirm if the responseClass API is showing cables, moorDyn and PTO-Sim twice? I'm not sure if this is a local issue with my docs build or not.

kmruehl (approved):

kmruehl:
@akeeste merged, up to you if you want to delete the branch

akeeste:
@kmruehl It does look like the responseClass API is still duplicating some values (cables, moordyn, ptosim) for some reason. I'll take a look at how to remove these duplicate parameters.

The empty parameters are shown as the last parameters, but the documented ones are listed above (cut off): 

image

kmruehl:
@akeeste you should be able to add those to the excluded parameter in the `.rst` file. However, I believe this happens because we have the description of the `responseClass` at the top, but we also create a structure for each of the other classes within the `responseClass` which is then documented by the API unless we explicitly tell it not to. I do not think it's an issue to have them twice, but I agree it would be cleaner to remove them.

PR #807, opened 2022-02-23
Remove passiveYawTest.m

jtgrasb:
In this pull request, I simply removed the passiveYawTest.m testing script as they were moved to the WEC-Sim Applications repo. These tests were failing when I tried to run tests locally, and it seems like they were supposed to be fully removed with PR #738.


akeeste:
@jtgrasb It makes sense to me to remove this. I see that:
1. this test is not called in WEC-Sim by ``wecSimTest``, so it is not normally used
2. the WEC-Sim_Applications passive yaw case contains the irregular regression test

The only missing component is that the regular wave passive yaw regression test is not included in the applications repo. @kmruehl Did we mean to move this too or do we only want to use the irregular wave regression for passive yaw?

Adam

kmruehl:
@jtgrasb yes, it looks like this file should have been removed before and we need to make sure it gets migrated to the applications repo

PR #808, opened 2022-02-23
Update to normalize to handle sorting mean drift forces

nathanmtom:
This pull request edits the Normalize.m to implement the correct sorting of the mean drift forces (if present). In the current implementation, Normalize.m uses the `exist` function to see if the mean drift field is present within the hydro structure; however, this does not appear to work as desired and have updated to check the `isfield` function to check if the mean drift fields are present in the `hydro` structure.


akeeste (commented):
@nathanmtom Good catch on this.

Only request, can you remove the ``examples/RM3/wecSimInputFile.m`` from your PR? Otherwise it looks good

nathanmtom:
@akeeste That is really strange as in my Github Desktop there is no mention of pushing a change to wecSimInputFile.m. But, I have reverted the wecSimInputFile.m to how it should be in the master branch and is now no longer coming up as a changed file.

akeeste:
@nathanmtom Strange! I looks good now though. The PR into dev looks good too

kmruehl:
Thanks @nathanmtom

PR #809, opened 2022-02-23
Update to normalize for mean drift force in dev

nathanmtom:
This is the same code changes as PR #808 , but in the next WEC-Sim release we are planning to have several variable and function name changes. This PR is an attempt to make the update to a future WEC-Sim easier.


kmruehl:
@nathanmtom this looks good to me

PR #812, opened 2022-02-24
Update BEMIO tests and documentation

akeeste:
This PR:
- [x] Turns compareBEM into BEMIO test scripts
- [x] Reorganizes BEMIO tests in an attempt to reuse data within the test and decrease runtime
- [x] Updates the BEMIO documentation to an API doc
- [x] minor update to responseClass API doc to fix duplication issue described at the end of #802
- [x] ~~Removes Write HDF application from documentation (functions are outdated and were removed in #790)~~
- [x] remove outdated bodyClass.loadH5File() method (should have been removed in #790)
- [x] Update formatPlot to take hydro structs with different number of bodies
- [x] Updates convention for BEM names in documentation (i.e. Aqwa instead of AQWA and Capytaine instead of CAPYTAINE)

PR #813, opened 2022-03-02
Fix readBEMIOH5 bug not creating gbm structure

H0R5E:
This PR fixes a bug in the `readBEMIOH5` function where the creation of the `gbm` structure was being skipped. This is because the variables `dof_start` and `dof_end` were not defined.

In general, the use of `try` blocks without a `catch` is not good practice. As in this case, any error will pass and go unhanded until it breaks something downstream. Consider handling the errors that are expected rather than silently passing on any issues.


akeeste:
@H0R5E I caught this in PR #812 which was just merged. This bug should be fixed now

Issue #674, opened 2021-08-11
[BUG] MoorDyn with Paraview - RM3 Application


akeeste:
**Note: italicized text below is include as an example and should be updated before submission. If you feel any section is not applicable to your request, please replace with 'N/A' rather than delete a section.**

**Describe the bug**
The Paraview application case using the RM3 with MoorDyn is not working. It seems the Lines.out file is not yet written completely when read, which causes issues when reading and manipulating the MoorDyn data in the post-processing stage.

**WEC-Sim file(s)**
write_paraview_mooring.m - line 31
responseClass.m - loadMoorDyn() method
Lines.out

**To Reproduce**
1. Install the latest version of MoorDyn for WEC-Sim at: https://github.com/WEC-Sim/MoorDyn. Run BEMIO.
2. Run the applications case: Paraview_Visualization\RM3_MoorDyn_Viz
3. See error:
```
Error using wecSim (line 45)
Error evaluating 'StopFcn' callback of Frames block (mask) 'RM3MoorDyn/Global Reference Frame'.
Callback string is 'run('wecSimStopFunction')'
Caused by:
    Error using wecSim (line 45)
    X and V must be of the same length.
```

Steps to better see the error during simulation:
1. Install the latest version of MoorDyn for WEC-Sim at: https://github.com/WEC-Sim/MoorDyn. Run BEMIO.
2. Place a breakpoint in responseClass.m loadMoorDyn() method (~line 228).
2. Run the applications case: Paraview_Visualization\RM3_MoorDyn_Viz
3. While the simulation is paused, open moordyn\Lines.out and view data.
    
When I stop the simulation at the breakpoint and reopen Lines.out, it appears fully written. Has anyone else seen this issue? Perhaps we can put in a check that ensures MoorDyn has fully written the output files before reading them back in.

**Expected behavior**
I expect the case to run as is without issue, and to fully print all output data to Lines.out

**Screenshots**
Error message:
![image](https://user-images.githubusercontent.com/72414466/129079607-29dc51e0-e356-4494-954c-ee5c5b409b80.png)

Partially written Lines.out:
![image](https://user-images.githubusercontent.com/72414466/129080835-b2f8b486-fcb1-4f25-a866-e50d13892a89.png)

Fully written Lines.out after simulation:
![image](https://user-images.githubusercontent.com/72414466/129081276-456d6b4d-d578-496f-957b-88d52d50367a.png)

**Workstation Information (this section MUST BE completed):**
[...]

Issue #782, opened 2021-12-17
[BUG] Parallel Computing Toolbox (PCT)


zash07:
**Note: italicized text below is included as an example and should be updated before submission. If you feel any section is not applicable to your request, please replace with 'N/A' rather than delete a section.**

**Describe the bug**
_The parallel computing (PCT), did not work for me even for the RM3 example._

**WEC-Sim file(s)**
_N/A_

**To Reproduce**
_Steps to reproduce the behavior:_
1. _Go to 'Rm3 example'_
2. _Click on 'to run pct '_
3. _Scroll down to 'command window'_
4. _See error_

**Expected behavior**
_To be able to work with PCT._

**Screenshots**
_I provide an error for my case the same error can be seen for RM3 original example._
![image](https://user-images.githubusercontent.com/92889661/146556624-b5849429-3cd8-447d-9bb2-77c70e952241.png)

**Workstation Information (this section MUST BE completed):**
 - OS: [Windows 10]
 - [MATLAB/Simulink Release](https://www.mathworks.com/help/matlab/release-notes.html) [2020b]
 - [WEC-Sim Release](http://wec-sim.github.io/WEC-Sim/man/release_notes.html) [Version: 4.3]

**Additional context**
_I wrote the following scripts for parallel computing to run for different sea states and I could solve my problem following that._
![image](https://user-images.githubusercontent.com/92889661/146557252-b89b376b-d658-41c0-86d7-6f4da27828fe.png)
 
![image](https://user-images.githubusercontent.com/92889661/146557302-bc8044cf-c63d-41f0-92a3-66b24d463500.png)

dav-og:
Hoping @jleonqu can help out with this when he's back

jleonqu:
Hello @zash07 

I would say the problem is that the information needed to run the simulation is not available for all the workers created with the parfor command. There are some specific restrictions when running simulink models within parfor. In the current documentation in the Mathworks webpage there is an explicit recommendation for not using the sim function within parfor, see this [link](https://www.mathworks.com/help/simulink/ug/not-recommended-using-sim-function-within-parfor.html). The other option is to use parsim, which could be the solution you are looking for.

[...]

PR #814, opened 2022-03-02
Bug fix: responseClass reading the MoorDyn Lines.out file too early, resolves #811

### LINKED ISSUE #811: [BUG]Paraview_Visualization\RM3_MoorDyn_Viz

**Describe the bug**

I encountered the same error in #674, then I tried to download the code in #682 , but the problem was still there 

**WEC-Sim file(s)**
WEC-Sim_Applications-master\WEC-Sim_Applications-master\Paraview_Visualization\RM3_MoorDyn_Viz

**To Reproduce**
Install the latest version of MoorDyn for WEC-Sim at: https://github.com/WEC-Sim/MoorDyn. Run BEMIO.
Run the applications case: Paraview_Visualization\RM3_MoorDyn_Viz
See error:

_State method of running WEC-Sim (command line)_

**Expected behavior**
I expect the case to run as is without issue

**Screenshots**

**Workstation Information (this section MUST BE completed):**
 - OS: [Windows 10]
 - [MATLAB/Simulink] [ 2020b]
 - [WEC-Sim Release](http://wec-sim.github.io/WEC-Sim/man/release_notes.html) [v4.4]

**Additional context**
_Add any other context about the problem here._

akeeste:
This PR fixes the issue described in #811 and #674. The problem is that the WEC-Sim post-processing is being called before the MoorDyn ``Lines`` library closing function is being called. This updates the stop and close callbacks of the MoorDyn block, and adds a check to ``postProcess`` to ensure that MoorDyn is closed before the Lines.out file is read.

Adam


akeeste:
@H0R5E I tried a similar loop to give MoorDyn time to finish writing output, but without success. I found that this is really a WEC-Sim order of operations issue, rather than a WEC-Sim/MoorDyn racing issue. No matter how much time is given, the Lines.out file is not complete until WEC-Sim calls the MoorDyn function ``LinesClose``.

This started occurring in the last couple of releases because we moved the WEC-Sim post-processing ``stopWecSim`` to be a stop function callback of the Frames blocks. These commands to close MoorDyn are in the MoorDyn block stop function callback.

Previously our post-processing order of operations was something like:
- Run Simulink
- MoorDyn callback when Simulink stops
    -  ``calllib('Lines','LinesClose');``
- ``stopWecSim``
    - ``output.loadMoorDyn()``

But right now our post-processing order of operations is:
- Run Simulink
- Frames callback when Simulink stops
    - ``stopWecSim``
        - ``output.loadMoorDyn()``
- MoorDyn callback when Simulink stops
    -  ``calllib('Lines','LinesClose');``

So our main post-processing is running before we actually close MoorDyn, which completes the Lines.out files. I had moved ``stopWecSim`` into the ``wecSim`` script to better debug and realized that this actually prevented the issue from occurring. I updated MoorDyn callbacks to be:
    if libisloaded('Lines')
        calllib('Lines','LinesClose');
        unloadlibrary Lines;
    end

This way MoorDyn will always close properly if Simulink crashes or stop unexpectedly, but it will not try to use the DLL if ``stopWecSim`` has already been run.

Adam

H0R5E:
OK, Adam, makes sense. I am curious as to why the order of post-processing is as it is - would it not be sensible to post-process in the last step? Also, is there any way to build a test to catch this if it happens again?

akeeste:
@H0R5E I agree, it would be better to run ``stopWecSim`` as our last step. We originally moved this to the Frames callback to make it easier for users to run From Simulink. However since this is a block-level callback and not a model level callback, we don't have control over when it is run. I believe the block callbacks are run in alphabetical order by block name, so the Global Reference Frame is running before MoorDyn. 

@kmruehl I think this should come into our larger discussion on the utility vs maintenance of Run From Simulink

And on the test -- hopefully one of our MoorDyn application tests can catch this. I'll think on this

kmruehl:
Is this PR ready for a merge? I think we have a similar issue going on with wecSimPCT https://github.com/WEC-Sim/WEC-Sim/issues/782

akeeste:
@H0R5E are there any other changes you recommend here before we merge?

H0R5E:
No @akeeste, go ahead.

kmruehl:
@akeeste I'm going to go ahead and merge this, but we need to put it in dev too. Can you make the same revisions on https://github.com/WEC-Sim/WEC-Sim/pull/822? Thank you!

PR #816, opened 2022-03-04
Fix documentation on dev branch

H0R5E:
I have been working on moving our docs compilation to github actions but on the way I noticed that the docs on the `dev` branch were not compiling properly. This PR fixes the issues and preps the dev branch for moving to compilation using Actions.

The issues were:
* Incorrect use of the `include` directive. Files to be included this way should be placed in the `_include` folder.
* Not escaping `*` in a docstring, which sphinx thinks is a bold formatting token.


kmruehl:
@H0R5E thanks, merging this now

PR #817, opened 2022-03-04
Move documentation compilation to GitHub Actions

### LINKED ISSUE #830: [BUG] API Section in Documentation not working
**Describe the bug**
The API section within WEC-Sim's documentation is not functioning correctly and only displays the class names (both dev and master).

**WEC-Sim file(s)**
The objects are failing to import for some reason.

**To Reproduce**
_Steps to reproduce the behavior:_
1. _Go to the WEC-Sim documentation and click on user manual -> API

**Expected behavior**
API should show the parameters associated with each class and their description

**Screenshots**
image
When built on my current branch:
image

H0R5E:
This PR for changes the documentation compilation service from TravisCI to GitHub Actions. Note that the compilation is now set to fail if warnings are generated, so that we can catch issues created within a PR before its merged.

Fixes #830


H0R5E:
I did a push directly to `dev` to correct the issues on the branch prior to merging this, as any errors will not cause a build fail.

H0R5E:
Going to merge while I know it's working. Then I will copy the workflow into dev.

PR #820, opened 2022-03-09
DRAFT: Warning/Error flags

jtgrasb:
In this PR, I’ve implemented warning and error flags that seek to provide more feedback to the user regarding common errors and issues. The following flags/messages have been implemented into wecSim as part of this PR:
- Argument checks were placed in some functions (such as plotResponse, plotForces, etc) of the response and wave classes to provide context on the expected function inputs. 
- Properties within the simulation and wave class were given more specific requirements (such as requiring a positive timestep value).
- Error message when non-hydro bodies are specified before hydro in the wecSimInputFile
- Warning message when waves.waterDepth is defined and there is an .h5 file present
- Error message when waveSpread does not sum to 1
- Error when the name or filename is not specified in the body, constraint, pto, cable, mooring, and ptoSim classes, which indicates to the user that the objects have likely been defined in the wrong order.
- Error messages when necessary parameters (such as cg and cb) are not defined for bodies based on type (hydro, non-hydro, etc)
- Warning messages when extra parameters specified that don’t affect result (such as cg and cb for hydro body)
- Warning message for STL greater than half the size of the domain and error message for when it is greater than the domain size
- Warning for noWaveCIC case that waves.T is not used
- Errors and warning messages for parameters needed for each wave type
- Warning message when simu.dtOut < simu.dt

I am working on more warning and error flags, but wanted to get a PR started here so that it can start to be reviewed.

PR #822, opened 2022-03-10
Refactoring classes and properties - part 2

kmruehl:
This PR is a continuation of https://github.com/WEC-Sim/WEC-Sim/pull/803. Refactoring WEC-Sim Classes and Properties based on [project board card](https://github.com/WEC-Sim/WEC-Sim/projects/58#card-64210391) and discussions with the WEC-Sim team. 

**Objectives:**
- clear and intuitive variable naming using camelCase
- consistent properties across classes, e.g. `pto.loc` versus `mooring.ref`
- convert properties for a single feature to property structures, e.g. `struct(body.morison)` 
- move properties to the appropriate class, e.g.  `simu.yaw` to `body.yaw`
- updating WEC-Sim library accordingly

**Wave Class**
- [x] `waves.etaImport` to `waves.waveImport` to `waves.elevationImport`
- [x]  `waves.spectrumDataFile` to `waves.waveSpectrumFile` to `waves.spectrumFile`
- [x] `waves.etaDataFile` to `waves.waveElevationFile`to `waves.elevationFile`
- [x] `waves.plotEta` to `waves.plotElevation`
- [x] `waves.waveDir` to `waves.waveDirection` to `waves.direction`
- [x] `waves.waveSpread` to `waves.spread`
- [x] `beta` to `theta` to match online documentation
- [x] `waves.deepWaterWave` to `waves.deepWater`
- [x] convert `waves.current` properties to structure, similar to morison element, 
  - `waves.currentOption` to `waves.current.option`
  - `waves.currentDepth` to `waves.current.depth`
  - `waves.currentDirection` to `waves.current.direction`
  - `waves.currentSpeed` to `waves.current.speed`
- [x] convert properties related to BEM data to `waves.bem` structure
  - `waves.numFreq` to `waves.freqNum` to `waves.bem.count`
  - `waves.freqDisc` to `waves.bem.option`
  - `waves.freqRange` to `wave.bem.range`
- [x] add checks for waveClass structures
- [x] convert `waves.marker` properties to structure
   - [x] `waves.markerLoc` to `waves.marker.loc` to `waves.marker.location`
   - [x] `waves.markerSize` to `waves.marker.size`
   - [x] `waves.markerStyle` to `waves.marker.style` 
   - [x] convert `waves.marker` properties to structure in Frames Library

**Simulation Class**
- [x] `simu.writetxt` to `simu.writeText`
- [x] `waves.waveStatisticsDataLoad` to `simu.mcrExcelFile`  (formerly Wave Class)
[...]

PR #826, opened 2022-03-14
Warning/Error flags

jtgrasb:
This is an updated version of PR #820 to be compatible with the new updates to dev.

In this PR, I’ve implemented warning and error flags that seek to provide more feedback to the user regarding common errors and issues. The following flags/messages have been implemented into wecSim as part of this PR:
- Argument checks were placed in some functions (such as plotResponse, plotForces, etc) of the response and wave classes to provide context on the expected function inputs. 
- Properties within the simulation and wave class were given more specific requirements (such as requiring a positive timestep value).
- Error message when non-hydro bodies are specified before hydro in the wecSimInputFile
- Warning message when waves.waterDepth is defined and there is an .h5 file present
- Error message when waveSpread does not sum to 1
- Error when the name or filename is not specified in the body, constraint, pto, cable, mooring, and ptoSim classes, which indicates to the user that the objects have likely been defined in the wrong order.
- Error messages when necessary parameters (such as cg and cb) are not defined for bodies based on type (hydro, non-hydro, etc)
- Warning messages when extra parameters specified that don’t affect result (such as cg and cb for hydro body)
- Warning message for STL greater than half the size of the domain and error message for when it is greater than the domain size
- Warning for noWaveCIC case that waves.T is not used
- Errors and warning messages for parameters needed for each wave type
- Warning message when simu.dtOut < simu.dt
- Moves check for geometry file to only when explorer is on because the geometry file is not necessary when simulation explorer is off

I am working on more warning and error flags, but wanted to get a PR started here so that it can start to be reviewed.


source/functions/initializeWecSim.m
```diff
@@ -96,8 +96,8 @@
 end
 
 % Waves and Simu: check inputs
-waves.checkinputs;
-simu.checkinputs;
+waves.checkInputs;
+simu.checkInputs;
```
kmruehl, line 101: please use `waves.checkInputs()` (etc) to make it clear we're calling a method

source/objects/bodyClass.m
```diff
@@ -168,6 +171,47 @@ function checkinputs(obj)
...
+            % Warning for cg and cb being overwritten
+            if ~isempty(obj.cg) || ~isempty(obj.cb)
+                warning('Center of gravity and center of buoyancy are overwritten by h5 data for hydro bodies.')
+            end
+        end
+        
+        function checkDragNonHydroInputs(obj,ii)
+            % This method checks WEC-Sim user inputs for each non-hydro
+            % body and generates error messages if parameters are not properly defined for the bodyClass.
+            if isempty(obj.cg)
```
kmruehl, line 205: Is there a reason to split these up? Could we just keep them in `body.checkInputs()`, and pass the `nonHydroBodLogic` logic, etc? Asking for consistency with the other classes that only have `checkInputs()`

source/objects/responseClass.m
```diff
@@ -250,6 +250,11 @@
             %         the number of MoorDyn lines
             %
             
+            arguments
+                obj
+                numLines (1,1) double {mustBeInteger}
+            end
```
kmruehl, line 256: Referring to #822, should this be changed to `linesNum` for consistency with e.g. `bodyNum`

source/objects/simulationClass.m
```diff
@@ -28,40 +28,40 @@
...
+        saveWorkspace {mustBeMember(saveWorkspace,0:1)} = 1             % (`integer`) FLag to save .mat file for each run, Options: 0 (off), 1 (on). Default = ``1``
+        simMechanicsFile                              = 'NOT DEFINED'   % (`string`) Simulink/SimMechanics model file. Default = ``'NOT DEFINED'``
+        solver                                        = 'ode4'          % (`string`) PDE solver used by the Simulink/SimMechanics simulation. Any continuous solver in Simulink possible. Recommended to use 'ode4, 'ode45' for WEC-Sim. Default = ``'ode4'``
+        stateSpace {mustBeMember(stateSpace,0:1)}     = 0               % (`integer`) Flag for convolution integral or state-space calculation, Options: 0 (convolution integral), 1 (state-space). Default = ``0``
+        startTime {mustBeNonnegative}                 = 0               % (`float`) Simulation start time. Default = ``0`` s        
+        zeroCross                                     = 'DisableAll'    % (`string`) Disable zero cross control. Default = ``'DisableAll'``
```
kmruehl, line 64: How do these property assets work? Can you share a MATLAB link? How does it affect our documentation, etc? Are you planning to add this to the other classes too?

kmruehl:
@jtgrasb can you resolve the merge conflicts?

kmruehl:
@jtgrasb sorry that dev is a moving target due to the other PRs. This should be the last major change to dev to resolve conflicts with and then we can merge this PR too.

source/objects/simulationClass.m
kmruehl: @jtgrasb I like this feature a lot, and think we should implement it, but I'm wondering whether it's something we should push to after v5.0 and implement across all the classes, or whether we should work to include it in v5.0, LMK what you think.

jtgrasb: @kmruehl I agree, it is a very useful feature. I think it may be best to wait until after v5.0 because although helpful, I don't think it affects WEC-Sim significantly with most use cases. It would also give me more time to work out the kinks and implement it for the other classes. If you think it should be part of v5.0 though, I can try to get the rest of the classes implemented.

kmruehl: I agree, let's hold off of this on v5.0, but please add it to the project board for future development.

kmruehl:
@jtgrasb I'm going to go ahead and merge this, thanks for resolving merge conflicts!

source/objects/simulationClass.m
kmruehl: Actually, can you revert the simulationClass property validation, and then I'll merge this.

source/objects/waveClass.m
```diff
@@ -39,20 +39,20 @@
...
         type            = 'NOT DEFINED';            % (`string`) Specifies the wave type, options include:``noWave``, ``noWaveCIC``, ``regular``, ``regularCIC``, ``irregular``, ``spectrumImport``, or ``elevationImport``. Default = ``'NOT DEFINED'``
         viz             = struct(...                % 
             'numPointsX',   50, ...                 %
             'numPointsY',   50 );                   % (`structure`) Defines visualization options, structure contains the fields ``numPointsX`` for the number of visualization points in x direction, and ``numPointsY`` for the number of visualization points in y direction. 
-        waterDepth      = [];                       % (`float`) Water depth [m]. Default to BEM water depth if not set. 
+        waterDepth {mustBePositive}  = [];          % (`float`) Water depth [m]. Default to BEM water depth if not set. 
         spread          = 1;                        % (`float`) Wave Spread probability associated with wave direction(s). Should be defined as a column vector for more than one wave direction. Default = ``1``
```
kmruehl, line 56: Same here, please revert this and add it to the project board

Issue #824, opened 2022-03-13
[BUG] Multiple cable blocks in SIMULINK


ckberinger:
**Describe the bug**
The cable block in Simulink currently only allows for one cable based on how it is written.
You see in yellow, it is asking for a cable body mass using "cable(2).bodyMass". But this calls my cable matrix row 2 which has nothing in it because when I make cable(1), cable(2), cable(3), cable(4) it is all in row one and spans the columns. 
![image](https://user-images.githubusercontent.com/61806038/158039370-a698c71d-ece0-4ff1-bb00-85b7ed8a5829.png)

**WEC-Sim file(s)**
Here is a zip of my file. I am trying to use the cable blocks as 4 taut mooring lines. Any other advice on my setup is appreciated.
[6DOF.zip](https://github.com/WEC-Sim/WEC-Sim/files/8238730/6DOF.zip)

**To Reproduce**
Run the WEC-Sim file provided from the command window.

**Expected behavior**
I get this error when trying to run with multiple cables. But it is only an error for cable 2 through 4 and not 1. 
Error using wecSim (line 46)
Error due to multiple causes.
Caused by:
    Error using wecSim (line 46)
    Error in 'LUPA14sim/Cable2': Failed to evaluate mask initialization commands.
        Error using wecSim (line 46)
        Invalid setting in 'LUPA14sim/Cable2/Follower Drag Body/Body Mass' for parameter 'Value'.
            Error using wecSim (line 46)
            Error evaluating parameter 'Value' in 'LUPA14sim/Cable2/Follower Drag Body/Body Mass'
                Error using wecSim (line 46)
                Index exceeds the number of array elements. Index must not exceed 1.

**Screenshots**
![image](https://user-images.githubusercontent.com/61806038/158039500-7096b7f7-be9f-4ef5-985b-61d9db33df1c.png)

**Workstation Information (this section MUST BE completed):**
 - OS: Windows 10
 - [MATLAB/Simulink Release] 2021b
 - [WEC-Sim Release] V4.4

**Additional context**
I tried to make a new cable block or edit the current one but I could not figure out how to do that. I made a new cable slx file, broke the library link, and tried editing but it always reverted back. I looked at how the body class works and it does not use "body(2).mass" it uses "body.mass" even for the second body which makes me think it is a typo in the cable block. 
How can I edit the simulink blocks that reference the library so I can continue debugging? 

Thank you,
Courtney

dforbush2:
[...]

PR #827, opened 2022-03-15
fix cable implementation

dforbush2:
This fixes issue #824 
Previously the mask initialization of the cable block used some 'eval()' syntax for setting the drag body mass. This fails when using multiple cables because this syntax will always try to read cable(1).cableNum, which is always 1, while the desired operation is to read cable(N).cableNum, which is always N. 

This has been replaced with 
'set_param(h,'Value', num2str(cable.bodyMass));'
for each the follower and the base drag body.

This is the only intended change. Tested on 2021b R1.


kmruehl:
@dforbush2 thank you! I'll pull this into PR #822 as well

PR #839, opened 2022-03-29
Remove deprecated functions from the BEMIO API

H0R5E:


H0R5E:
The docs branch build is passing but the final build test is failing because `dev` is in a broken state. This will fix that, so I am going to merge.

kmruehl:
@H0R5E thanks for resolving the issue. I made another commit to dev because the functions were renamed, not removed

PR #845, opened 2022-04-13
mask_variable_list

salhus:
This PR has a sample script in the RM3 example that can extract the mask variables. This can help develop some private debugging method in simulation class.


kmruehl:
@salhus it looks like this branch is on the main WEC-Sim/WEC-Sim repo, and not on your fork: https://github.com/WEC-Sim/WEC-Sim/tree/mask-variables. Can you please move this to your fork and open a new PR? Then we can figure out where to put this script after the v5.0 release.

salhus:
@kmruehl 
sorry about that. I will open a new PR

PR #846, opened 2022-04-13
enable aqwa sign convention

salhus:
AQWA uses a different convention to generate irregular waves using a spectrum import.
The WEC-Sim convention for spectrum import uses,
image
while the AQWA convention is,
image

This PR introduces a new property, waves.aqwaConvention, which when set to 1 will use the AQWA conventions to generate the irregular wave time history using an imported spectrum.


nathanmtom:
@salhus If Aqwa uses the sign convention of -\omega t rather than \omega t then we should correct the coefficients in BEMIO rather than changing the way the wave elevation is calculated within WEC-Sim. We had this same issue with NEMOH which is why in the diffraction coefficients we multiple the imaginary component by -1 to resolve the difference between -\omega t versus \omega t. Unless you are trying to compare the exact same time history between Aqwa and WEC-Sim.

This was an issue brought up previously by collaborators at Oregon State University and also by researchers at Plymouth who had reached out to Yi-Hsiang with questions before, but I do not recall them being resolved. This might confirm our initial thoughts on how the Aqwa notation follows NEMOH convention and not WAMIT convention.

salhus:
@nathanmtom  
I agree about the bemio correction. I asked @kmruehl recently about how the aqwa bemio treats this. I think I should go through that code and see how this issue is being treated.

This PR was to resolve issues I faced in a recent project. The user provided spectrum information using aqwa convention. I couldn't get the irregular wave time-histories to match. So after I made these tweaks I could get them to match.

Thanks for telling me about those issues, it will make it easier for me to see the context. I will update this thread after I have spent some more time on the AQWA and NEMOH bemio codes.

kmruehl:
Thanks @salhus and @nathanmtom. If I'm reading this correctly you'd like to close this PR and open a new one to make the revisions to BEMIO? Since there is a good discussion on this thread, it may be useful to just make the revisions on this PR and merge it before v5.0. 

Also, @salhus  it looks like this PR is on a branch on the main WEC-Sim/WEC-Sim repository: https://github.com/WEC-Sim/WEC-Sim/tree/aqwa-sign-convention, and not on your fork. In the future, please do all WEC-Sim development on your fork.

salhus:
@nathanmtom  and @kmruehl 

I ran bemio on the cylinder example for wamit, aqwa, and nemoh. I compared them using the compare feature. Unfortunately, the phases in aqwa have the opposite sign. NEMOH seems to have solved that issue like you indicated.

so here is a comparison of excitation force phases from wamit and aqwa

image

and here is the same comparison between the excitation phases of wamit and aqwa,

image


I am going to go check other ones as well. I will update this thread.

kmruehl:
@salhus  Thank you. Can you migrate this to your fork too? I'd like to keep this thread though. You should be able to amend the PR's base to your fork.

salhus:
@kmruehl 
Yes, of course. I have the files I need.
You can go ahead an delete this branch.

nathanmtom:
@salhus Thanks for sharing the phase of the excitation force. Given we would like to incorporate the AQWA changes into v5.0, if you compare the real and imaginary parts of the excitation force for the cylinder and the imaginary component is the opposite sign for AQWA than both WAMIT then I think it is safe to go ahead and apply the -1 modification in the `readAQWA` script.

salhus:
@nathanmtom 
Sorry about the delay.
I think it is a bit involved for modes involving motion in the x direction.  The motions in +-x, rotations about y, (surge, pitch, and roll) will have their signs change differently than heave. I have made a checker script that compares ex, fk, and sc  for magnitude/phase, and real/imag. And in the case of ex also the ex_K has a sign related issues.
Will update this thread.

salhus:
This PR is reintroduced as #855

Issue #836, opened 2022-03-20
CWR in regular waves versus irregular waves


jwills801:
**Is your question request related to a problem? Please describe.**
I am trying to simulate an OSWEC in both regular and irregular waves using the same PI controller. I am using a PM spectrum with waves.H = 2.5 meters and waves.T = 8 seconds. I use the same period and wave height when I simulate regular waves. The PI gains were tuned for the regular wave case using the optimization toolbox in matlab - specifically fminsearch - to maximize energy absorption. These gains were implemented using the pto(1).c and pto(1).k parameters in wecsim.

I am trying to compare the capture width ratio obtained in the regular wave case versus the irregular wave case. I am getting the wave energy resource from waves.Pw, and I am using 18 meters as the width of the WEC. I calculate the absorbed power by multiplying the PTO force by the pitch velocity. I calculate the capture width by dividing the average absorbed power by waves.Pw. Then, to find the capture width ratio, I divide the capture width by the device width.

I am getting a CWR of 76% for regular waves and 87% for irregular waves. I am also getting that the wave resource is 4 times larger in regular waves than irregular waves.

**Describe the theory or implementation approach that you have a question**
I am confused by two things. 
     First, why is there so much less energy in irregular waves for the same significant wave height and period? Is this the expected result or am I doing something wrong?
     Second, why would the CWR be larger in the irregular wave case? I tuned the PI gains for the regular wave case so I thought that it would preform best at this specific frequency, and worse at other frequencies. I thought that in irregular waves, some of the energy would be in frequencies other than the one I optimized for, and my PI controller could be less effective at absorbing the energy in these frequencies. I figured that this suboptimal behavior at other frequencies would result in a smaller capture width ratio - which I have taken as a kind of stand in for efficiency.

**Personal project context**
I am still trying to figure out what is going on in WEC-sim and I have some confusions I need help in sorting out. I am researching wave energy in school.

**Describe the type of conclusion or resolution you are looking for**
I am hoping to achieve more understanding. If these results make sense, I would like to understand the rational. If these results do not make sense, I would like to figure out what I m doing wrong in my implementation.

**Additional context**
N/A

nathanmtom:
@jwills801 Thank you for your issue and I will do my best to answer your questions. 

[...]

PR #847, opened 2022-04-15
Update the WEC-Sim Theory Documentation to Clarify Wave Power Calculation

nathanmtom:
Issue #836 highlighted an issue in the calculation of wave power within WEC-Sim for regular waves that were not in deepwater. In order to improve the user understand of wave power calculated within WEC-Sim additional documentation in the regular and irregular theory sections have been added.


kmruehl:
@H0R5E I can't figure out what is causing the docs to trigger the following warning: 
``` /home/runner/work/WEC-Sim/WEC-Sim/docs/theory/theory.rst:376: WARNING: Duplicate explicit target name: "here".``` but we need to merge this PR, so I'm going to remove the `-W --keep-going` requirement from `sphinx-build -W --keep-going -b html docs docs/_build/html`.

H0R5E:
@kmruehl, I see the issue. I think the ideal solution is split this into two, with a stricter check stage and a more permissive build stage. In this way the CI will tell you if something is wrong but won't prevent the docs from building if you choose to merge the docs in a broken state. What do you think?

kmruehl:
@H0R5E I'm okay with that suggestion. I think failing on a warning is a bit restrictive, but I'm open to this middle ground solution.

PR #852, opened 2022-04-20
extract-mask-variables

salhus:
This PR has a script that can extract mask variables to debug errors in mask intialization.


salhus:
@kmruehl @nathanmtom 
I introduced a new folder developmentTools for the non-essential functionality in this PR.
This is ready from my end.

PR #855, opened 2022-04-22
Aqwa-phase-bug-resolved

salhus:
Added a line to add a phase offset to the phase data extracted from aqwa to account for the different sign conventions used.

Thus follows from #846

### DIFF
--- a/examples/BEMIO/compareBEMIO.m
+++ b/examples/BEMIO/compareBEMIO.m
@@ -1,0 +1,114 @@
+%% BEMIO example comparing different hydro data 
+
+%% Sphere (1 body)
+clear all;clc; close all
+
+% Load WAMIT hydro data 
+WAMIT_hydro = struct();
+WAMIT_out = '.\WAMIT\Sphere\sphere.out';
+WAMIT_hydro = readWAMIT(WAMIT_hydro,WAMIT_out,[]);
+WAMIT_hydro = radiationIRF(WAMIT_hydro,15,[],[],[],[]);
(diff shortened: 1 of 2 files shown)


kmruehl:
Thanks @salhus !

tuisong1106:
Dear Salhus, the phase of excitation force read by this program is not consistent with the value in LIS file. Could you please help test whether the program needs to be improved? Thank you.

salhus:
@tuisong1106 
Could you please share some examples, with if possible the time histories?
That will help me understand the potential issue.

tuisong1106:
> @tuisong1106 您能否分享一些例子，如果可能的话，包括时间历史？ 这将帮助我理解潜在的问题。

I used AQWA to calculate the interaction between the two cylinders, and the reading program found that the excitation force was not matched, and the additional mass and damping coefficient were opposite in the upper right corner (the influence of 1 on 2) and the lower left corner (the influence of 2 on 1).

tuisong1106:
> @tuisong1106 Could you please share some examples, with if possible the time histories? That will help me understand the potential issue.

I used AQWA to calculate the interaction between two cylinders, and the reading program found that the excitation force was not matched, and the added mass and radiation Damping were opposite in the upper right corner (the influence of 1 on 2) and the lower left corner (the influence of 2 on 1).

salhus:
@tuisong1106 
Thanks for describing the issue.
However, it will help me understand if you can share the actual files .
The hydrodynamic couplings are typically symmetric matrices so I am not sure I followed what opposite implies.


Cheers,
Sal

kmruehl:
Thanks @salhus

PR #856, opened 2022-04-25
Remove fixed mass option

akeeste:
[Project card ](https://github.com/WEC-Sim/WEC-Sim/projects/58#card-79563001)

The bodyClass fixed mass option can be somewhat misleading if there is not actually a fixed constraint. If the body isn't fixed it will likely sink indefinitely and become unphysical. Also, if a fixed constraint is connected to the body, ``body.mass='fixed'`` is not required and its mass doesn't affect the simulation.

This PR removes the ``body.mass='fixed'`` option, the associated logic, and a small related section in the documentation. For after v5.0


kmruehl (approved):
This looks good. Making a minor change to the doc, then I'll merge.

PR #858, opened 2022-04-26
Combine rigid flex bodies

akeeste:
In response to a [project card](https://github.com/WEC-Sim/WEC-Sim/projects/58#card-64210449) and discussion on how to manage the body library, this PR creates works to combine the rigid and flexible blocks. The flexible subsystem is completely encapsulated into one extra variant subsystem that is activate if ``body.flex=1``, which is determined by the hydrodata read from the .h5 file.

This removes many links between the rigid body and flex body blocks, as well as a few redundant functions that could not be linked (excitation force, hydrostatic restoring force). 

- [x] Moved flex body to a Body variant subsystem. Rigid body functions the same when ``body.flex=0``
- [x] Updated forwarding table so that all previous ``Rigid Body`` and ``Flex Body`` blocks redirect to the single ``Body`` in the Body Elements library
- [x] Added checks to ensure that passive yaw and B2B is not used with GBM. This could be made functional, but is not as of yet.

For after v5.0


kmruehl:
Two Perspectives to this review:
- SCM perspective, how can we better manage the broken library links between hydro and flex bodies?
- User perspective, what makes sense for users? Should we have multiple body blocks?

kmruehl:
This is a larger team discussion on how we want to restructure the body library. Let's table this for now.

akeeste:
Closing this PR in favor of the project board item

Issue #864, opened 2022-05-11
[Developer Issue]


SimonMatdp:
**Project**
_responseClass.saveViz_

**Describe the issue**
When trying to execute responseClass.saveViz for the RM3, several errors are found:
Problem 1:  Unrecognized field name "faces"  Error in responseClass/saveViz (line 466)
Solution 1: change faces to face

problem 2: Index in position 2 exceeds array bounds. Error in responseClass/saveViz (line 472)
                if max(bodyMesh(ibod).Points(:,3))+max(bodyMesh(ibod).deltaPos(:,3)) > maxHeight
                This is due to both body.geometry.vertex and body.geometry.face being empty

solution 2: in lines 465 and 466 change 
               bodyMesh(ibod).Points = body(ibod).geometry.vertex; 
               bodyMesh(ibod).Conns = body(ibod).geometry.face;
                                              to 
                readBodyMesh = stlread(body(ibod).geometryFile); 
                bodyMesh(ibod).Points = readBodyMesh.Points; 
                bodyMesh(ibod).Conns = readBodyMesh.ConnectivityList;

jtgrasb:
Thanks for pointing out this issue and providing a solution! Your solution has been implemented in PR #866.

PR #866, opened 2022-05-17
Fix saveViz

jtgrasb:
This PR is in reference to Issue #864 and updates the saveViz feature to work with the v5.0 release.

### DIFF
--- a/source/objects/responseClass.m
+++ b/source/objects/responseClass.m
@@ -462,8 +462,9 @@ classdef responseClass<handle
             % Read in data for each body
             for ibod=1:length(obj.bodies)
                 % Read and assign geometry data
-                bodyMesh(ibod).Points = body(ibod).geometry.vertex;
-                bodyMesh(ibod).Conns = body(ibod).geometry.faces;                
+                readBodyMesh = stlread(body(ibod).geometryFile);
+                bodyMesh(ibod).Points = readBodyMesh.Points;
+                bodyMesh(ibod).Conns = readBodyMesh.ConnectivityList;              
                 % Read changes and assign angles and position changes over time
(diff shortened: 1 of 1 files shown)


source/objects/responseClass.m
```diff
@@ -462,8 +462,9 @@ function saveViz(obj,simu,body,waves,options)
             % Read in data for each body
             for ibod=1:length(obj.bodies)
                 % Read and assign geometry data
-                bodyMesh(ibod).Points = body(ibod).geometry.vertex;
-                bodyMesh(ibod).Conns = body(ibod).geometry.faces;                
+                readBodyMesh = stlread(body(ibod).geometryFile);
```
akeeste, line 465: Alternate option:
   
                if isempty(body(ibod).geometry.vertex)
                    % Import body geometry for linear hydro cases
                    body(ibod).importBodyGeometry(simu.domainSize);
                end
                bodyMesh(ibod).Points = body(ibod).geometry.vertex;
                bodyMesh(ibod).Conns = body(ibod).geometry.face;

akeeste (changes requested):
@jtgrasb Thanks for this fix, this works fine as is. But let's use the import body geometry function that is already in the bodyClass. This way we can also take advantage of the geometry checks that are contained there. Then we can use the geometry.vertex and geometry.face as is. Alternatively we could run ``importBodyGeometry`` in the pre-processing for all cases, instead of just for the nonlinear hydro ones.

I'll change this to be into master as well since it's a bug fix.

Adam

source/objects/responseClass.m
jtgrasb: Thanks for the suggestion @akeeste! I do think it may be useful to import geometry for all cases to have those checks. We can discuss further with the team in our next meeting.

PR #870, opened 2022-05-23
wecSimPCT Fix (Master)

yuyihsiang:
Fixed a few bugs, which cause WEC-Sim to crash when running wecSimPCT:
- "initializeWecSim.m": Change the project directory and move the "temp" inside the pctDir folder so that each WEC-Sim run won't interfere.
-  "stopWecSim.m": Move the command cd (['..' filesep pctDir filesep '..' filesep]); inside the "if simu.saveWorkspace==1" if loop. The command is only needed when WEC-Sim writes out the output file.
-  "wecSimPCT.m" :  Remove the pctDirs after the simulation.

### DIFF
--- a/source/functions/initializeWecSim.m
+++ b/source/functions/initializeWecSim.m
@@ -28,11 +28,18 @@ diary('simulation.log')
 %% Add 'temp' directory
 % Store root directory of this *.m file
 projectRootDir = pwd;
+if exist('pctDir')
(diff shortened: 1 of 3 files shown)


kmruehl (approved):
Thanks @yuyihsiang. I haven't been able to reproduce this bug, so we think it may be a MAC OS bug, but I was able to run this version without issue, so I'll go ahead and merge it.

PR #874, opened 2022-05-30
[Bug fix] Mac path fixes and make outputDir public

ahmedmetin:
Small suggestion to have outputDir public instead of private and Fixed paths in CompareBEMIO.m so they work both on Windows and Mac.


source/functions/BEMIO/readAQWA.m
```diff
@@ -35,7 +35,7 @@
 
 hydro(F).code   = 'AQWA';
 V182            = 0; % Set AqwaVersion flag "Version 18.2 or larger" to 0
-tmp             = strsplit(ah1Filename,{' ','\','.'});
+tmp             = strsplit(ah1Filename,{' ','\', '/', '.'});
```
akeeste, line 38: Update using MATLAB ``filesep`` to be most robust:

    tmp             = strsplit(ah1Filename,{' ', filesep, '.'});

akeeste (commented):
Hi @ahmedmetin 

Thank you for these fixes. I don't have an issue making ``outputDir`` public to allow users to specify a new output directory.

Your path updates still work for me on Windows. I'm happy with updating these paths to use ``fullfile()`` as that is typically very robust across various OS. Do you also have issues running the following examples on Mac? If so, we should update the relevant ``bemio.m`` files as well.
- examples/RM3
- examples/BEMIO/Capytaine
- examples/BEMIO/NEMOH

Lastly, does ``readAQWA`` fail on Mac without this update? If so we can keep this into the master branch as it is a bug fix.

Adam

ahmedmetin:
Hello,
I think it is better if the user can specify the outputDir, because the user might want to integrate WECSim to their existing software framework or work flow. But it is up to you of course. :)

Yes, all these examples fail on a Mac, because Mac uses / as the file separator, so suggest to replace them either with filesep or fullfile().

readAQWA() does not fail without this change however the intended behaviour that is on Windows is not achieved, i.e. it is not consistent on different platforms. On Mac hydro.file takes the full path to the file without the . in front, while on Windows it is equal only to the filename without the path. This creates and error later when creating a .h5 file, because it does not want to create the .h5 file in a different folder than you are at for some reason (I think it requires the ./ at the beginning).

Best regards,
Ahmed

akeeste:
Hi @ahmedmetin

Sorry for the delay. Thank you for this updates for WEC-Sim on Mac. Once I confirm that these changes work with out applications repo I will merge this PR.

Adam

Issue #876, opened 2022-06-02
[Developer Issue] Simulation initialization fails w/ subfunction calls


dforbush2:
**Project**
Associated with In Progress/Wave Phase correction w/ mean translation

**Describe the issue**
W/ in simulink, to facilitate changes and version tracking, matlab function blocks have syntax like so:

```
function [outputs]= functionName(inputs);
% initialize outputs
outputs = 0;
% call to external *m function file functionName.m
[outputs]= functionName(inputs);
end
```

When developing this feature, I have a feedback loop output-to-input broken w/ a memory block to keep track of the position of the last wave phase adjustment.

If I maintain our standard syntax as above, the simulation fails to initialize b/c it does not know how to size dispLast, and it "is inferred as a variable-size parameter". (Designating it as variable-size causes other problems, and, because it will be the same size w/in a simulation, shouldn't be necessary). What is odd is that an equivalent structure is used in the yaw excitation blocks and seems to work fine. The only exception is that the source of the pass-through variable in that case is a constant and not an in-port. 

If, however, include the *m file directly in the matlab function block (as opposed to referencing the external .m), the simulation runs correctly. Because one of these blocks exists in every version of the wave excitation block (8-ish in total), this is NOT the preferred development pathway, but I can't figure out how to get it functional w/ in our preferred syntax with an external *m file. 

Open to any advice on this. I will also initiate a pull request for anyone developer interested in playing w/ this. Issue will be closed when resolved satisfactorily in the pull request.

PR #877, opened 2022-06-02
initial commit largeXYDispOption

dforbush2:
This is the initial pull request for the wave excitation correction for large XY displacement project. The wave excitation force is calculated with respect to a body's instantaneous X,Y position, not just the force at 0,0. It is appropriate to model bodies for which large lateral displacements are expected, particularly where wave phase between multiple bodies is important. 

Files Changed:
bodyClass: added 1 new structure "largeXYDisplacement" with 2 fields: "option" (1 or 0 boolean to turn on/off) and "threshold" (float, in m: how far the body can translate before an excitation phase correction will be recalculated). 
body block wave excitation: added phase arguments to regular waves. Added matlab function block calling calcPhaseDisp to all body/wave excitation calculation methods except no wave and user defined elevation, and the displacement input to all methods for which it did not already exist. The sum of phase from waves.phase and the displacement phase correction is passed to the excitation force calculation. 

Files Added:
functions/simulink/ model/calcDispPhase.m: the excitation phase calculating function (see comments for I/O description)

To test: the RM3 example works, just add the following where appropriate in the input file
```
body(1).largeXYDisplacement.option=1;
body(1).largeXYDisplacement.threshold=1;
body(2).largeXYDisplacement.option=1;
body(2).largeXYDisplacement.threshold=1;
```


dforbush2:
DO NOT MERGE YET. See issue #876

dforbush2:
I have also tried using persistent variables in the same way as the radiation convolution integral. Again, this is giving an error, even though it appears to operate the exact same way as the convolution integral block.

[MATLAB Function3](matlab:Simulink.ID.hilite('RM3:824:4071:6694')) uses constructs that are invalid when the block specifies or inherits a continuous sample time. Invalid constructs include using global variables, writing to initialized persistent variables, calling exported functions, or using coder.extrinsic. For help on persistent variables, see [the documentation](matlab:sfhelp('initialize_persistent_variables')).

dforbush2:
Because of the unresolved issues in the items above, and their apparent dependency on MATLAB/Simulink version, I have removed the threshold for re-calculation. If the largeXYDisplacement option is enabled, it will calculate a phase correction at every time step. This is a single additional matrix multiplication: for irregular waves, a few timing tests with the RM3 example showed no significant difference in run time with and without this option enabled. This is not ideal, but removes the need for feedback loops or persistent variables in the MATLAB function.

Still need to  address failed tests and add documentation.

dforbush2:
All checks now passing. Need docs, but otherwise ready for merge!

examples/RM3/wecSimInputFile.m
```diff
@@ -14,20 +14,21 @@
 % waves = waveClass('noWaveCIC');       % Initialize Wave Class and Specify Type  
 
 % Regular Waves  
-waves = waveClass('regular');           % Initialize Wave Class and Specify Type                                 
-waves.height = 2.5;                     % Wave Height [m]
-waves.period = 8;                       % Wave Period [s]
+% waves = waveClass('regular');           % Initialize Wave Class and Specify Type                                 
+% waves.height = 2.5;                     % Wave Height [m]
+% waves.period = 8;                       % Wave Period [s]
```
jtgrasb, line 19: Should we leave this as regular waves for our example case?

examples/RM3/wecSimInputFile.m
```diff
@@ -63,12 +64,16 @@
     % Body Mass. The 'equilibrium' Option Sets it to the Displaced Water 
     % Weight.
 body(1).inertia = [20907301 21306090.66 37085481.11];  % Moment of Inertia [kg*m^2]     
+body(1).largeXYDisplacement.option = 0;
```
jtgrasb, line 67:
```suggestion
body(1).largeXYDisplacement.option = 0; % Change to 1 to account for large XY displacements
```

jtgrasb (commented):
Just reviewed this and looks good. I also tested it with the mean drift application and there is clear difference between the case without the displacement calc and with the displacement calc, which can be seen when plotting/saving the visualization. The displacement calc accounts for the x translation and the body motion appears to be much more in phase with the waves as it translates. Docs look fine too, looks ready to merge to me!

examples/RM3/wecSimInputFile.m
dforbush2: Hi @jtgrasb 

Thanks for your review!

I didn't make a dedicated example for this yet: I only modified the existing RM3 example for the purposes of testing. On those grounds, I would prefer to leave the RM3 example file as it was, without any of my changes, since it is meant to be an approachable "first time user" sort of example. I should make an example file for WEC-Sim Applications, but I'll do that in a separate PR. 

Were you able to build and display the docs properly? I was having trouble on my local machine but that is more likely my own faulty Python environment than anything I did in the *rst.

jtgrasb: @dforbush2 

Makes sense on the RM3 example, just wasn't sure since it was in the PR. 

Yeah, docs built and looked fine!

examples/RM3/wecSimInputFile.m
kmruehl: I'm reverting this input file back to the RM3 example, but this version will remain on Dom's branch on a prior commit.

examples/RM3/wecSimInputFile.m
kmruehl: same comment. I'm reverting this for the merge, but it'll be on Dom's fork in a prior commit.

PR #880, opened 2022-06-08
Ainf fixes for hydro and h5 files

salhus:
There are some situations when the WAMIT output files do not have the $A_\infty$ term. Like discussed here,
image

This PR modifies the readWAMIT.m and writeBEMIOH5.m files to use the last term of the added-mass as a substitute instead.


nathanmtom:
@salhus Thanks for trying to find a way to address the situations where the infinite frequency added mass is not calculated in WAMIT with the appropriate free surface boundary condition. 

I am wondering why you chose to take the highest frequency added mass to represent the infinite frequency added mass? In the 'radiationIRF.m', lines 68-82 we use the expression for the inverse Fourier Transform of the time domain IRF to estimate the infinite frequency for NEMOH as the code also has issues with solving for the infinite frequency added mass. 

My thought was that this process could also be used for cases where WAMIT does not report the infinite frequency added mass, but perhaps you have thoughts on this.

kmruehl:
@salhus can you convert this to a draft until it's ready for review?

Issue #875, opened 2022-06-01
readCAPYTAINE not working for current Capytaine .nc output files [BUG]


jtgrasb:
**Note: italicized text below is include as an example and should be updated before submission. If you feel any section is not applicable to your request, please replace with 'N/A' rather than delete a section.**

**Describe the bug**
When Capytaine is run, it outputs a .nc file with the desired hydrodynamic data. readCAPYTAINE is run to read in the hydro data, but gives errors due to the name and format of the complex indices. This seems to be an error related to the current formatting of the .nc file. I'm not sure if an update within Capytaine is leading to differences in formatting or if it's MATLAB's ncread script.

**WEC-Sim file(s)**
readCAPYTAINE.m

**To Reproduce**
_Steps to reproduce the behavior:_
1. Open the anaconda command prompt.
2. Navigate to the WEC-Sim directory, then to examples->BEMIO->CAPYTAINE, and go into one of the specific example folders (ex: sphere)
3. Type python sphere.py to run the sphere case. 
4. Within the Capytaine->Sphere folder, open the BEMIO script and change to filename from '.\sphere_full.nc' to '.\sphere.nc', and click run. 

**Expected behavior**
BEMIO should run normally and read in the hydro data from the .nc file.

**Screenshots**

**Workstation Information (this section MUST BE completed):**
 - OS: [Windows 10]
 - [MATLAB/Simulink Release](https://www.mathworks.com/help/matlab/release-notes.html) [2021b]
 - [WEC-Sim Release](http://wec-sim.github.io/WEC-Sim/man/release_notes.html) [v5.0]

PR #884, opened 2022-06-13
Fix readCAPYTAINE

jtgrasb:
This PR addresses issue #875. The issue has 2 parts:

1. Body name string vs. cell array formatting

- Old .nc files have body names formatted as strings while new ones have body names formatted as cell arrays
- Fix: The simple fix for this is to unpack each element of the body names.

2. Complex indices format

- Old .nc files have horizontal vectors of for specifying order of the complex dimension, while new ones have vertical vectors.
- Fix: Check whether name is a string to check version, and read in order of the complex dimension based on new vs. old version.

### DIFF
--- a/source/functions/BEMIO/readCAPYTAINE.m
+++ b/source/functions/BEMIO/readCAPYTAINE.m
@@ -88,6 +88,15 @@ tmp = ncread(filename,'body_name')';
 for i=1:s1
     hydro(F).body{i} = erase(tmp(i,:), char(0)); % assign preliminary value to body names
 end
+
+hydro(F).body = hydro(F).body{:};
+% Check for newer formatting
+if isstring(hydro(F).body)
+    ncFormat = 'new';
+else 
+    ncFormat = 'old';
+end
+
 hydro(F).body = split(hydro(F).body,'+');
 
 % sort radiating dof into standard list if necessary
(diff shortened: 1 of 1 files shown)


akeeste:
@jtgrasb Yes I used Capytaine v1.3 with the [demo.py](https://github.com/WEC-Sim/WEC-Sim/blob/master/examples/BEMIO/CAPYTAINE/demo.py) file as is. However I still don't see any changes to how the body name or real/imaginary indices format in ``readCapytaine``. Before merging this PR, I would like to better pinpoint what option causes this change and fix it on the Capytaine side if possible.

Adam

jtgrasb:
@akeeste This is strange. I have checked a lot of potential sources for this error, but can't seem to figure it out. 
- The .nc files already in our repo (sphere_full.nc, etc.) work fine with our current readCapytaine script.
- Any new runs using the cases already set up without any changes (sphere.py, demo.py, etc.) lead to the "new" format with changes to the body name and real/imaginary indices.
- We are both using Capytaine v1.3 so that doesn't seem to be the issue (I also tried updating to v1.4 and it didn't change anything) 
- I am using MATLAB 2021b, but I also tried MATLAB 2020b and actually got a different error shown below:
image

Let me know if you have any ideas where these error might be coming from.

Thanks,
Jeff

akeeste:
@jtgrasb Thanks for the update. I'm not sure where our workflows are differing or why R2021b and R2020b give different errors. At the next team meeting let's assign a couple different developers to try and recreate this issue and pin down what is happening.

mancellin:
> We are both using Capytaine v1.3 so that doesn't seem to be the issue (I also tried updating to v1.4 and it didn't change anything)

Nothing changed in these modules of Capytaine for years.

I expect the change to come from either `xarray` or the library used in the backend by `xarray` to export the netCDF file.
[The choice of a backend engine by `xarray` depends on what is installed in your environment.](https://github.com/pydata/xarray/blob/3f7cc2da33d81e76afbfb82da57143b624b03a88/xarray/core/dataset.py#L1845-L1848)

Maybe the "old" format could be reproduced using a conda environment with an old version of `xarray` and/or explicitely setting the `engine` argument of `to_netcdf`.

kmruehl:
@akeeste and @jtgrasb I can confirm this issue with the body name. I just reran the OSWEC case in Capytaine v1.4, and ran BEMIO in 2022a, and got the following error:

``` matlab
Error using split
First argument must be text.

Error in readCAPYTAINE (line 91)
hydro(F).body = split(hydro(F).body,'+');

Error in bemio (line 4)
hydro = readCAPYTAINE(hydro,'oswec.nc');
 
```

kmruehl:
I can confirm that this PR resolves the issue for v1.4, and also works for the prior v1.3 release. I'm going to go and merge it so that dev is compatible with Capytaine v1.4 . 

@jtgrasb and @akeeste please open a new PR if there are other issues that need to be resolved.

PR #896, opened 2022-06-26
Fix image bug in PTO-Sim in Library Browser

jleonqu:
There is a bug in the library browser and the images for the PTO-Sim library are not shown. This PR fixes this bug. In addition to solving the bug, new images for the electric generator models are added to the library.


kmruehl (approved):
@jleonqu thank you, this works for me. Merging now!

Issue #873, opened 2022-05-25
[Theory or Implementation] Running WEC-Sim and using elevationImport


Aiswariak:
I have been using WEC-Sim 4.2 version and has now updated to 4.4 version. My MATLAB is R2021b. When I am trying to run wecsiminputfile.m, I get an error message as follows:

Unrecognized function or variable 'simFile'.

Error in wecSim (line 46)
sim(simFile, [], simset('SrcWorkspace','parent'));

I am reposting this as per your suggestion @ https://github.com/dforbush2
https://github.com/WEC-Sim/WEC-Sim/issues/872#issue-1246015653

The same code works fine with the older version. I could also run RM3 tutorial successfully. Kindly help
![wecsiminputfile](https://user-images.githubusercontent.com/87517969/170194226-74f0d913-69b5-40fd-a248-189f9ed7eb82.jpg)

Aiswariak:
Hi @ [dforbush2](https://github.com/dforbush2),
I believe that I have  properly updated the matlab path and can run RM3 tutorial successfully too.

The line 45 of wecSim.m is:

`sim(simuFile, [], simset('SrcWorkspace','parent'))'
Does this mean that my path is still pointing to old WEC-Sim installation?

I tried changing this line to :
`sim(simu.simMechanicsFile, [], simset('SrcWorkspace','parent'))'. 
Now its running and I am getting the output folder. Still it shows some error at the end as follows:

Error using wecSim (line 46)
Error evaluating 'StopFcn' callback of Frames block (mask) 'RHP1610/Global Reference Frame'.
Callback string is 'run('stopWecSim')'
Caused by:
    Error using wecSim (line 46)
    Unrecognized function or variable 'projectRootDir'.

Does this effect my results?

I have attched the zip folder for your reference.

[T=1.6s, H=0.10m.zip](https://github.com/WEC-Sim/WEC-Sim/files/8768383/T.1.6s.H.0.10m.zip)

dforbush2:
I'm browsing the repository at the v4.4 release. In wecSim.m I see:

```
% Initialize WEC-Sim
run('wecSimInputFile');
simFile = simu.simMechanicsFile;
clear simu waves body cable pto constraint ptosim mooring 
clear simu waves body pto constraint ptosim mooring 

runWecSimCML = 1;
run('initializeWecSim');
sim(simFile, [], simset('SrcWorkspace','parent'));

```

[...]

PR #907, opened 2022-07-07
Dev: adding ampSpectraForWS function

dforbush2:
This is related to issue #873 . While we recommend spectrumImport instead of elevationImport as a more robust solution method for user-defined wave time series, we previously have not provided a tool to convert the elevationImport.mat --> spectrumImport.mat. The addition of this single function fixes that.


akeeste:
@dforbush2 I tested the function with a sample irregular wave. It works well and gives very similar results to MHKiT. I don't have any issue with the function, but prefer to encourage users to use MHKiT for this purpose. MHKiT already has tests and documentation built in as well. If users have MHKiT-MATLAB installed, it would be straightforward to use their ``elevation_spectrum`` function.


pr907 amp spectrum

akeeste (changes requested):
@dforbush2 

Per my previous comments, I made a PR into your dev branch with a few updates, including:
- remove WEC-Sim ``ampSpectraForWS`` funtion
- add to your documentation update pointing users to MHKiT
- remove changes to ``source/functions/simulink/model/nonLinearBuoyancy.m`` which i believe is meant for your other PR #910

After these changes we can merge this PR

akeeste:
TODO for Adam:
- [ ] also compare complex amplitude of the spectra and compare how the timeseries elevation of each spectra comes out in WEC-Sim

kmruehl:
@dforbush2 @akeeste is this ready for a merge?

Issue #886, opened 2022-06-15
Issue in Nonlinear Buoyancy and Froude-Krylov Wave Excitation Tutorial - Heaving Ellipsoid


hamzaJayy:
Dear Developers,

I have doubts in simulation results of WEC-sim application of non-linear hydrodynamics for a Heaving Ellipsoid.

I ran simulation using the settting and files provided by WEC-sim application master for a heaving ellipsoid for 4 different cases, which are given below:

1. Linear hydrodynamics with PTO
2. Linear hydrodynamics without PTO
3. Non linear hydrodynamics with PTO
4. Non linear hyrdrodynamics without PTO

Here, PTO is a simple damper. Without PTO means making the damping coefficient zero in wecsim input file. I was expecting that wecsim will calculate the response of the ellipsoid using non-linear hydrodynamics when the body(1).nlHydro =2 no matter there is a PTO or no PTO. But this is not the case. The system behaves linearly even the body(1).nlHydro =2 with PTO  turned off (as it can be seen in graphs below)

All other settings kept constant. Regular wave with h=12.2m and T=16.61 is simulated with simu.wave('regular').

Is there something I am missing about nonLinearHydro formulation using Froude-kryolov force? Please guide me what is happening. 

I have attached the output graphs of Heave, restoring force and force exciatation for you reference in the same order below: 
![image](https://user-images.githubusercontent.com/106804145/173722152-ad1a8ad4-6972-4bb2-9697-95ca544a9baf.png)

![image](https://user-images.githubusercontent.com/106804145/173720659-39c4d328-e1d9-4480-baec-2576bbfd56a5.png)

![image](https://user-images.githubusercontent.com/106804145/173720691-d89c5573-4e89-420f-98de-aa8e40e6f458.png)

**Workstation Information (this section MUST BE completed):**
 - OS: , Windows 10
 - MATLAB/Simulink Release, 2021b
 - WEC-Sim Release v4.4

I am not sure where is the issue. But it does not make sense to me for now. I am facing another issue with nonLinearHydro that I asked in #878 . I guess this issue is related to theory implementation.

dforbush2:
Hi @hamzaJayy 

Thanks for bringing this to our attention. This is a bug: the nonlinear hydro implementation is not properly initializing for some reason in all cases. Note that the shown results are not in absence of a PTO, but with pto.c = 0 . If the PTO block is removed and the system is thus fixed, everything does appear to be working properly.

[...]

PR #910, opened 2022-07-13
fix non-linear hydro

dforbush2:
This is in regard to issue #886 to revise the nonLinearHydro calculation to use the difference between body position and wave elevation, rather than just body position. This affects both non-linear hydro option 1 and 2.


dforbush2:
Some figures for reference: This fixes the transient over-estimation of restoring force (in excess of saturation values) for all cases. 
T16Compare
 
For zero damping, what #886 called 'without PTO', we see that there is some "ringing" near the saturation value. For this non-physical case with no damping, this "ringing" represents the change in hydrostatic equilibrium due to the changing wave elevation and panels at a given Z entering/exiting the water, changing the restoring force in a step-wise fashion. While this is non-physical, it is an expected numerical artifact and suggests this implementation is correct. 
T16C

salhus:
Hi @dforbush2 and @kmruehl ,

Sorry about the delay. I checked the issue in the restoring forces. And it looks all good to me. I would recommend this to be merged.

dforbush2:
This PR introduced a bug in the FKForce calculation and should be reverted.

PR #911, opened 2022-07-14
update to v5.0 citation

akeeste:
Update to v5.0 citation
Deferring to @kmruehl on author order.


kmruehl (approved):
Thanks. I moved the citation page to the release page, plus it'll make it easier to remember to update both.

Issue #913, opened 2022-07-18
[Feature request] Update Capytaine BEMIO examples to v1.5


kmruehl:
**Is your feature request related to a problem? Please describe.**
N/A

**Describe the solution you'd like**
[Capytaine v1.5](https://github.com/mancellin/capytaine/releases/tag/v1.5) was recently released. WEC-Sim's examples are based on a prior version of Capytaine, and should be updated to v1.5. This will remove the dependency on Meshmagick for hydrostatic calculations. 

**Describe alternatives you've considered**
N/A

**Personal project context**
N/A

**Additional context**
Updating the WEC-Sim examples to Capytaine v1.5 should eliminate some of the prior package dependencies and ease Capytaine installation for users.

kmruehl:
Updated issue to reflect latest Capytaine release. @dav-og  and @salhus  are working with @mancellin  to update Capytaine to support multi-body, and then WEC-Sim examples

dav-og:
just opened a Capytaine PR to include a Hydrostatics.dat and KH.dat file writer: https://github.com/capytaine/capytaine/pull/285

once this is approved we can update the Capytaine BEMIO examples 👍

Issue #188, opened 2017-07-01
Running examples - Error using h5readc


aritradatta:
We are trying to run the examples but we are running into the error below when we run the 'wecSim' command:

WEC-Sim Read Input File ...   
Error using h5readc
Unable to open file. Filename may be corrupt or have unsupported characters.

Error in h5read (line 58)
[data,var_class] = h5readc(Filename,Dataset,start,count,stride);

Error in bodyClass/readH5File (line 81)
            obj.cg = h5read(filename,[name '/properties/cg']);

Error in wecSim (line 85)
        body(ii).readH5File;

We noticed that in the readme file it is mentioned that "The *hf5 needs to be updated from BEMIO to include StateSpace" for the RM3 example. Can you please help us understand the steps involved? We are relatively new to this software. Any help will be greatly appreciated!

EDIT after doing some debugging it seems that matlab is having an issue with opening oswec.h5 .. We have tried to run matlab as admin and that didnt help.

aritradatta:
EDIT after doing some debugging it seems that matlab is having an issue with opening oswec.h5 .. We have tried to run matlab as admin and that didnt help.

kmruehl:
I just ran the OSWEC tutorial and had no issues loading the oswec.h5 files. I've attached the file i ran. What version of MATLAB are you running? Are you sure you have the *.h5 file in the correct path? The ReadMe note is only relevant if you're running WEC-Sim with state space, it will not cause an error when importing the *h5 file to matlab. 


[oswec.zip](https://github.com/WEC-Sim/WEC-Sim/files/1125319/oswec.zip)

aritradatta:
@kmruehl we found the issue - we downloaded the repo from GitHub as a zip but for some reason the h5 files didn't download properly. For some reason we ended up with a ~1KB file with a Git hash, instead of the original ~7MB file. Replacing the file correctly solved our issue! Apologies for the confusion and thank you for your help :)

kmruehl:
I'm glad you were able to resolve this issue. We've taken note of this, and will add a README (or similar) to remind users to check their H5 files.

PR #925, opened 2022-07-29
updating readCapytaine to include hydrostatics 

kmruehl:
This responds in part to #913 to upgrade BEMIO examples to Capytaine v1.4. In order to do that, we also need to remove BEMIO's dependency on MeshMagick hydrostatic data, and use Capytaine's hydrostatic calculations directly.

### DIFF
--- a/source/functions/BEMIO/readCAPYTAINE.m
+++ b/source/functions/BEMIO/readCAPYTAINE.m
@@ -63,7 +63,7 @@ req_vars = {
     'disp_volume',...
     'center_of_mass',...
     'center_of_buoyancy',...
-    'hydrostatic_stiffness'...
+%     'hydrostatic_stiffness'...
     };
 
 % Other vars in Capytaine .nc file not currently used:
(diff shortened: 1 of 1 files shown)


kmruehl:
@mancellin right now the `*.nc` file from Capytaine stores `hydrostatic_stiffness`, but not `disp_volume`, `center_of_mass`, or `center_of_buoyancy`. WEC-Sim needs all of them. Do you want those those parameters added to the Capytaine output by default? If not, I'll update the BEMIO examples on the WEC-Sim repo to include something like this:

``` python 
capyData = cpt.assemble_dataset(results,hydrostatics=True)

capyData = capyData.assign_coords({"disp_volume":hydrostatics['disp_volume']})
capyData.expand_dims({'center_of_mass':3})
capyData = capyData.assign_coords({"center_of_mass":np.array(hydrostatics['center_of_mass'])})
capyData.expand_dims({'center_of_buoyancy':3})
capyData = capyData.assign_coords({"center_of_buoyancy":hydrostatics['center_of_buoyancy']})
```

mancellin:
> @mancellin right now the `*.nc` file from Capytaine stores `hydrostatic_stiffness`, but not `disp_volume`, `center_of_mass`, or `center_of_buoyancy`. WEC-Sim needs all of them. Do you want those those parameters added to the Capytaine output by default? If not, I'll update the BEMIO examples on the WEC-Sim repo to include something like this:

Yes, it could be automatically added to the dataset. With the caveat that it might be meaningless for multiple bodies, as discussed in #188

kmruehl:
Thanks @mancellin I'll respond on the [Capytaine issue](https://github.com/capytaine/capytaine/issues/188).

dav-og:
At the moment we use Meshmagick to write KH.dat and Hydrostatics.dat files.

@mancellin and I agreed that the best approach at the moment is to simply have Capytaine write these .dat files instead of Meshmagick.

This will remove the need to modify BEMIO and/or fundamental data structures within Capytaine. It should also be intuitive for useres and removes the risk of invalidated hydrostatic data being written to the .nc file (e.g. if a body is created, combined with other bodies to create a system, then translated).

In the future Capytaine may use a different structure to organize systems of bodies, in which case this approach can be revisited.

PR #931, opened 2022-08-08
Update coordinate system figure

JiaMiGit:
Updated coordinate system with the following features:
1. Add the mean water surface to the global system;
2. Add different view perspectives to help overall system understanding.


kmruehl:
@JiaMiGit thank you! Can you share the source files with us too? Thanks again.

PR #938, opened 2022-08-24
Ainf using radiationIRF for all BEM software

salhus:
This PR follows #880 .
If the added-mass infinity was missing from the BEM run, radiationIRF function calculates the Added-mass at infinity.
However, this only happens for non-WAMIT BEM codes.
This PR generates the added-mass at infinity, if it is missing from the BEM run regardless of the BEM code.


nathanmtom:
@salhus Thank you for re-submitting this PR after pulling in the latest versions of WEC-Sim into your branch. 

Unfortunately, the changes you have implemented for using the Fourier transform to generate the infinite frequency added mass only apply when `hydro.Ainf` is not present in the hydro structure. This works for the WAMIT example you have listed; however, under our current `readNEMOH` and `normalizeBEM` functions after running `readNEMOH` the `hydro` structure does have an `Ainf` variable defined so the updates made in the calculation in `radiationIRF` are not implemented. If you want to keep the `radiationIRF` unchanged then you may want to just remove line 61 in `normalizeBEM` so we are not pulling the added mass at the highest frequency simulated in the BEM code. 

Therefore, we cannot merge this PR until changes are made and checked across the other BEM import scripts.

source/functions/BEMIO/radiationIRF.m
```diff
@@ -68,7 +68,7 @@
 % Calculate the infinite frequency added mass
 ra_Ainf_temp = zeros(length(hydro.w),1);                                    %Initialize the variable
 [~,F] = size(hydro);                                                        %Last data set in
-if strcmp(hydro(F).code,'WAMIT')==0
+if isfield(hydro,'Ainf') == 0
```
nathanmtom, line 71: @salhus So this change will not update the infinite frequency added mass value for readNEMOH since this is populated in the normalizeBEM function. We either need to add a check here to see if Ainf is empty or the code is not WAMIT to implement the fourier transform to estimate the infinite frequency added mass. Or remove a line in normalizeBEM to not pre-populate the infinite frequency added mass.

salhus: @nathanmtom 
Sorry I didnt realize normalizeBEM was doing that. I think I like the approach where normalizeBEM is not prepopulating the infinite added mass. This would make future debugging easier.

nathanmtom:
@salhus Thanks for some updates to this PR. Unfortunately right now all of the tests appear to be failing so we will need to resolve that issue before we can finalize updating this PR.

salhus:
> @salhus Thanks for some updates to this PR. Unfortunately right now all of the tests appear to be failing so we will need to resolve that issue before we can finalize updating this PR.

@nathanmtom 
I was using an obsolete version of dev. I am closing this PR and putting in another PR #946 with just the required changes.

PR #940, opened 2022-08-30
readCapytaine_bugFixes_for_bemio_not_reading_multibody_runs_hydrostatics_and_normalization

salhus:
Minor bug fixes,
1. Modify readCapytaine.m to be able to extract body names in hydro using the input strings in capytaine in body#+body#... format,
2. Modify readCapytaine.m indices to correctly extract hydrostatics coefficients,
3. Modify normalizeBEM to have the normalization denominators to be in double format instead of int64. ensuring the denominator to be in double wouldn't affect other BEM reading code but will allow it to be used for normalization of  capytaine generated files.


source/functions/BEMIO/readCAPYTAINE.m
```diff
@@ -154,7 +154,7 @@
         if hydro(F).Nb == 1
             fileID = fopen(fullfile(meshdir,'Hydrostatics.dat'));
         else
-            fileID = fopen([fullfile(meshdir,'Hydrostatics_'),num2str(m-1),'.dat']);
+            fileID = fopen([fullfile(meshdir,'Hydrostatics_'),num2str(m),'.dat']);
```
dav-og, line 157: I think this should be left as m-1 to keep with existing convention of starting from KH_0.dat, Hydrostatics_0.dat, etc

source/functions/BEMIO/readCAPYTAINE.m
```diff
@@ -202,7 +202,7 @@
         if hydro(F).Nb == 1
             fileID = fopen(fullfile(meshdir,'KH.dat'));
         else
-            fileID = fopen([fullfile(meshdir,'KH_'),num2str(m-1),'.dat']);
+            fileID = fopen([fullfile(meshdir,'KH_'),num2str(m),'.dat']);
```
dav-og, line 205: I think this should be left as m-1 to keep with existing convention of starting from KH_0.dat, Hydrostatics_0.dat, etc

dav-og:
Left some comments inline:

https://github.com/WEC-Sim/WEC-Sim/pull/940/files#r959383951

https://github.com/WEC-Sim/WEC-Sim/pull/940/files#r959383714

salhus:
@dav-og 
I changed the m index counter back to m-1.

salhus:
I will be putting in this pull request again to avoid conflicts.

PR #962, opened 2022-11-02
included readCAPYTAINE() argument to explicitly define KH.dat & Hydro…

dav-og:
…statics.dat locations

### DIFF
--- a/source/functions/BEMIO/readCAPYTAINE.m
+++ b/source/functions/BEMIO/readCAPYTAINE.m
@@ -1,4 +1,4 @@
-function hydro = readCAPYTAINE(hydro,filename)
+function hydro = readCAPYTAINE(hydro, filename, hydrostatics_sub_dir)
 % Reads data from a Capytaine netcdf file
 % 
 % See ``WEC-Sim\examples\BEMIO\CAPYTAINE`` for examples of usage.
@@ -10,6 +10,9 @@ function hydro = readCAPYTAINE(hydro,filename)
 %     
 %     filename : string
 %         Capytaine .nc output file
+%
+%     hydrostatics_sub_dir : string
+%         Path to directory where Hydrostatics.dat and KH.dat files are saved 
 %
 % Returns
 % -------
@@ -27,7 +30,15 @@ end
 p = waitbar(0,'Reading Capytaine netcdf output file...'); %Progress bar
 
 hydro(F).code = 'CAPYTAINE';
-[meshdir,name,~] = fileparts(filename);
+
+[base_dir, name, ~] = fileparts(filename);
+
+if(~isempty(hydrostatics_sub_dir))
+    hydrostatics_dir = append(base_dir, hydrostatics_sub_dir);
+else
+    hydrostatics_dir = base_dir;
+end
+
 hydro(F).file = name;  % Base name
 
 % Load info (names, size, ...) of Capytaine variables, dimensions, ...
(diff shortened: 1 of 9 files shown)


source/functions/BEMIO/readCAPYTAINE.m
```diff
@@ -27,7 +30,15 @@
 p = waitbar(0,'Reading Capytaine netcdf output file...'); %Progress bar
 
 hydro(F).code = 'CAPYTAINE';
-[meshdir,name,~] = fileparts(filename);
+
+[base_dir, name, ~] = fileparts(filename);
```
akeeste, line 34: Update to camelCase

source/functions/BEMIO/readCAPYTAINE.m
```diff
@@ -27,7 +30,15 @@
...
+
+if(~isempty(hydrostatics_sub_dir))
+    hydrostatics_dir = append(base_dir, hydrostatics_sub_dir);
```
akeeste, line 37: Update to camelCase

akeeste (commented):
Hi @dav-og 

Can you describe the use case here? Typically these files would save to the same directory as the output structure. Also, does this change fit Capytaine's new internal hydrostatics? I haven't run Capytaine since hydrostatics was included, so I'm not sure if that output format has changed at all

source/functions/BEMIO/readCAPYTAINE.m
```diff
@@ -200,9 +211,9 @@
 for m = 1:hydro(F).Nb
     try
         if hydro(F).Nb == 1
-            fileID = fopen(fullfile(meshdir,'KH.dat'));
+            fileID = fopen(fullfile(hydrostatics_dir,'KH.dat'));
         else
-            fileID = fopen([fullfile(meshdir,'KH_'),num2str(m-1),'.dat']);
+            fileID = fopen([fullfile(hydrostatics_dir,'Hydrostatics_'),num2str(m-1),'.dat']);
```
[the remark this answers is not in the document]
dav-og: good catch! fixed

dav-og:
> Can you describe the use case here? Typically these files would save to the same directory as the output structure. Also, does this change fit Capytaine's new internal hydrostatics? I haven't run Capytaine since hydrostatics was included, so I'm not sure if that output format has changed at all

Use case is you can have your hydrostatics files saved in a separate location now (if you want to). This is handy for organizing data on a project I'm currently working on that has a _lot_ of floating bodies.

source/functions/BEMIO/readCAPYTAINE.m
dav-og: Can do but I think all of the other variables in this file are using underscore convention?

akeeste: Fair point, I forgot that we haven't yet updated the hydro struct variables to camelCase before since that would be a large change. In that case let's leave it as you have it. It can be addressed in a future refactoring if necessary.

source/functions/BEMIO/readCAPYTAINE.m
akeeste, line 37: Last thought, can you add to the logic here so that user's old BEMIO scripts can function without the ``hydrostatics_sub_dir`` argument? Maybe something like ``if exist('hydrostatics_sub_dir','var') & ~isempty(hydrostatics_sub_dir)``. I believe that should allow users to not input the directory, but the single '&' should not break the second condition if the variable doesn't exist.

akeeste:
@dav-og That makes sense, seems like a benefit to allow this for cases with many floating bodies. Last request, can you update the logic to make it backwards compatible with no sub directory input? This will only be an easy work around until other arguments are added to ``readCapytaine``, but might be worthwhile until then 

Adam

source/functions/BEMIO/readCAPYTAINE.m
dav-og: sure 👍

akeeste:
Thanks for the update! Approved and I will merge if there's nothing else.

salhus:
Thanks @akeeste for pointing out the backwards compatibility.
I reviewed this PR yesterday and all looks good.

For the future, I think we could introduce a 'config' argument, that will be a structure for all bemio settings.
Although, the bemio related functions have arguments for changing parameters like time steps etc. it would be much easier to allow user to define fields in a config structure.  That way the function call will not change if future parameters for bemio runs are introduced. Also, bemio calls for all BEM software will have the same arguments.

akeeste:
@salhus Agreed. I'd like to do this for the plotting functions too. I made a project card for Q2 to address this topic https://github.com/WEC-Sim/WEC-Sim/projects/58#card-86577803

PR #967, opened 2022-11-22
Controller Class and Library

jtgrasb:
I have added a controller class to WEC-Sim to allow users to easily add control methods to their WEC models. I wanted to start a PR to get the review started/any feedback. This PR includes:
- A controller library containing passive, reactive, latching, declutching, and MPC controllers
   - Passive and reactive act the same as PTO damping and stiffness except that they can be negative (which is often required for optimal stiffness)
   - Latching and declutching are only applicable in regular waves for the current implementation (latches/declutches for length of time during each period), but may be able to be extended to irregular waves in the future
   - MPC currently works well for 1 body with heave motion, but not yet applied to OSWEC (a bit complex because it requires BEM solved at hinge location for plant model, but cog for WEC-Sim) or 2 body 
- A controller class object which sets/takes inputs for the controller variables 
- Examples of each of the 5 controllers using the RM3 float body
   - These are in the examples folder for now to be reviewed but will be moved to WEC-Sim Applications before merging
- Documentation for the controller class and each of the examples (still need to finish MPC documentation)


docs/user/advanced_features.rst
```diff
@@ -868,6 +868,268 @@ Constraint and PTO Features
...
+By characterizing the impedance of the WEC, a greater understanding of the 
+dynamics can be reached. The figure below is a bode plot of the impedance of 
+the RM3 float body. The natural frequency is defined by the point at which the 
+phase of impedance is zero. By also plotting the frequency of the incoming 
+wave, it is simple to see the difference between the natural frequency of 
+the device and the wave frequency. Complex conjugate control (and many other
+control methods) seeks to adjust 
+the natural frequency of the device to match the wave frequency. 
+
+.. figure:: /_static/images/impedance.PNG
```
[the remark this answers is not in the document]
jtgrasb: @H0R5E 
Thanks! Should be fixed now

kmruehl:
Let's move this PR to the applications repo, but first we need to review the controller class and library.

dforbush2:
Hi @jtgrasb 
I'm reviewing this. I feel like we ought to reconsider pushing this to an applications case with the amount of work that has been done with the integration into the library. Certainly need documentation to show the use and make toolbox requirements clear, but unless the additional toolbox requirements can actually break the rest of WEC-Sim for someone running with the minimum WEC-Sim  The exception to this might be the MPC: as implemented, this seems potentially more appropriate as an example, but open to discussion on all points. 

TODO Test: Someone with the bare minimum WEC-Sim required toolboxes test this and ensure functionality on everything except where toolboxes are missing (e.g., MPC). 

Overall, this is excellent and is major addition to WEC-Sim functionality and I can see it serving a great role in future WEC-Sim courses especially if they are integrated with a broader Wave Energy course. 

Detailed remarks to follow. I avoided making changes to the library/functions unless I was sure they were intended.
1) Is there an implementation difference between (1/z) and Memory blocks? My understanding was the latter were the more robust option for all of the different solvers a user might pick. Specifically, 1/z used in feedback loop for declutching, latching and maybe others?
2) Check Simulink function call for "latchingTime" in the Simulink model. It is empty but probably should call latchingTime.m.
3) MPC workflow is distinct and doesn't appear to have a corresponding library block (looks like maybe this was unintended: the MPC control block link was broken. If possible to re-link, please do so. The MPC overall stands out as more of an example than a library: doesn't seem to generalize to other WEC geometries, particularly those with coupled DOF? Also pretty sure this will not work as intended in a variable time step. 
4) makePlantModel.m: could this leverage code already used in the state-space approximation? How does this method compare in terms of accuracy?
5) makePredictiveModel: This uses a c2d transform for discretization. I have a couple of concerns. 1) does this imply that MPC can only work for fixed step solvers? The ZOH method can cause some phase probably undesired phase behavior at high frequencies in its approximation of the continuous model. 
6) S-function: the MPC uses an s-function for fExcPrediction. This has caused version compatibility issues in the past. Also, this requires the stats toolbox additionally with the 'regress' function. 
7) See line 69 of fExcPrediction. Was this in-house code? Any licensing concerns?

examples/Controllers/Declutching/optimalTimeCalc.m
```diff
@@ -0,0 +1,73 @@
...
+omega = (1/T)*(2*pi);
+
+% Define excitation force based on wave conditions
+ampSpect = zeros(length(floatHydro.simulation_parameters.w),1);
+[~,closestIndOmega] = min(abs(omega-floatHydro.simulation_parameters.w));
+ampSpect(closestIndOmega) = A;
+FeRao = squeeze(floatHydro.hydro_coeffs.excitation.re(3,:))'*simu.rho*simu.gravity + squeeze(floatHydro.hydro_coeffs.excitation.im(3,:))'*simu.rho*simu.gravity*1j;
+Fexc = ampSpect.*FeRao;
+
+% Define the intrinsic mechanical impedance for the device
```
nathanmtom, line 29: This is just the heave mechanical impedance.

examples/Controllers/Declutching/optimalTimeCalc.m
```diff
@@ -0,0 +1,73 @@
...
+mass = simu.rho*floatHydro.properties.volume;
+addedMass = squeeze(floatHydro.hydro_coeffs.added_mass.all(3,3,:))*simu.rho;
```
nathanmtom, line 31: This seems hard coded to only be the vertical motion and for generalization would have expected it to be selectable; however, you have mentioned that we need coordinate transformations to get the the impedpence at a different location. The MPC WECCCOMP example does something similar by building the dynamic equation of motion about the rotational point. Because most of these examples are custom, I think putting these as examples in the applications case is better than inside of WEC-Sim Master.

examples/Controllers/Latching/optimalTimeCalc.m
```diff
@@ -0,0 +1,74 @@
...
+% Define excitation force based on wave conditions
+ampSpect = zeros(length(floatHydro.simulation_parameters.w),1);
+[~,closestIndOmega] = min(abs(omega-floatHydro.simulation_parameters.w));
+ampSpect(closestIndOmega) = A;
+FeRao = squeeze(floatHydro.hydro_coeffs.excitation.re(3,:))'*simu.rho*simu.gravity + squeeze(floatHydro.hydro_coeffs.excitation.im(3,:))'*simu.rho*simu.gravity*1j;
+Fexc = ampSpect.*FeRao;
+
+% Define the intrinsic mechanical impedance for the device
+mass = simu.rho*floatHydro.properties.volume;
+addedMass = squeeze(floatHydro.hydro_coeffs.added_mass.all(3,3,:))*simu.rho;
```
nathanmtom, line 31: Similar to my above comment, but if we are hard coding these then I think it needs to be an applications case.

examples/Controllers/MPC/mpcFcn.m
```diff
@@ -0,0 +1,214 @@
...
+    u_arrow_upperbound; ...
+    u_arrow_upperbound; ...
+    y_arrow_upperbound-Sx*x_k-Sv*v_arrow_k; ...
+    y_arrow_upperbound+Sx*x_k+Sv*v_arrow_k];
+
+
+%% Run quadprog
+
+if mpcStruct.MPCSetup.currentIteration > mpcStruct.modelPredictiveControl.Ho + mpcStruct.modelPredictiveControl.order % Only run mpcStruct after buffer is full, otherwise Fe predictions are not accurate
+    options = optimoptions('quadprog','Algorithm','interior-point-convex','Display','off', 'MaxIter', 600); % Can't set x0 w/ interior-point-convex
```
nathanmtom, line 182: Quadprog is part of the optimization toolbox and will not run with base MATLAB.

docs/user/advanced_features.rst
```diff
@@ -868,6 +868,316 @@ Constraint and PTO Features
...
+device which is equal to the complex conjugate of the impedance. Hence, it is 
+also known as impedance matching and is a common practice within electrical 
+engineering. Complex conjugate control is not a completely realizable control 
+method due to its acausality. which means it requires exact knowledge of 
+future wave conditions. Still, complex conjugate control presents a reference 
+for the implementation of optimal control. The WEC impedance can be modeled by 
+the following equation and can be used to formulate optimal control 
+implementation:
+
+.. math::
```
nathanmtom, line 923: This equation is appropriate for a single oscillating body, but is not generalizable to a multiple body system. As such, I am still in favor of an example case as this class is not going to be fully generalizable to more complex WECs that we see users researching through issues.

nathanmtom:
Thanks for proposing this PR @jtgrasb, clearly there is a lot of work gone into getting this controller toolbox ready be be integrated into the WEC-Sim. My initial thought is that this is still better suited as a suite of WEC-Sim Application Cases as you already have the 5 controller cases for the RM3 (only float), but none of the other examples. Also the limitation for using the latching/declutching only in regular waves is a bit misleading since there are several papers out there that tie the hold/release times that are tied to the natural period of the device and this is going to perhaps result in quite a few questions from the user community which is going to have us dive deeper into control theory. By itself this would not be an issue, but considering where we draw the line for technical support I think needs to be considered. 

Also, since the toolbox suite you have generated is not generalizable to say the OSWEC or a two body attenuator design because of needing to shift the hydro coefficients away from the center of gravity to a hinge I think prevents this from being used for a wider range of concepts that WEC-Sim can support. Given we have kept WECCCOMP separate and it has an advanced MPC controller available, I think this would make for an excellent applications case, but this is not ready to be pulled into WEC-Sim master. At least not at this time, I think research from both labs have shown that sysID using WEC-Sim is an option to generate a linearized system of equations that you can then use to build a controller. This would allow us to generalize the controller better to represent the wide diversity of WECs we see.

You are also using just the float of the RM3 model, when it was designed as a two-body system, and although it can still demonstrate the control approaches you have coded, in my opinion it will not be recognized by our users about the additional complexities such as including a second body or changing from say translational to rotational motion.

Again I think you have done a lot of great work to get the PR to be mergeable with WEC-Sim Master, but I still think it is still better suited as an application case.

More focused questions/comments are:

- controllerBlocks.png needs to be updated as two of the 5 controllers have "Force1" rather than "Force" as label.  Also, if MPC is removed then that is not an option we will be making available but there still seems to be a lot of MPC functions available although a previous commit mentions removing MPC.
- I might be doing something wrong in the compiling, but when I build docs locally and look for your updated section nothing is present. Potentially double check that with the latest push you did not do something to effect the docs build. There also seems to be continued mention of MPC but it sounds like you removed this.
- Similar, when I build the docs locally I am not seeing any of the new sections.
- Have you compared the MPC approach in our WECCCOMP applications case against what you are implementing? We may need to highlight the differences so user do not assume that they are implementing the same approach.

docs/user/advanced_features.rst
```diff
@@ -868,6 +868,316 @@ Constraint and PTO Features
...
+    Z_i(\omega) = j\omega (I + I_{A}(\omega)) + R(\omega) + \frac{C_{hs}}{j\omega}
+
+By characterizing the impedance of the WEC, a greater understanding of the 
+dynamics can be reached. The figure below is a bode plot of the impedance of 
+the RM3 float body. The natural frequency is defined by the point at which the 
+phase of impedance is zero. By also plotting the frequency of the incoming 
+wave, it is simple to see the difference between the natural frequency of 
+the device and the wave frequency. Complex conjugate control (and many other
+control methods) seeks to adjust 
+the natural frequency of the device to match the wave frequency.
```
jleonqu, line 934: I would add a couple of comments about the challenges of adjusting the natural frequency of the device to match the wave frequency. In general, a resonant condition is not desire in any application, but this is not the case here, so it would be good to comment on this.

docs/user/code_structure.rst
```diff
@@ -746,13 +746,55 @@ restrictions on the number of MooringMatrix blocks.
...
+.. _user-code-structure-controller-class:
+
+Controller Class
+^^^^^^^^^^^^^^^^
+
+The controller class contains all the information for the controller blocks, which can be used to simulate WEC control methods 
+in WEC-Sim. At a high level, the controller class interacts with the rest of 
+WEC-Sim as shown in the diagram below:
+
+.. figure:: /_static/images/code_structure/PTOSimClass_diagram.png
```
jleonqu, line 758: Change figure to controller class diagram

jtgrasb:
@dforbush2 Thanks for the comments!
Agreed, see [Kelley’s comment](https://github.com/WEC-Sim/WEC-Sim/pull/967#issuecomment-1341154287), the plan is to have the examples reviewed here and then move them to WEC-Sim applications. One component to address is the controllerClass and library and whether those should be a part of WEC-Sim master or if all of it should be integrated into the examples for an applications case.
1. I’ll look into this. I didn’t use memory blocks simply because I’m not as familiar with them, but I’m guessing they can accomplish the same task, so I’ll update if possible.
2. Should be fixed, thanks!
3. This is intended - MPC functions require extra toolboxes so this was left out of the library so it could exclusively be an application.  
4. Will look into
5. Will look into
6. I hadn’t realized the compatibility issues with the S-functions. Are there any past discussions that I could look at that reference this issue? As for the stats toolbox, this shouldn’t be an issue as it will be part of an applications case
7. This was code from the control-sim repo, which I believe was public at one point, so I think it should be okay? Will remove these comments from the code though

examples/Controllers/Declutching/optimalTimeCalc.m
jtgrasb: Yes, this code is currently just addressing heave motion for simplicity

examples/Controllers/Declutching/optimalTimeCalc.m
jtgrasb: Yes, currently it is hard-coded for just heave, but I could work on it being selectable DOF. See [Kelley’s comment](https://github.com/WEC-Sim/WEC-Sim/pull/967#issuecomment-1341154287) - the plan is to move the examples to the applications case after review. The reason it was put in WEC-Sim master to be reviewed is to make it function with the controllerClass and library - do you think these are useful in master or should they also be removed and integrated as part of the applications case?

examples/Controllers/Latching/optimalTimeCalc.m
jtgrasb: Good point, I can work on switching the float out with IEA OES Task 10 sphere to be more clear.

examples/Controllers/MPC/mpcFcn.m
jtgrasb: With this being moved to the applications case, the extra toolbox shouldn't be an issue from what I understand.

jleonqu:
Great work on this PR @jtgrasb. The examples are clear and the documentation is well written. I made a couple of comments on the documentation one of them regarding the challenges of adjusting the natural frequency of the device to match the wave frequency. It would be useful to add a couple of comments on this.
I agree with Nathan, the examples are great but at this point I think is difficult to apply the library to a general case. As Nathan said, the examples that you have so far will be better suited for application cases, which will be really useful for WEC-Sim users.
Regarding multibody systems, I would suggest to have a different approach. Maybe we can formulate the equations with the control coefficients for a given controller, and then we can run an optimization routine to find the optimal coefficients for the multibody WEC. This is just an idea, but I think it could be useful for an application case.
Also, I was checking the simulink blocks, and all of then have the same description regardless of the type of block. All the blocks have the following description: Use WEC-Sim input file to define Latching Controller parameters. You should update all the blocks to have the appropriate description depending on the type of controller.

jtgrasb:
Thanks for all the comments and suggestions. I am working on incorporating your comments into a PR for the wec-sim applications case. I had some questions that I wanted to address first.

@dforbush2 
For the state space approximation used by WEC-Sim, my understanding is that it is just used for the radiation calculation while the state space model for MPC needs to take into account excitation and radiation forces. 
The c2d transformation here does restrict the MPC to a fixed step, but I think potentially the rest of the simulation could work with a variable step solver, but I’m not sure if it would work with the rate transition blocks. I think this is okay since it will be an applications case. As for the method (zoh, etc), I’m not too familiar with the different methods, is there a different one you’d suggest?
I’ve found a way to do it without S-functions, so working on transitioning to normal MATLAB functions now.

@nathanmtom  
You mentioned sysID has been used with WEC-Sim before to create a linearized system of equations to build a controller - is there a paper you could direct me to on this topic? 
For the comparison to the WECCCOMP example, I think the difference lies in the state space representation of the WEC motion and constraints (and obviously geometry), but I am still looking to pinpoint the exact differences in formulation.

@jleonqu 
I’d like to talk more about the multibody approach you mentioned as well as trying to integrate these examples with PTO-Sim better to emphasize the importance of considering electrical power rather than mechanical power as it is now.

Best,
Jeff

Issue #952, opened 2022-10-04
[Theory or Implementation] Moments of inertia and Geometry orientation on import


Dadidal:
**Is your question request related to a problem? Please describe.**
N/A

**Describe the theory or implementation approach that you have a question**
I am want to understand how to correctly prepare the geometry and set Inertia values.

**Personal project context**
I try to model floating ship that consists of different parts. I want to use hydrodynamic body for floating part and non-hydrodynamic bodies for non-submerged parts. I created the geometry (.stl mesh files) for each of the ship part, but I am not clear how to set the Inertia in WEC-Sim input file.
The problem is that WEC-Sim requires input of 3x1 inertia matrix. Therefore these inertia terms are principal moments, right? Can WEC-Sim automatically calculate the direction of principal axes? This is important because geometry of ship parts in .stl file is oriented directly to the final position that should be on the ship. In such a case principal axes are not aligned with global axes during import (see Figure), but terms of Inertia in WEc-Sim called Ixx Iyy and Izz (not Ip1, Ip2, Ip3).

Should I rotate the parts in CAD so that the principal axes correspond to the global axes and only after that import the .stl file and manually change the orientation of the bodies already in WEC-Sim?

**Describe the type of conclusion or resolution you are looking for**
I would appreciate if you will explain how to prepare geometry or how to set inertia 3x1 matrix for the case if principal axes of the body in .stl file are not aligned along the global axes.

**Additional context**
![image](https://user-images.githubusercontent.com/32517441/193935053-53c9ba6d-67a7-44c5-b7fb-c48c1153e0f9.png)
![image](https://user-images.githubusercontent.com/32517441/193935707-a7cc0628-087c-4975-ade7-a48a4ffd0801.png)

akeeste:
Hi @Dadidal 

Thanks for reaching out. You are correct that this can be an issue and I will try to clarify.

WEC-Sim will not automatically calculate a body's principal axes. If there is no initial rotation, it is assumed that the coordinate system in ``*.stl`` file is the global coordinate system (horizontal x and y, vertical z). The ``body.inertia`` from the ``wecSimInputFile.m`` must be defined according to the coordinate system in the ``*.stl`` file. Ixx is about the file's x-axis, Iyy about the file's y-axis, etc. There are two ways you can do this:

**Option 1**
[...]

PR #981, opened 2023-01-12
Products of Inertia in WEC-Sim

akeeste:
In their equilibrium position, the principal inertial axes of a body are not necessarily aligned with WEC-Sim's global coordinate system. Allowing products of inertia to be defined would make the inertia calculation much easier for users with highly irregular, asymmetric or angled bodies. Adding this feature involves:
- [x] adding a body class property `inertiaProducts` and linking it to Simscape in the body library
- [x] adding `inertiaProducts` to the added mass adjustments
- [x] documenting the proper convention for products of inertia. 
 
I chose to match Simscape's convention for the product of inertia calculation. Some CAD, like Solidworks, use a negated convention.

This addresses project card https://github.com/WEC-Sim/WEC-Sim/projects/58#card-86097325 and issue #952


salhus:
@akeeste Sorry I didn't merge this sooner. I checked the code, and I think this can be merged.

Issue #993, opened 2023-02-14
[BUG] Sign mismatch in Rotary To Linear Adjustable Rod


jwills801:
**Note: italicized text below is include as an example and should be updated before submission. If you feel any section is not applicable to your request, please replace with 'N/A' rather than delete a section.**

**Describe the bug**
I am using the check valve power take off for the RM5 as is done in the applications directory of wec-sim and is documented in the WEC-SIM PTO-Sim documentation page. I wanted to investigate the sign convention used (i.e. does positive velocity mean the the cylinder is moving toward side A or B?), so I added a To Workspace block to look at the position and velocity of the hydraulic cylinder in matlab. They did not seem to match so I integrated the velocity signal and compared it to the position signal. They were mirrors of each other. Changing the sign of one of them resulted in a near perfect match between position and integral of velocity.

**WEC-Sim file(s)**
The Rotary to Linear Adjustable Rod library block in PTO-Sim

**To Reproduce**
1. Go to WEC-Sim/WEC-Sim_Appications/OSWEC/OSWEC_Hydraulic_PTO
2. Adjust the .slx file so that translational position and velocity of the hydraulic cylinder can be observed.
3. run wecSim from command line
4. Plot translational position and velocity

**Expected behavior**
The integral of velocity should be position

**Screenshots**
![image](https://user-images.githubusercontent.com/62731220/218875627-6c735125-69f1-43c2-b57a-17dde77927fb.png)
![image](https://user-images.githubusercontent.com/62731220/218875662-a810a3bb-845b-4005-8958-120573342622.png)

**Workstation Information (this section MUST BE completed):**
 - OS: macOS Monterey 12.6.1
 - MATLAB/Simulink Release: 2021a
 - WEC-Sim Release: 5.0.1

**Additional context**
N/A

jleonqu:
Hello @jwills801 
Can you upload your working files? I would like to see where and how you are measuring the speed and the position of the cylinder. Thanks!

jwills801:
Hello @jleonqu 

Yes. Here is the repo: [https://github.com/jwills801/OSWEC_Hydraulic_PTO](https://github.com/jwills801/OSWEC_Hydraulic_PTO)

Thank you for the help!

jleonqu:
Hello @jwills801 

[...]

PR #999, opened 2023-02-23
Bug fix for direction in readCapytaine

akeeste:
This PR solves #997 and updates:
- [x] ``readCapytaine`` to convert direction from radians (Capytaine's convention) to degrees (WEC-Sim's convention)
- [x] updates ``examples/BEMIO/Capytaine`` scripts to use radians, not degrees
- [x] updates ``examples/BEMIO/Capytaine`` results

PR #1002, opened 2023-02-27
Fix sign bug reported on issue #993

jleonqu:
This is a PR to solve the sign bug reported on issue #993 
I have checked the equations for linear position and linear velocity. The linear velocity equation is correct, the problem was in the definition of the linear position.


akeeste (commented):
Hi @jleonqu 

I'm noting for future reference that this PR wants to change the sign of the ``Sum`` block immediately preceding the ``linearPos`` outport in ``PTOSim_Lib/Motion Conversion/Rotary to Linear Adjustable Rod/Angular to Linear Position``.
Previous set-up:
* linearPos = ptoSim.adjustableRod.rodInit - rod Position

Changed to:

* linearPos = rod Position - ptoSim.adjustableRod.rodInit

The change is simple enough and makes sense to me, but I can't find a good reference on what this should be. Can you update this thread or the documentation to have the equation for the adjustable rod? Is there a way we can add a test to this part of PTO-Sim to prevent future issues?

jleonqu:
The equation for the length of the adjustable rod is based on this figure: 
image

The rod vector is:
$\vec{d}_{rod} = (rsin\theta-d_x)\hat{i}+(rcos\theta-d_{off})\hat{k}$

Then, the rod final length is:
$d_{rod} = \sqrt{(rsin\theta-d_x)^2 + (rcos\theta-d_{off})^2}$

So, the rod linear position is the final rod length minus the initial rod length:

$d_{rod,LinPos} = d_{rod} - d_{rod,ini}$

akeeste:
@jleonqu Thanks for adding that reference. I will merge this PR, then pull #1002 and #999 into dev

PR #1012, opened 2023-03-16
Update to MoorDyn to fix MoorDyn crashing MATLAB session.

AlixHaider:
When MoorDyn is used in WEC-Sim design, it is called as an external DLL executable library, and it runs on a separate console.exe under MATLAB's main process, with its separate process-ID(PID) in the windows task manager. The issue is related to the following scenarios on the Windows platform:
1>> if the lines data text file has an error in it.
2>> MoorDyn fails to converge.
3>> MoorDyn is launched, but the model has some other error.
Under such a scenario, MATLAB fails to terminate the MoorDyn Console and hangs instead, and either crash immediately or crashes if we press the close button on the MoorDyn console window. Either way, there seems to be no way the MATLAB session can recover from MoorDyn failure without crashing. It becomes a major issue if we are trying to tune mooring lines and want to try different lines configurations and stiffnesses, etc., to find an optimal mooring system because some configurations will not work, which will keep crashing MATLAB sessions during the test and trial. 

The goal is for MoorDyn's failure not to crash the MATLAB session but instead MoorDyn dll execution to close and the MoorDyn console to close. I've found a solution as described in the next section. For scenarios 1 and 2, the following Fix#01 seems to work on the Windows platform. And to cover scenario 3, an additional Fix#02 is needed.
**Fix#01**
In the MoorDyn block's 'InitFcn' call, making the following suggested changes check for MoorDyn Initialization failure and terminates the MoorDyn.dll console without crashing the MATLAB Session.
image

```
% Initialize MoorDyn Lib (Windows:dll or OSX:dylib)
disp('---------------Starting MoorDyn-------------')
if ismac
    loadlibrary('Lines.dylib','MoorDyn.h');
elseif ispc
%-----------------
if libisloaded('Lines')
    calllib('Lines','LinesClose');
    unloadlibrary Lines;
end

loadlibrary('Lines.dll','MoorDyn.h');

%---------------
elseif isunix
  loadlibrary('Lines.so','MoorDyn.h');
else
  disp('Cannot run MoorDyn in this platform');
end


if calllib('Lines','LinesInit', [0 0 0] ,[0 0 0 0 0 0])<0  % if error in moorDyn
if libisloaded('Lines')
    calllib('Lines','LinesClose');
    unloadlibrary Lines;
end
    disp('-------------Error: Mooring Line Initialization failed in MoorDyn Block--------')
    disp('------------MoorDyn is Terminated----------')
end
```

**Fix#02**
Suppose the MoorDyn console has been launched, but there is an error in the Simulink model. In that case, MATLAB throws an error and terminates the Simulink execution without terminating the MoorDyn console window. A try-catch statement in the wecsim file can address this issue. If a Simulink model error occurs, we can print the error and terminate the MoorDyn Console process.
image
```
run('wecSimInputFile');
clear simu waves body cable pto constraint ptoSim mooring 

runWecSimCML = 1;
run('initializeWecSim');

try
catch e %e is an MException struct
    fprintf("------Error in the Model, Error-Causes/Messsages are :\n");
    for i=1:size(e.cause)
        fprintf(e.cause{i}.message);
        fprintf('\n');
    end
            % terminate MoorDyn Conhost.exe instances.
if libisloaded('Lines')
    calllib('Lines','LinesClose');
    unloadlibrary Lines;
end
 end
```


dav-og:
Thanks Ali, the PR is working with the stock MoorDyn demo from the WEC-Sim Applications repo.

Working through the 3 scenarios you described:

1. If the lines data text file has an error in it.
    - ✔ I made a deliberate error in the lines.txt file (simply deleted 'Vessel' from L9) and see this output

> ---------------Starting MoorDyn-------------
> -------------Error: Mooring Line Initialization failed in MoorDyn Block--------
> ------------MoorDyn is Terminated----------
> Elapsed time is 6.445772 seconds.
> 
> Post-processing and saving...   
> Elapsed time is 3.282007 seconds.
> ------Error in the Model, Error-Causes/Messsages are :

2. MoorDyn fails to converge.
    - ✔ I changed the MoorDyn timestep from 0.0005s to 0.5s and see this output:

> ---------------Starting MoorDyn-------------
> -------------Error: Mooring Line Initialization failed in MoorDyn Block--------
> ------------MoorDyn is Terminated----------
> Elapsed time is 6.518951 seconds.
> 
> Post-processing and saving...   
> 
>  No moorDyn *.out file saved for Line1
> 
>  No moorDyn *.out file saved for Line2
> 
>  No moorDyn *.out file saved for Line3
> 
>  No moorDyn *.out file saved for Line4
> 
>  No moorDyn *.out file saved for Line5
> 
>  No moorDyn *.out file saved for Line6
> Elapsed time is 0.731101 seconds.
> ------Error in the Model, Error-Causes/Messsages are :


3. MoorDyn is launched, but the model has some other error.
    - ✔ I changed the WEC-Sim timestep from 0.01s to 2.0s and see this output:

> ---------------Starting MoorDyn-------------
> Elapsed time is 10.562687 seconds.
> 
> Post-processing and saving...   
> Elapsed time is 0.788705 seconds.
> ------Error in the Model, Error-Causes/Messsages are :
> ['RM3MoorDyn/Floating (3DOF)/CONSTRAINT (Planar Joint)']: 'RM3MoorDyn/Floating (3DOF)/CONSTRAINT (Planar Joint)' has a degenerate mass distribution on its follower side.

Everything seems great to me! Thanks a lot for making this contribution! 😀

@AlixHaider can you just confirm that these error messages are as you'd expect, then I think its good to merge 👍

Cheers,
Dave

AlixHaider:
Hi Dave,
Thanks for your review. The error messages seem good to me. It's good to go!
Cheers,
-Alix

treppe8k:
I am using Matlab 2020b, when i try install WecSim last version, I got this error

####################

.>> slLibraryBrowser
Error using load_system (line 21)
Model 'WECSim_Lib_Moorings' was created with a newer version (R2022a) of Simulink
To disable this error message, use Simulink preferences.
To create a model that is compatible with this version of Simulink, load the model in Simulink
R2022a and select Save > Export Model to > Previous Version.

####################

You can try re-export the 'WECSim_Lib_Moorings.slx' with the the MatLab 2020b version ?
According to guide , that is the version in use for WecSim 5.0.1

thanks in advance

PR #1022, opened 2023-04-05
Remove 'fixed' mass option from OSWEC input file

jtgrasb:
This PR addresses a bug where the 'fixed' mass option was still being used for the OSWEC case, causing it to give errors when trying to run. This is only an issue in the dev branch. 

The 'fixed' mass option was removed with PR #856, but the input file to the OSWEC was changed back with [this commit](https://github.com/WEC-Sim/WEC-Sim/commit/51ef9e9322b7c08ea808263443d32092fdaa97db). This PR simply removes the 'fixed' mass and sets the mass and inertia equal to 999.


nathanmtom (approved):
@jtgrasb Thanks for catching this bug. I've confirmed the issue using the 'Dev' branch and approve this pull request to solve the issue.

PR #1023, opened 2023-04-05
Save the applied added mass time series

akeeste:
This PR adds a call in ``postProcess.m`` to the pre-existing ``bodyClass.storeForceAddedMass()`` function.  This allows WEC-Sim to save the timeseries of the added mass force that is applied during the simulation. 

This is necessary for a TEAMER award where the added mass force is varied during simulation. I'd like to change a couple other items in the added mass post-processing, but this is the least intrusive and most important for visualizing the current TEAMER work.

### DIFF
--- a/source/functions/postProcessWecSim.m
+++ b/source/functions/postProcessWecSim.m
@@ -130,6 +130,7 @@ end; clear iMoor
 % Calculate correct added mass and total forces
 for iBod = 1:simu.numHydroBodies
     body(iBod).restoreMassMatrix
+    body(iBod).storeForceAddedMass(output.bodies(iBod).forceAddedMass, output.bodies(iBod).forceTotal);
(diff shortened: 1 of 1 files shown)


nathanmtom (approved):
@akeeste the first change file is related to updates made by a recent PR so no concerns. The modification to the postProcessWECSim is only adding an additional call during execution and this should not influence dynamic performance. 

My only comment is that we should add some documentation in the developer section to illustrate how this postProcessing is being used to correct the addedmass force time history. Currently the advanced features developer section only describes the addedMass adjustment process. This does not need to be implemented in this PR but flagging as this book keeping is important to document to reduce future confusion.

nathanmtom:
@akeeste Thanks for this PR, I approve and will let you merge when ready.

akeeste:
@nathanmtom Thanks for your quick review. I removed the extraneous file and added a [project card for myself](https://github.com/WEC-Sim/WEC-Sim/projects/58#card-88786783) to expand the developer docs on our post processing corrections.

Issue #1037, opened 2023-04-21
[BUG]  WECSim_Lib_Moorings incompatible matlab 2020b


treppe8k:
I am using Matlab 2020b, when i try install WecSim last version, I got this error

####################

.>> slLibraryBrowser
Error using load_system (line 21)
Model 'WECSim_Lib_Moorings' was created with a newer version (R2022a) of Simulink
To disable this error message, use Simulink preferences.
To create a model that is compatible with this version of Simulink, load the model in Simulink
R2022a and select Save > Export Model to > Previous Version.

####################

You can try re-export the 'WECSim_Lib_Moorings.slx' with the the MatLab 2020b version ?
According to guide , that is the version in use for WecSim 5.0.1

thanks in advance

jtgrasb:
@treppe8k Thank you your inquiry into using WEC-Sim's Mooring Library and notifying us of the issue. We are working on addressing this issue now. It seems to be related to PR #1012, but I am getting some issues with MATLAB crashing when trying to update it. I've created issue #1038 to work through the bug. 

If you have any other issues, please note we are requiring that all inquiries follow the issues template and please answer every section as requested without deleting the section headers so we know which questions on the template are being addressed by your statements. We prefer to stick with the templates as it helps us track issues historically and new users can better parse the knowns and unknowns of a posted issue.

![image](https://user-images.githubusercontent.com/87095491/233644575-e1ebee36-5e4d-479c-9018-6b1f2766d284.png)

treppe8k:
**the issue persist but with other paths** 

> the 2 examples (verified) and tutorials (maybe)

Warning: Model 'OSWEC' was created with a newer version (R2021a) of Simulink
To create a model that is compatible with this version of Simulink, load the model in Simulink
R2021a and select Save > Export Model to > Previous Version. 
> In load_system (line 21)
In simulationClass/loadSimMechModel (line 194)
In initializeWecSim (line 463)
In run (line 91)
In wecSim (line 43) 
Warning: Cannot load version 'R2021a_202009111111' of family 'sltp_mm'.

jtgrasb:
Hi @treppe8k,

We only require our WEC-Sim Library files to be saved in R2020b per discussion in PR #919. The newer models should run fine in 2020b. This error message can be disabled in Simulink's preferences or using the following command (which is also called when you run wecSim.m):

set_param(0,'ErrorIfLoadNewModel','off')

Issue #1038, opened 2023-04-21
[BUG] Mooring Library in wrong MATLAB Version and crashing MATLAB


jtgrasb:
**Note: italicized text below is include as an example and should be updated before submission. If you feel any section is not applicable to your request, please replace with 'N/A' rather than delete a section.**

**Describe the bug**
The mooring library was updated in PR #1012 to fix some moordyn crashing issues, but it was saved in R2022b, when we specify R2020b in our documentation (Issue #1037). I tried to revert the mooring library to R2020b and confirm that the PR's changes are still present, but my MATLAB crashes every time I try to look under the Mooring or MoorDyn block masks. 

**WEC-Sim file(s)**
_If known, please identify the the WEC-Sim files causing the bug or error:_
_Please upload relevant case files if possible to speed debugging._

**To Reproduce**
_Steps to reproduce the behavior:_
1. Open WECSim_Lib_Moorings.slx
2. Click on the Mooring Matrix or MoorDyn block

**Expected behavior**

**Screenshots**
![image](https://user-images.githubusercontent.com/87095491/233647818-bdf5cc0f-fe5d-4b11-b986-59e747315e27.png)

**Workstation Information (this section MUST BE completed):**
 - OS: Windows 10
 - [MATLAB/Simulink Release](https://www.mathworks.com/help/matlab/release-notes.html) [2021b]
 - [WEC-Sim Release](http://wec-sim.github.io/WEC-Sim/man/release_notes.html) [v5.0.1]

**Additional context**

akeeste:
I also tried saving the library version back to R2020b. I used R2022b to do this and do not have any crashing problems afterwards. #1040 should resolve this issue

PR #919, opened 2022-07-25
Add tests to check that SLX file versions do not exceed R2020b

H0R5E:
This PR adds some new tests to check whether the versions of any SLX files stored in the repository exceed R2020b.

Having these tests will make sure that incompatible versions are not merged accidentally.

PR #1040, opened 2023-04-21
Save mooring library to R2020b

akeeste:
This PR resolves issues #1037 #1038

I will pull this change from master into dev after merging.


jtgrasb:
Thanks Adam, I am confirming that this successfully resolves Issues #1037 and #1038. I will merge it once the tests pass.

Per our discussion, I looked at the existing test suite and there does not seem to be any tests to check the version of MATLAB used in our library files.

jtgrasb:
Hi @akeeste,
I was looking back at old PRs to find our version saving practice and found where the version tests are added (PR #919). These changes are still only in dev while the mooring library issue was in master, so that is why it was not detected by any tests.

Issue #1020, opened 2023-03-30
[Theory or Implementation] How to create h5 file with mean drift forces?


Dadidal:
**Is your question request related to a problem? Please describe.**
I do not understand how to create a hydrodynamic data file (*.h5) with mean drift forces.

**Describe the theory or implementation approach that you have a question**
Firstly I calculated BEM data in ANSYS Aqwa. This BEM data includes mean drift forces. I know that according to documentation, WEC-Sim can read the mean drift forces only from WAMIT. Therefore I used BEMRosetta to change the BEM output format: I imported results from AQWA and then exported them to WAMIT format. Then I used BEMIO to read "WAMIT" *.out file and tried to simulate my model with "body(1).meanDrift = 1" option but obtained an error:
```
Error using h5readc
The HDF5 library encountered an error and produced the following stack trace information:

    H5G__traverse_real    component not found

Error in h5read (line 93)
[data,var_class] = h5readc(Filename,Dataset,start,count,stride);

Error in readBEMIOH5 (line 98)
    hydroData.hydro_coeffs.mean_drift =  reverseDimensionOrder(h5read(filename, [h5BodyName '/hydro_coeffs/mean_drift/control_surface/val']));
```
It seemed that my h5 file does not contain mean drift forces even though they exist in BEM data. I checked h5 file and observed that all matrixes of mean drift forces have 0 values.
Could you give some advice on how to fix it?
I have attached the following files (BEM data.zip):
1. BEM output from ANSYS Aqwa
2. Converted BEM data to WAMIT format
3. Generated .h5 file

**Personal project context**
I am trying to include 2nd order wave forces in my simulation

**Describe the type of conclusion or resolution you are looking for**
I want to clarify and understand what I should do in order to include mean drift forces in WEC-Sim simulation

**Additional context**
![image](https://user-images.githubusercontent.com/32517441/228757296-30d9e78f-c139-4da5-b62c-6049ca1f163b.png)
[BEM data.zip](https://github.com/WEC-Sim/WEC-Sim/files/11108894/BEM.data.zip)

nathanmtom:
@Dadidal Thank you for your issue and I'll do my best to provide options to resolve the problem.

- First off, I want to let you know that we do have an applications case with an example of the .out file format from WAMIT with mean drift. https://github.com/WEC-Sim/WEC-Sim_Applications/blob/master/Mean_Drift/WAMIT/outputFiles/sphere.out
[...]

PR #1046, opened 2023-04-27
Update read- and writeBEMIOH5 to allow for pressure integration for mean drift 

nathanmtom:
This is an update to the read- and writeBEMIOH5 to allow for pressure integration for the mean drift force import. 

This update should address issue #1020 as the import of the pressure integration is successful when reading the input file, but is not written to the .h5 file (writeBEMIOH5) and when reading the h5 file there is no option for specifying the pressure integration option.

### DIFF
--- a/source/functions/BEMIO/readBEMIOH5.m
+++ b/source/functions/BEMIO/readBEMIOH5.m
@@ -98,6 +98,8 @@ elseif meanDrift == 1
     hydroData.hydro_coeffs.mean_drift =  reverseDimensionOrder(h5read(filename, [h5BodyName '/hydro_coeffs/mean_drift/control_surface/val']));
 elseif meanDrift == 2
     hydroData.hydro_coeffs.mean_drift =  reverseDimensionOrder(h5read(filename, [h5BodyName '/hydro_coeffs/mean_drift/momentum_conservation/val']));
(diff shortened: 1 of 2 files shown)


jtgrasb:
Hi @nathanmtom I tested this PR with the mean drift applications case and got an error because the bodyClass has a check to make sure meanDrift is between 0:2. This needs to be changed to 0:3. Other than that, this PR seems to work well!

nathanmtom:
Thanks for the quick review @jtgrasb, did not catch that as I didn't actually try to run WEC-Sim so missed the bodyClass checkInput warning. I believe this only occurs here and couldn't break the code by changing other parameters, but let me know if this check needs to be implemented in another place. 

Checking the meanDrift application case also shows that I'll need to update that case as it does not seem to reflect the v5.0 refactoring and will need to update to demonstrate the 'pressure integration' option. I'll put this as a project board tag and get to it in the near future.

jtgrasb:
Looks great @nathanmtom. Merging now!

As for the applications case, I noticed this as well yesterday and made a [PR](https://github.com/WEC-Sim/WEC-Sim_Applications/pull/31) updating the userDefinedFunctions. Let me know if there was anything else you saw that should be updated.

PR #1048, opened 2023-05-02
Add function to read h5 file to hydro data structure

jtgrasb:
This PR adds a function to BEMIO to allow the h5 file to be read in as a hydro structure. This allows the hydro data in an h5 file to be plotted even without access to the BEM files that it was read from. A couple lines were added to writeBEMIO to make sure all the hydro data was written to the h5 file, and this lead to small additions to rm3.h5 and oswec.h5. This also adds an example to compareBEMIO.m to show how the h5file can be read into a hydro data structure.


nathanmtom (approved):
@jtgrasb Thanks for building this function to allow a user to quickly check their hydrodynamic coefficients without re-running the full BEMIO script. I approve this PR. 

One note, is that when looking at the plots for the comparison you added in compareBEMIO.m, it is hard to tell there are four lines because the line style is solid. I am not requesting this as part of this PR, but potentially a nice change would be to update the linestyles for each new line in the formatPlot.m.

nathanmtom:
Thanks for creating this function and updating the writeBEMIOh5.m to reflect other values that should be stored in the .h5. I'll merge this shortly.

PR #1049, opened 2023-05-04
Normalize quaternion to increase simulation robustness

akeeste:
An issue came up in a TEAMER award that uses two cables with spherical joints on all ends. The device motion causes the spherical joints to continuous rotate, causing an increasingly large angular position. This should be allowed to happen with a spherical joint, but our current ``quaternion2EulXYZ`` function will break if the quaternion value becomes too large. 

This PR normalizes the quaternion of any angular rotation in Simulink so that it can be robustly converted to Euler angles.


akeeste:
@dav-og I believe you've worked with the quaternion convention before. Will this normalization have any unexpected consequences?

dav-og:
> @dav-og I believe you've worked with the quaternion convention before. Will this normalization have any unexpected consequences?

Not to my knowledge 👍

PR #1061, opened 2023-05-22
Add windows runner to dev branch unit test workflow

H0R5E:
This PR changes the following:

+ Added a Windows runner to the dev branch unit tests workflow by using the beta version of setup-matlab
+ Added untested versions of MATLAB (R2022a, R2022b)
+ Allows the workflow to be triggered manually

### DIFF
--- a/.github/workflows/run-tests-dev.yml
+++ b/.github/workflows/run-tests-dev.yml
@@ -2,6 +2,7 @@ name: Run MATLAB tests on dev branch
 on:
   # Trigger the workflow on push or pull request,
   # but only for the dev branch
+  workflow_dispatch: {}
   push:
     branches:
       - dev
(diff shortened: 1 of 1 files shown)


H0R5E:
I can create a clone PR for the master branch tests, but I wondered if we'd like to try this out in dev for a while first before doing that?

kmruehl:
@H0R5E this is great, thank you! I agree with your assessment to keep this on dev for now. We have a milestone for a new release by the end of the FY, so it'll be on master soon enough. 

Also, thanks for the FYI on the runtime. We may want to revisit how many versions of MATLAB we want to support (and test). Thanks again!

H0R5E:
> We have a milestone for a new release by the end of the FY, so it'll be on master soon enough.

@kmruehl, just to clarify this point, the actual master workflow (`.github/workflows/run-tests-master.yml`) still needs to be updated to use the windows runners.

Issue #1028, opened 2023-04-11
[BUG] Hydrostatic restoring matrix


hachikoi1:
**Note: italicized text below is include as an example and should be updated before submission. If you feel any section is not applicable to your request, please replace with 'N/A' rather than delete a section.**

**Describe the bug**
1. In hydrostatic restoring matrix, C(6,4) and C(6,5) should always be zero and C(4,6) and C(5,6) may be non-zero. Automatically calculated body.hydroForce.linearHydroRestCoef seems to be the transpose of the correct result if I do not specify obj.hydroStiffness.
![image](https://user-images.githubusercontent.com/106663159/231125214-c04c248b-761c-4901-8ef2-aec86fe5df60.png)
2. In AQWA theory manual, C(4,4), C(5,5), C(4,6), and C(5,6) use displacement mass but not body mass, but those in WAMIT use body mass. For floating bodies with moorings, these two results are not the same. Therefore, a correction of C(4,4), C(5,5), C(4,6), and C(5,6) is necessary for AQWA.

![image](https://user-images.githubusercontent.com/106663159/231127324-842c0126-d485-443e-bb37-be4b7b5c7725.png)
![image](https://user-images.githubusercontent.com/106663159/231127394-023aab00-201c-431e-b26c-bdc6435081a3.png)

**WEC-Sim file(s)**
_If known, please identify the the WEC-Sim files causing the bug or error:_
_Please upload relevant case files if possible to speed debugging._

**To Reproduce**
_Steps to reproduce the behavior:_
1. _Go to '...'_
2. _Click on '....'_
4. _Scroll down to '....'_
5. _See error_

_State method of running WEC-Sim (command line, from simulink with input file, from simulink with custom parameters)_

**Expected behavior**
_A clear and concise description of what you expected to happen._

**Screenshots**
_If applicable, add screenshots to help explain your problem. For example, a screen shot of the exact error message from the MATLAB command line._

**Workstation Information (this section MUST BE completed):**
 - OS: [e.g. iOS8.1, Windows 10, Linux 5.12.1]
 - [MATLAB/Simulink Release](https://www.mathworks.com/help/matlab/release-notes.html) [2019b, 2020a, 2020b, 2021a]
 - [WEC-Sim Release](http://wec-sim.github.io/WEC-Sim/man/release_notes.html) [v3.1, v4.0, v4.1]

**Additional context**
_Add any other context about the problem here._

nathanmtom:
[...]

PR #1063, opened 2023-05-23
Update writeBEMIOH5.m

salhus:
The commit addresses issue #1028 in a better way.
Although, BEMIO was reading the AQWA data correctly, the writeBEMIOH5 did not do the necessary transpose.

BEMIO was reading the AQWA data correctly, but the writeBEMIOH5.m function was not doing the necessary transpose.  
For reference, here is a check that readAQWA function was working correctly.
* In the images that follow, the black background images are from MATLAB, and white background images are from the H5 file viewer.

image

However, when I looked at the h5 file in an h5 file viewer, the writeBEMIOH5 function should have taken a transpose.

image

Following the fix I made, here is how the hydrostatics matrix looks, confirming that the writeBEMIOH5 and readBEMIOH5 are working as intended.

image

and here is the h5 file viewer,

image

### DIFF
--- a/source/functions/BEMIO/readBEMIOH5.m
+++ b/source/functions/BEMIO/readBEMIOH5.m
@@ -48,8 +48,8 @@ hydroData.properties.dofEnd   = (number-1)*6+6;
 
 % Update if DOFs included in hydroData
 try hydroData.properties.dof       = h5read(filename,[h5BodyName '/properties/dof']);       end
-try hydroData.properties.dofStart = h5read(filename,[h5BodyName '/properties/dof_start']); end
-try hydroData.properties.dofEnd   = h5read(filename,[h5BodyName '/properties/dof_end']);   end
+try hydroData.properties.dofStart  = h5read(filename,[h5BodyName '/properties/dof_start']); end
+try hydroData.properties.dofEnd    = h5read(filename,[h5BodyName '/properties/dof_end']);   end
 
 % Read hydrostatic stiffness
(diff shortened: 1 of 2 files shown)


nathanmtom:
@salhus This PR does not appear to rollback the other PR merged to resolve the question on the hydrostatic stiffness matrix. Does this PR mean it still needs those changes or are you planning to rollback that PR then merge here?

nathanmtom:
@salhus It does appear that the readBEMIOH5.m file in this branch does not have the extra line to zero out the final yaw row, but in WEC-Sim 'Dev' that line can be found. Given this is a small change I am requesting if you can close this issue, and submit a new PR after setting up a branch with an updated 'Dev' branch. 

Also, have you checked to see if this impacts the hydrostatic matrix for NEMOH, Capytaine, and WAMIT? When I reviewed the hydrostatic matrix for the AQWA and NEMOH, we appear to be importing the 6x6 matrix where as for WAMIT it is only the upper right diagonal. Therefore, we shouldn't need any transpose for AQWA and NEMOH and a fix should then occur only in the readWAMIT.m to correct for the different importation of the hydrostatic stiffness matrix.

salhus:
@nathanmtom,
re: New PR,
Sure, sorry about the confusion.

re:  Hydrostatics matrix transpose,
Although, the readAQWA.m imports the hydrostatics matrix correctly to MATLAB workspace, as I show above, the writeBEMIOH5.m currently writes it as a transpose to the h5 file. The h5file is being written in two stages, 1. Importing data to MATLAB, 2. and, then writing the h5 file from the MATLAB workspace. 
So the fix needs to be done for readWAMIT and writeBEMIOh5.

I describe this more clearly in the new PR.

PR #1065, opened 2023-05-24
fix_readWAMIT_and_writeBEMIOh5

salhus:
@nathanmtom 

As mentioned in the last comment for #1063 the fix needs to be done to readWAMIT and writeBEMIOh5.
Here is an illustration.
I modified an .out file to have $(4,6)$ and $(5,6)$ terms. 
modOut

Here is how readWAMIT would read it to MATLAB workspace,

readWAMITout

I then fixed the readWAMIT to have zeros in the $6^{th}$ row. The readWAMIT function then read the .out file correctly to the MATLAB workspace but the writeBEMIOh5 still automatically does a transpose (perhaps the writing indices are switched in the writing function),

Here is how readWAMIT read the .out file,
after_readWAMIT_fix

While after the writeBEMIOh5 is done the h5 file is written like this,
h5File_readWAMIT_fix

After having the trasnpose done in the writeBEMIOh5 as well, it can be seen that readWAMIT reads the matrix correctly to the workspace, and the writeBEMIOh5 writes the h5 file correctly,

here is readWAMIT reading the matrix correctly to the workspace,
after_readWAMIT_and_writeBEMIOh5_fix

and here is the h5 file correctly written,
h5_after_readWAMIT_and_writeBEMIOh5_fix

### DIFF
--- a/source/functions/BEMIO/readWAMIT.m
+++ b/source/functions/BEMIO/readWAMIT.m
@@ -93,6 +93,7 @@ for n = 1:N
         tmp = textscan(raw{n+3}(find(raw{n+3}==':')+1:end),'%f');
         hydro(F).Khs(5,5:6,hydro(F).Nb) = tmp{1};
         hydro(F).Khs(5:6,5,hydro(F).Nb) = tmp{1};
+        hydro(F).Khs(6,:)               = 0*hydro(F).Khs(6,:); % 6th row must be zeros
     end
     if isempty(strfind(raw{n},'Wave period'))==0
         if isempty(strfind(raw{n},'Wave period = infinite'))==0  T = 0;  end
--- a/source/functions/BEMIO/readWAMIT.m
+++ b/source/functions/BEMIO/readWAMIT.m
@@ -92,7 +92,6 @@ for n = 1:N
         hydro(F).Khs(4:5,4,hydro(F).Nb) = tmp{1}(1:2);
         tmp = textscan(raw{n+3}(find(raw{n+3}==':')+1:end),'%f');
         hydro(F).Khs(5,5:6,hydro(F).Nb) = tmp{1};
-        % 6th row must be zeros
     end
     if isempty(strfind(raw{n},'Wave period'))==0
         if isempty(strfind(raw{n},'Wave period = infinite'))==0  T = 0;  end
--- a/source/functions/BEMIO/readWAMIT.m
+++ b/source/functions/BEMIO/readWAMIT.m
@@ -89,11 +89,10 @@ for n = 1:N
         hydro(F).Khs(3:5,3,hydro(F).Nb) = tmp{1};
         tmp = textscan(raw{n+2}(find(raw{n+2}==':')+1:end),'%f');
         hydro(F).Khs(4,4:6,hydro(F).Nb) = tmp{1};
-        hydro(F).Khs(4:6,4,hydro(F).Nb) = tmp{1};
+        hydro(F).Khs(4:5,4,hydro(F).Nb) = tmp{1};
         tmp = textscan(raw{n+3}(find(raw{n+3}==':')+1:end),'%f');
         hydro(F).Khs(5,5:6,hydro(F).Nb) = tmp{1};
-        hydro(F).Khs(5:6,5,hydro(F).Nb) = tmp{1};
-        hydro(F).Khs(6,:)               = 0*hydro(F).Khs(6,:); % 6th row must be zeros
+        % 6th row must be zeros
     end
     if isempty(strfind(raw{n},'Wave period'))==0
         if isempty(strfind(raw{n},'Wave period = infinite'))==0  T = 0;  end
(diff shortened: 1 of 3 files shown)


nathanmtom:
@salhus Thanks for double checking the implementation here. As I'm reviewing this PR one quick request is that in 'readBEMIOH5.m' to delete line 56 where you originally zeroed the 6 row of the hydrostatic matrix. I'll follow-up with comments directly on the files you've changed so far.

salhus:
@nathanmtom Thanks for pointing that. I just made a commit that removes the line 56.

source/functions/BEMIO/writeBEMIOH5.m
```diff
@@ -67,7 +67,7 @@ function writeBEMIOH5(hydro)
         writeH5Parameter(filename,['/body' num2str(i) '/hydro_coeffs/linear_restoring_stiffness'],tmp(1,m_add + 1:m_add + m,:),'Hydrostatic stiffness','N/m');
         clear tmp;
     else
-        writeH5Parameter(filename,['/body' num2str(i) '/hydro_coeffs/linear_restoring_stiffness'],hydro.Khs(:,:,i),'Hydrostatic stiffness matrix','');        
+        writeH5Parameter(filename,['/body' num2str(i) '/hydro_coeffs/linear_restoring_stiffness'],hydro.Khs(:,:,i)','Hydrostatic stiffness matrix','');
```
nathanmtom, line 70: @salhus I do not have an .h5 viewer, but you mention that this is flip is to make the viewer have the right matrix format; however, we then use the reverseDimensionOrder in the readBEMIOH5 to get it back in the original form for WEC-Sim. Therefore, we are flipping it forward and flipping it backward. I tried running without the transpose and then removing the reverseDimensionOrder and the WAMIT and AQWA matrices were in the right format for WEC-Sim. 

Therefore, my question is if we can clean up this implementation to not have unnecessary flipping if it is only for the .h5 viewer. As mentioned above the results come out as we want, just seeing if the implementation can be modified.

source/functions/BEMIO/readWAMIT.m
```diff
@@ -93,6 +93,7 @@
         tmp = textscan(raw{n+3}(find(raw{n+3}==':')+1:end),'%f');
         hydro(F).Khs(5,5:6,hydro(F).Nb) = tmp{1};
         hydro(F).Khs(5:6,5,hydro(F).Nb) = tmp{1};
+        hydro(F).Khs(6,:)               = 0*hydro(F).Khs(6,:); % 6th row must be zeros
```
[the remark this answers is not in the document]
salhus: That makes sense. That would remove the 6th row being written in the first place. I made a commit to that effect

source/functions/BEMIO/writeBEMIOH5.m
salhus: @nathanmtom,
I see your point that it is redundant. However, I think having the h5 file represent the coefficients is also important.
The h5 files are useful in their own right because it can be a standardised way of expressing the bemio processed results regardless of BEM software being used. This is helpful when comparing hydrodynamic coefficients, or importing it to optimizers etc.

nathanmtom: @salhus I understand and agree with that point. But now wondering why the .h5 viewer does a flip when importing into its own data format. Did you have the opportunity to verify if other .h5 viewers provided the same result? As perhaps if there is an import issue with the one .h5 viewer you are testing while others do not have this issue then perhaps we eliminate this redundancy and note one .h5 viewer is perhaps importing incorrectly. This might be a non-issue just do not personal use .h5 viewers at this time.

salhus: @nathanmtom 
Yes, I did check with another h5viewer, and observed the same dimension order.

source/functions/BEMIO/readWAMIT.m
```diff
@@ -89,10 +89,10 @@
         hydro(F).Khs(3:5,3,hydro(F).Nb) = tmp{1};
         tmp = textscan(raw{n+2}(find(raw{n+2}==':')+1:end),'%f');
         hydro(F).Khs(4,4:6,hydro(F).Nb) = tmp{1};
-        hydro(F).Khs(4:6,4,hydro(F).Nb) = tmp{1};
+        hydro(F).Khs(4:5,4,hydro(F).Nb) = tmp{1}(1:2);
         tmp = textscan(raw{n+3}(find(raw{n+3}==':')+1:end),'%f');
         hydro(F).Khs(5,5:6,hydro(F).Nb) = tmp{1};
-        hydro(F).Khs(5:6,5,hydro(F).Nb) = tmp{1};
+        % 6th row must be zeros
```
nathanmtom, line 95: @salhus Minor comment, can we delete this line. Without more context a user might not understand why and we do not provide greater description in our documentation. Therefore, requesting to either delete, or reference the AQWA or WAMIT manuals.

salhus: Sounds good. I just made that commit.

nathanmtom:
Thanks for bearing with me and answering my questions on a fairly straightforward PR, but still nice to have documentation of the decisions made if we need to revisit in the future. Merge coming shortly.

salhus:
Thanks @nathanmtom ! I appreciate it

PR #1070, opened 2023-06-05
Multiple wave-trains

salhus:
This PR introduces the capability to have multiple wave fronts each with their own wave-heights, wave-periods, direction, and spreading. Each wave-train is used to calculate its respective wave excitation forces by using code-generation to have an additional instance of wave-excitation force block. The wave-trains that shall be used in the simulation are defined using 2 steps,
1.  Using a function waveGen to generate the necessary wave elevations for the respective wave-trains, e.g. >>waveGen(waves1)
2. Defining waves.waveGroup to contain wave-trains to be included in the simulation, e.g. waves.waveGroup = [waveGen(waves1);waveGen(waves2)]

An example is available here https://github.com/salhus/WEC-Sim_Applications/tree/N-wave-Example. 
Use the 'Passive_Yaw_with_N_Waves' example to see a demo.

Relevant documentation is under development.

relates to https://github.com/WEC-Sim/WEC-Sim/projects/58#card-64154778


akeeste (changes requested):
@salhus I did an initial review of this PR, but based on our discussion last week I think some changes are missing. Were there not significant changes to the bodyClass to give the excitation forces an additional dimension? I don't see any updates to the bodyClass right now. Let me know if I'm missing something here.

Noting a couple other things we should do:
- [x] Update ``waves.marker`` in the ``waveClass`` definition to include the new visualization parameters added to the library. Ensure the documentation reflects the new parameters too
- [x] Move ``waveGroup`` to a ``waves()`` variables. This simplifies the workflow for the user and allows us to utilize the ``waves.setup()`` function already in ``initializeWecSim``. I started working on this, but got stuck on the ``bodyClass.hydroForcePre`` calculations.
- [x] Add several checks on the ``waves`` (or ``waveGroup``) variable. Users should only be able to define multiple waves of the same type and convolution method, or else the hydroForce parameters will not be accurate. Some of these checks can go in ``waves.checkInputs()``

- [x] I'll make a PR into your branch that makes progress on the second point above. See https://github.com/salhus/WEC-Sim/pull/10

salhus:
@akeeste
Hi Adam, 

I reviewed your changes and seems to resolve the comments you had about the data management and how the multiple instances of waves are being handled.
If there is no more work to add, I am moving to the docs right?

Issue #585, opened 2021-05-12
Review implementation of convolution integral in radiation problem


johannes-spinneken:
![RunTime](https://user-images.githubusercontent.com/49918117/117899374-da697d00-b27b-11eb-99ad-925fab9eb398.png)

**Note: italicized text below is include as an example and should be updated before submission. If you feel any section is not applicable to your request, please replace with 'N/A' rather than delete a section.** 

**Is your feature request related to a problem? Please describe.**
We compared the execution time of the convolution vs. the state-space implementation within the radiation problem. The attached image shows the comparison. For context, we ran WECSim on a Speedgoat Real-Time system using Simulink Real-Time (compiled C code generated from Simulink). While we would expect some difference between the convolution and state-space approaches, the execution time difference observed was larger than we expected. In the attached image, Baseline refers to a Baseline Speedgoat system, and Performance relates to a Performance Speedgoat system. The convolution time window was 20s.

**Describe the solution you'd like**
Review the implementation of the convolution implementation. Can the execution time of the convolution operation be improved to be closer to that of the state-space implementation?

**Describe alternatives you've considered**
We can run the state-space radiation implementation to meet our performance targets (e.g. computing a time step in <0.001s on our real-time target), but this requires additional steps (formulation of the state-space system) which may introduce inaccuracies in the radiation problem. 

**Personal project context**
Running WECSim in real-time for the FOSWEC device, such that it can be used for HIL work.

**Additional context**
N/A

yuyihsiang:
The convolution integration calculation (CIC) time really depends on your convolution time window AND time step size. If your time step size is small, e.g., 0.001s, the execution time of the convolution will be much longer than the execution time for the state-space model. In my experience, WEC-Sim simulation is about real time or slightly faster than real time on my laptop when using convolution integration calculation with a time step size of 0.01 s and a convolution time window of 30 s. In addition, the CIC was coded in MATLAB Function with some additional computation overhead, where we have to use the "persistent" variable in the function. I would expect it significantly slows down the simulation. 

Thank you for bringing up the question. We will take a look and see if it is possible to refine the code and make it faster. 

--Yi-Hsiang

kmruehl:
@johannes-spinneken we've added this to our list of priorities, but we are curious is there is a _need_ for the CI formulation using HIL, or whether SS is sufficient. Thanks!

johannes-spinneken:
[...]

PR #1071, opened 2023-06-07
Implement an FIR filter to calculate radiation forces

salhus:
This follows from #585 
This PR implements an FIR filter to calculate the radiation forces. The user needs to set a flag simu.FIR = 1 to enable this implementation.
This PR uses automatic code generation to FIR filter blocks for each mode (both diagonal and off-diagonal terms) and works with body-2-body interactions enabled.
I have modified a line of code in the RM3 example that has the simu.FIR flag for easier review.

The PR still needs documentation.


akeeste (changes requested):
Nice work @salhus!

There's a couple remaining items to fine tune this capability:
- [x] Update Line 3 in the ``FIR Filter Matrix`` mask to include body-to-body interactions. You might need to use ``body.total`` instead of ``length(body)`` because the highest level body block already defines ``body`` as ``body(1)``, ``body(2)``, etc
- [ ] Add a short line of documentation in the comment at ``WECSim_Lib_Body_Elements/Rigid Body/Hydrodynamic Body/Wave Radiation Forces Calculation/SS CI and Constant-Damping-CoeVariant Subsystem`` to state what flag will turn on the FIR option, e.g. "If another wave type and simu.FIR == 1, use the FIR filter calculation"
- [ ] Add documentation
- [ ] Add a WEC-Sim Application that compares the various radiation damping calculations. This will ensure this functionality is tested and would be a nice way to highlight the differences to users.
- [ ] Confirm if this works with GBM modes. I think right now the blocks are hard coded for 6x6 matrices

akeeste:
@salhus I look at this PR again and have a few comments and additional requests to make:
- [ ] Please document the masks of ``FIR Filter Matrix`` and ``FIR_filter_row`` with inline comments
- [ ] Please add Simulink comments inside of either the ``FIR Filter Matrix`` and ``FIR_filter_row``  blocks. Simulink's automatic formatting for programmatically generated blocks is not very good and it becomes difficult to dive into the blocks and understand what is happening.
- [ ] Is there a way to make automatic generation of ``FIR_filter_row`` blocks work better with saving the library? I had a lot of trouble saving the library because one must delete the extra blocks beforehand. But Simulink frequently recreates them when editing or navigating other parts of the library.
- [ ] Add rate transitions on either side of the ``FIR Filter Matrix`` block in ``WECSim_Lib_Body_Elements/Rigid Body/Hydrodynamic Body/Wave Radiation Forces Calculation/SS CI and Constant-Damping-CoeVariant Subsystem/FIR Filter Calculation``. This is required so that this PR is compatible is our support variable step solvers. We may also consider using ``simu.cicDt`` instead of ``simu.dt`` to match how users can manipulate the convolution integral calculation
- [ ] FIR, B2B, GBM
   - Unlikely to work, but I don't have a good case to test it with.
- [ ] FIR, B2B, non-hydro bodies
   - Doesn't work right now. Update ``nFilt`` in the ``FIR_filter_row`` to be ``nFilt = 6 + B2B*(body(1).total-1)*6;``
- [ ] FIR, GBM
- [ ] FIR, non-hydro bodies

salhus:
@akeeste 
Thanks so much for the in-depth review. Here are some responses to the list items,
 Please document the masks of FIR Filter Matrix and FIR_filter_row with inline comments
 
- okay.
 
 
 Please add Simulink comments inside of either the FIR Filter Matrix and FIR_filter_row blocks. Simulink's automatic formatting for programmatically generated blocks is not very good and it becomes difficult to dive into the blocks and understand what is happening.

- I have documented the approach for programmatically generated blocks. A general guidance can be added to the developer docs.

 Is there a way to make automatic generation of FIR_filter_row blocks work better with saving the library? I had a lot of trouble saving the library because one must delete the extra blocks beforehand. But Simulink frequently recreates them when editing or navigating other parts of the library.
 
 
 - Unfortunately, not that I know of.  Few things to keep in mind, run a wec-sim input file to populate the workspace, and make sure to delete auto-generated blocks before saving. I think Kelley and I had a discussion on this some time ago.
 
 
 Add rate transitions on either side of the FIR Filter Matrix block in WECSim_Lib_Body_Elements/Rigid Body/Hydrodynamic Body/Wave Radiation Forces Calculation/SS CI and Constant-Damping-CoeVariant Subsystem/FIR Filter Calculation. This is required so that this PR is compatible is our support variable step solvers. We may also consider using simu.cicDt instead of simu.dt to match how users can manipulate the convolution integral calculation,


- I will work on that, and check the non-hydro update you are suggesting.

I don't think the GBM or flexible bodies will be served by this implementation. However, I do not know how a GBM interacts with multiple bodies in terms of signal handling of kinematics. This could be a future update, as I am not sure it can be finished before the next release.

Thanks again for helping and taking a thorough look through the PR, I know auto-generated masks can be difficult to test and debug.

Cheers,
Sal

akeeste:
Hi Sal. Thanks for the response. Please just mark off the TODO list as you continue developing. And adding on a couple comments I didn't get pushed to the last message in time---
The GBM application (no B2B) does also fail. Can you look into this one? There's likely a hard coded 6 DOF in one of the masks.

Another thought is that maybe we can make the ``FIR Filter Matrix`` mask dependent on an input file variable like ``body``? This might prevent the initialization from running automatically. Something similar should happen in the wave marker programmatic generation. Maybe we can leverage that block to improve the block generation here

salhus:
Hi Adam @akeeste , 
I added the rate transition blocks, and the sampling time to simu.cicDt.
I set the number of row blocks to 6 and wrapped a try statement when adding additional instances, so working on the block is easier. The row adding cannot be linked to body, since each body block will have their own index, and it will mess up the rest of the logic downstream. At this point I am deciding against enabling GBM. 

In terms of development, I think this is done. The tests are passing.
I am only left with documentation.

cheers,
Sal

akeeste:
Hi @salhus

Thanks for those updates. I added a few more changes that I described above into PR salhus/WEC-Sim#11. Please review and merge that PR into your branch. I updated the rate transition blocks to specify use of ``simu.cicDt`` or else they will inherit the same time step as the rest of the simulation. I also updated the counting of DOFs and bodies in the library block so that the FIR setting can now be used with:
- [x] nonhydro bodies
- [x] B2B
- [x] GBM

Note that non-hydro bodies can't be used with B2B if they're included in the h5 file, but that is a larger issue already present in WEC-Sim. I also don't think this will work with B2B and GBM, but that is a niche case that can't be tested easily and I'm unsure if GBM+B2B is currently supported or not.

salhus:
@akeeste 
Hi Adam,
Thanks for those additions ! I merged your PR. Thanks for the changes to the class. And it seems to work for body.dof, since body.total would have not worked as intended.

So just to confirm with you the work is done, and I am moving on the documentation now.

Cheers,
Sal

nathanmtom:
@salhus Thanks for developing this feature as it provides an alternative to the primary CIC option and highlights how we can leverage automatic generation to auto build. 

After completing my initial review, I believe you have responded to @akeeste requests expect for documentation and developing an applications to compare the three different approaches to the convolution integral calculation. We also should be adding a test case to ensure this feature is not broken in the future. I do need to run the WEC-Sim Applications Test with this branch to see if there are any further bugs to fix.

In terms of next steps the priority is documentation. Secondary is the applications and test case which can be added as a project board item and addressed in the next fiscal year. I'll be reviewing the N Waves PR next and if I have time I will try to assist with the applications and test case development.

salhus:
@nathanmtom Thanks for reviewing the PR. I think I should be able to wrap-up the documentation by Monday. I have a PR on the applications repository for N-Wave PR and will add another PR that corresponds to this FIR PR tonight.

docs/user/advanced_features.rst
```diff
@@ -175,8 +175,32 @@ setting for ``simu.reloadH5Data`` in the WEC-Sim input file.
...
+1. The State-Space Representation,
+2. The FIR Filters.
+
+Here is a comparison of simulation run-times for the RM3 example simulated for 1000 s, when the radiation forces 
+are calculated using different routes:
+
+	+------------------------------------------------+-------------------------------------------+
+	|   *Radiation Force Calculation Approach*       | *Time (s)*  	                             |
+	+------------------------------------------------+-------------------------------------------+
+	|   No Convolution                               | 3.699 s    	                             |
```
nathanmtom, line 193: @salhus Minor comment: can we call No Convolution "Constant Coefficients" as this is what the block is called in the Wave Radiation Forces Calculation Block.

Second when I run the simulation I get slightly different times, so I wanted to request we report normalized times. So No Convolution would be 1 and all other times would be referenced to that. I think there will be confusion if these numbers are not reproduced by users unfortunately there is a lot of factors impacting the final values.

docs/user/advanced_features.rst
```diff
@@ -187,6 +211,20 @@ must be defined in the WEC-Sim input file, for example:
 
 .. _user-advanced-features-time-step:
 
+
+FIR Filters
```
nathanmtom, line 215: @salhus Minor comment can we spell out FIR and say "Finite Impulse Response (FIR)". Also on line 218 below write out fully the first time and then use FIR for the rest.

nathanmtom:
Thanks for updating the documentation @salhus. I have some minor requests on the documentation that I would appreciate you incorporating. Also if you can respond to @akeeste comments about reverting two files back to the original we should be ready for merging by end of this week. I'll respond with feedback on the applications case in that PR.

docs/user/advanced_features.rst
salhus: Updated the docs.
- done

docs/user/advanced_features.rst
salhus: Makes sense.
- done.

salhus:
@nathanmtom ,
Thanks Nathan.
I updated the docs and restored the RM3 example to its original state.

nathanmtom:
Thanks @salhus for addressing my comments for this PR. I accept your latest changes and all tests are passing except for the `windows-latest` which has been identified in issue #1128.

salhus:
@nathanmtom 
Thanks Nathan for patiently going through it all!

PR #1076, opened 2023-06-28
adding bemio cleanup function

dforbush2:
Not ready for merge, but eventually we want it and devs may find it useful in the interim.


akeeste:
@dforbush2 I tested this functionality on several noisy BEM cases today and it works wonderfully. Both excitation and radiation IRFs are much cleaner and more realistic. I don't have the expertise to mess with the ``deSpike`` settings that go into ``findpeaks`` or ``filtfilt``, but it works great with the defaults. I can't post screenshots of my tests because they're proprietary, but can share another time.

dforbush2:
Added a fix to readWAMIT to handle both GBM and FSP additional modes.

akeeste:
I pushed the removal of extra files and indexing update directly to your branch @dforbush2. Merging now

PR #1081, opened 2023-07-07
Feature paraview

agmoore4:
Several changes were made to improve visualization with ParaView and make WEC-Sim compatible with Paraview 5.11.1:
- Previously, the script WEC-Sim.py used BlockIndices with ExtractBlock. As of Paraview 5.10, ExtractBlock no longer works with BlockIndices. ExtractBlock was updated to use Selectors instead.
- Previously, the script pressureGlyphs.py could not find the properties it needed (Hydrostatic Pressure, etc.) because whitespace is automatically removed in calculator.Function strings. To resolve this, spaces were removed from the property names in writeParaviewBody.m (eg, HydrostaticPressure).
- The documentation for the ParaView section of user/advanced_features was updated to reflect the current examples in WEC-Sim_Applications. New animations for the two examples will be sent to Kelley to be included in the documentation.
- Previously, simu.paraview.startTime and simu.paraview.endTime were set to 0 and 100 by default in simulationClass.m, which led to the simulation and the visualization running for different amounts of time. This was changed so that they are initialized as empty in simulationClass.m. Then in paraviewVisualization.m, if still empty after reading the input file, they are set to match simu.startTime and simu.endTime, respectively.
- Previously, the displayed times in ParaView were forced to always start at t=0 by subtracting off simu.paraview.startTime in paraviewVisualization.m. This was changed for clarity so that ParaView always refers to the beginning of the simulation as t=0, meaning that the visualization's first time step will always be the same value as simu.paraview.startTime.

### DIFF
--- a/docs/_include/viz.rst
+++ b/docs/_include/viz.rst
@@ -20,14 +20,14 @@ For more information about using ParaView for visualization, refer to the :ref:`
 
 Install ParaView and WEC-Sim Macros
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-First, install `ParaView 5.9.1 <http://www.paraview.org/>`_.  
+First, install `ParaView 5.11.1 <http://www.paraview.org/>`_.  
 Then, add the WEC-Sim specific macros with the following steps:
 
 * Open ParaView
 * Click on ``Macros => Add new macro``
 * Navigate to the WEC-Sim ``source/functions/paraview`` directory
 * Select the first file and click ``OK``
-* Repeat this for all files in the ``source/functions/paraview`` directory
+* Repeat this for all .py files in the ``source/functions/paraview`` directory
 
 
 ParaView Visualization Parameters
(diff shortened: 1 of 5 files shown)


kmruehl (commented):
Thanks @agmoore4 for your Paraview PR. My apologies for the delay reviewing it. Overall this looks really good. I had a comment regarding `paraview.dt`, otherwise I think this is ready for a merge.

agmoore4:
Implemented the requested changes.

kmruehl:
@agmoore4 this looks good, and it's ready for a merge once you update the embedded videos. The update should be something like this for the RM3:

```
.. raw:: html

	<iframe width="560" height="315" src="https://www.youtube.com/embed/yL6LHdYTBIo" title="YouTube video player" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture; web-share" allowfullscreen></iframe>
```

and this for nonlinear:
```
.. raw:: html

	<iframe width="560" height="315" src="https://www.youtube.com/embed/JfKxQ1AgQBk" title="YouTube video player" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture; web-share" allowfullscreen></iframe>
```

kmruehl:
Thanks @agmoore4. As soon as these tests run and pass I'll merge the PR

kmruehl:
All tests are passing, so I'm going to merge this PR. Thanks again!

Issue #1033, opened 2023-04-18
[WEC-Sim Applications] Poor power output with direct linear PTO


tedman223:
**Note: italicized text below is include as an example and should be updated before submission. If you feel any section is not applicable to your request, please replace with 'N/A' rather than delete a section.**

**Cite the [WEC-Sim applications](https://github.com/WEC-Sim/WEC-Sim_Applications) case in question**
_Please identify the WEC-Sim applications case in question._

**Describe the bug or issue**
I am currently simulating a two body point absorber with the direct linear genrator PTO. I am getting results, how ever it is not good data. I believe that I am having some incorrect numbers in my PTO variables that are resulting in this. I have carefully selected what I beleive are resonable values and have iterated with all these values to see what works best. I am hoping to get a more resonable power, voltage, and current outputs as I have not been able to produce anything about 0.9 Watts. 

**To Reproduce**
_Steps to reproduce the behavior:_

N/A

**Expected behavior**

I am expecting at least 10 Watts of Power is not more. 

**Screenshots**
![code](https://user-images.githubusercontent.com/126424529/232669702-04ccc864-5d97-42cf-bb5b-5d39010f66b7.jpg)

![power](https://user-images.githubusercontent.com/126424529/232669978-b5fe1094-a799-4d3d-a6cf-868048ac602b.jpg)

![voltage](https://user-images.githubusercontent.com/126424529/232669994-a6894733-53e5-4f83-b50c-427af3c62869.jpg)

![current](https://user-images.githubusercontent.com/126424529/232670006-fbd79ae5-bd7c-4873-b322-01829b7e3b3b.jpg)

**Workstation Information (this section MUST BE completed):**
 - OS: [Windows 11]
 - [MATLAB/Simulink Release](https://www.mathworks.com/help/matlab/release-notes.html) [ 2022a]
 - [WEC-Sim Release](http://wec-sim.github.io/WEC-Sim/man/release_notes.html) [v4.1]

**Additional context**

This is for the MECC competition

dav-og:
Hello @tedman223,

Its a bit difficult to say if/why your power output is too low without knowing more around your model inputs. Would it be possible to share your input files? (Feel free to email them if you don't feel comfortable uploading here).

Also, I'm wondering:
- If you tried a linear PTO model at first? (i.e. just constant coefficients for stiffness and damping)
- How you determined the parameters for your direct drive PTO?

Cheers,
Dave

tedman223:
[...]

PR #1085, opened 2023-07-25
Fix output order for direct drive PTO

jtgrasb:
Output order of direct drive PTO is off and leads to mix-up between electrical power and velocity. 

Note: this is also likely the cause of an old issue: #1033 .


kmruehl:
@jtgrasb is this is a bug, should it go in master? @jleonqu  can you review this PR?

PR #1092, opened 2023-08-02
use capytaine v2 to compute hydrostatics

dav-og:
added a check:
- if capytaine version < 2.0, still use meshmagick to compute hydrostatics


salhus:
@dav-og I checked the BEMIO examples using the modified call_capytaine.py with both Capytaine 1.5, and Capytaine 2.0 dev.
It works as expected ! I think this can be merged.

PR #1095, opened 2023-08-07
Fix Direct Drive PTO Output order

jtgrasb:
Output order of direct drive PTO is off and leads to mix-up between electrical power and velocity.

Note: this is also likely the cause of an old issue: https://github.com/WEC-Sim/WEC-Sim/issues/1033 .


jleonqu (approved):
Thanks for catching this bug @jtgrasb ! The changes look good to me.

Issue #1090, opened 2023-07-31
[BUG] readAQWA phase


DeminLiLi:
**Note: italicized text below is include as an example and should be updated before submission. If you feel any section is not applicable to your request, please replace with 'N/A' rather than delete a section.**
The  phase of wave excitation force in two-body system in "readAQWA.m"
**Describe the bug**
_A clear and concise description of what the perceived bug is._
I am currently using version 5.0.1 of wecsim. When calculating the double body, I found that the heave relative motion of the double body is different from the test results. I used version 4.3 as a verification and found that the calculation results of version 4.3 are consistent with the test values. After It was identified that there was a problem with the phase of the wave excitation force in version 5.0.1.
When using wecsim version 5.0.1 for two-body calculation, the phase difference in the heave direction between the inner float and the wave increases by 180°, resulting in a 180° phase difference in the relative motion of the heave between the two floats. Directional motion becomes counter-directional motion, and eventually the relative motion amplitude of the heave becomes the superposition of the amplitudes of the two floats. This is obviously wrong. After inspection, it was found that the inner float in the heave motion did not call hydro(F).ex_ph(3,ind(2),ind(3)) but called  hydro(F).ex_ph(((ind(1)-1)*6+1):(ind(1)*6),ind(2),ind(3))  in "readAQWA.m", but the outer float was called by the correct wave excitation force.
**WEC-Sim file(s)**
_If known, please identify the the WEC-Sim files causing the bug or error:_
_Please upload relevant case files if possible to speed debugging._
The line 168 and 169
![readAQWA](https://github.com/WEC-Sim/WEC-Sim/assets/134644090/84563089-c801-446e-be28-ab974021a0f1)
I compared the experimental results with the calculation results of different versions. The blue line, orange line and gray line are the phase difference of outer float and wave, the inner float and wave, and the phase difference between the two floats calculated with the 5.0.1 version of readAQWA.m. The phase difference results of the yellow line, the cyan line and the green line are calculated with the 4.3 version of readAQWA.m, and the black line is the phase difference of the two floats in the test. It can be seen from the figure that the phase difference between the inner float and the relative motion calculated by the wave excitation force phase of version 4.3 is correct and consistent with the test results. In the modified readAQWA.m, the phase  of the original wave excitation force is updated to the phase of line 168 which is same with the version 4.3.
[...]

Issue #865, opened 2022-05-11
[Theory or Implementation] Differences in AQWA and WEC-Sim conventions


salhus:
**The development team requests that before submitting a question on the theory and implementation behind WEC-Sim to first please:**
1. Read through our [WEC-Sim Documentation](http://wec-sim.github.io/WEC-Sim/man/theory.html) with additional description found in our [Advanced Features Section](http://wec-sim.github.io/WEC-Sim/man/advanced_features.html)
1. Review MATLAB source code ([functions](https://github.com/WEC-Sim/WEC-Sim/tree/master/source/functions), [objects](https://github.com/WEC-Sim/WEC-Sim/tree/master/source/objects)) and the Simulink blocks ([libraries](https://github.com/WEC-Sim/WEC-Sim/tree/master/source/lib))
1. Review the curated [WEC-Sim Examples](https://github.com/WEC-Sim/WEC-Sim/tree/master/examples) and the [WEC-Sim applications](https://github.com/WEC-Sim/WEC-Sim_Applications)
1. Review the [WEC-Sim issues page](https://github.com/WEC-Sim/WEC-Sim/issues) which allows you to search past resolved issues or questions that might help answer your own question.

**Note: italicized text below is include as an example and should be updated before submission. If you feel any section is not applicable to your request, please replace with 'N/A' rather than delete a section.**

**Is your question request related to a problem? Please describe.**
_A clear and concise description of what the problem is. Ex. I do not understand why I get 'X' result when running 'Y' simulation [...]_

This issue is a continuation of questions raised by @tuisong1106 in the context of #855.
I would request @tuisong1106 to use the format here to describe their question here so that it can be resolved.

**Describe the theory or implementation approach that you have a question**
_Please clearly describe the assumptions, equations, or code script that you would like clarification. Also, please provide references to other research papers, open-source material, or other information that may be relevant in the WEC-Sim developers understanding your question._

**Personal project context**
_If this request is related to your project, please describe what you are attempting to do in your project so the WEC-Sim development team knows your end goal._

**Describe the type of conclusion or resolution you are looking for**
_[Ex.] I am still unclear on how an irregular wave train is generated within WEC-Sim and would appreciate being pointed to where in the documentation and code where I can find additional information._

**Additional context**
_Add any other context or screenshots about the feature request here._

hachikoi1:
Hi [salhus](https://github.com/salhus)
[...]

PR #1096, opened 2023-08-07
Update readAQWA.m

salhus:
WAMIT/WEC-Sim and AQWA use different conventions to write the wave potential. 
Here is the WEC-Sim convention,

image

and here is the AQWA convetion,
image

The difference in conventions means a negative sign is needed to convert the phase from AQWA to WEC-Sim conventions. However, while previously addressing this issue in #855 an additional phase of pi was being added. 

This was wrong.

After revisiting the PR in issue #1090 it has become clear to me that an additional phase of pi is not needed for non-heave modes. The wrong phase was being added because sample AQWA cases in the WEC-Sim repository had 3 wave headings [-180, 0 , 180], instead of the corresponding WAMIT example that only had 0.

I would like to thank @DeminLiLi @hachikoi1 in helping me identify the issue.

The other issues and threads that are relevant are listed here, so that future referees of those threads are referred to here.  #865 #846

### DIFF
--- a/source/functions/BEMIO/readAQWA.m
+++ b/source/functions/BEMIO/readAQWA.m
@@ -165,8 +165,7 @@ for ln = n:length(raw1)
                     tmp2 = str2num(raw1{ln+(k-1)*hydro(F).Nh*hydro(F).Nf*2+(j-1)*hydro(F).Nf*2+(i-1)*2+2});
                     ind = tmp1(1:3); tmp1(1:3)=[];
                     hydro(F).ex_ma(((ind(1)-1)*6+1):(ind(1)*6),ind(2),ind(3)) = tmp1; % Magnitude of exciting force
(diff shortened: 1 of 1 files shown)


nathanmtom (approved):
@salhus As you mentioned, this is a very minor fix and is consistent with how the readNEMOH.m  and BEMrosetta handle the phase angle. I approve this PR under the condition that you confirm if this should be pulled into 'Dev' or 'Master.' Once confirmed, and adjusted as needed, I will merge.

salhus:
@nathanmtom 
Hi Nathan,
Thanks for the quick follow-up. I think this should go in master, as my branch worked off of dev and will have conflicts with master. I think it will be easier for this to go to dev for now, and then the merge to dev will be streamlined (given we will be doing it within 6-8 weeks).

Cheers,
Sal

kmruehl:
@salhus I just updated your fork to be current with master, and then added your bugfix in commit [e2c6d24](https://github.com/WEC-Sim/WEC-Sim/pull/1096/commits/e2c6d24b4087ef694a96134fd995ed11bc61baf9). Please revise this PR to merge into WEC-Sim:master, and then we can merge it.

Issue #1099, opened 2023-08-09
[BUG] Potential Bug: Accelerator and Rapid Accelerator modes giving errors


jtgrasb:
**Note: italicized text below is include as an example and should be updated before submission. If you feel any section is not applicable to your request, please replace with 'N/A' rather than delete a section.**

**Describe the bug**
When trying to run WEC-Sim using the accelerator or rapid-accelerator mode, I get the following error:
<img width="721" alt="image" src="https://github.com/WEC-Sim/WEC-Sim/assets/87095491/13cd9db5-19b8-423f-ad5d-f30a019827cb">

**WEC-Sim file(s)**
Unknown

**To Reproduce**
_Steps to reproduce the behavior:_
1. Go to RM3 example
2. Set simu.mode = 'accelerator'
3. Run WecSim from command window

**Expected behavior**
WEC-Sim to run without error

**Screenshots**

**Workstation Information (this section MUST BE completed):**
 - OS: Windows 10
 - [MATLAB/Simulink Release](https://www.mathworks.com/help/matlab/release-notes.html) 2021b and 2022b
 - [WEC-Sim Release] v5.0

**Additional context**_

kmruehl:
resolved by #1100

PR #1100, opened 2023-08-09
Fix accelerator modes issue

jtgrasb:
This PR attempts to fix the error with the accelerator modes from Issue #1099. It seems to be an issue with the temp folder used to store the accelerator (MEX) files. 

This fix allows the accelerator modes to run without errors. I do not know the intended use-case of the accelerator modes though, so I find it difficult to be sure if they're working as intended.


kmruehl:
@jtgrasb I confirmed that this bug-fix works. Thank you so much! I'll merge it into master, then I'll open a new PR to pull master into dev.

PR #1106, opened 2023-08-21
Add simple direct drive PTO model

jtgrasb:
This PR adds a simple direct drive PTO model, which I will use to create a controls + PTO-Sim applications case.


source/objects/ptoSimClass.m
```diff
@@ -103,6 +103,13 @@
             'rho',                              'NOT DEFINED',...           % Fluid density
             'k1',                               'NOT DEFINED',...           % Valve coefficiente
             'k2',                               'NOT DEFINED')              % Valve coefficient, it's a function of the other valve variables
+        simpleDirectDrivePTO (1,1) struct      = struct(...                 % simple direct drive linear PTO Block properties
+            'Kt',                               'NOT DEFINED',...           % Torque constant
+            'Ng',                               'NOT DEFINED',...           % Gear ratio
+            'J',                                'NOT DEFINED',...           % Drivetrain inertia
```
jleonqu, line 109: I suggest to change the names of these variables for something more comprehensive. Instead of "J", something like "drivetrainInertia", instead of "B" something like "drivetrainFriction", etc.

jleonqu:
I have tested this PR locally and it's working. I just have one minor comment about the variable naming (see comment above)

jtgrasb:
Thanks @jleonqu, I've updated the variable names.

jleonqu:
Thanks for the update @jtgrasb.
There is one more comment from me regarding the docs. Please see this [comment](https://github.com/WEC-Sim/WEC-Sim_Applications/pull/40#issuecomment-1714231386) on the WEC-Sim Applications repo.

PR #1108, opened 2023-08-22
Control+pto docs

jtgrasb:
Documentation to go with new controls + PTO applications case (https://github.com/WEC-Sim/WEC-Sim_Applications/pull/40)


docs/user/advanced_features.rst
```diff
@@ -1243,6 +1244,50 @@ computation time and complex setup.
...
+sphere body with a simplified direct drive PTO model to maximize electrical power. Within 
+the Simulink subsystem for determining the PTO force, the controller prescribes the ideal or 
+desired force which is fed into the direct drive PTO. The current in the generator is then 
+used to control the applied force. 
+
+.. figure:: /_static/images/piPTOSimulink.png
+   :width: 500pt 
+   :align: center
+
+The PTO parameters used for this example are defined in the wecSimInputFile and correspond to
```
nathanmtom, line 1266: Editorial comment: can we use `wecSimInputFile.m` instead of the current labeling to assist in defining there is a separate file that can be referenced. Also not possible now, but can we add a citation, link, or point to the soon to be included applications case?

docs/user/advanced_features.rst
```diff
@@ -1243,6 +1244,50 @@ computation time and complex setup.
...
+
+.. figure:: /_static/images/piPTOSimulink.png
+   :width: 500pt 
+   :align: center
+
+The PTO parameters used for this example are defined in the wecSimInputFile and correspond to 
+the Allied Motion Megaflux Frameless Brushless Torque Motors–MF0310 :cite:`allied`. The 
+results for the applied gains from Section :ref:`control-reactive` are shown in the figure 
+below. Mechanical power maximization requires significant net input electrical power 
+(signified by red bar - negative power is power harvested from the system). Thus,
```
nathanmtom, line 1270: Editorial comment: Add "...-negative power, green bars, is power harvested..."

docs/user/advanced_features.rst
```diff
@@ -1243,6 +1244,50 @@ computation time and complex setup.
...
+.. figure:: /_static/images/piPTOSimulink.png
+   :width: 500pt 
+   :align: center
+
+The PTO parameters used for this example are defined in the wecSimInputFile and correspond to 
+the Allied Motion Megaflux Frameless Brushless Torque Motors–MF0310 :cite:`allied`. The 
+results for the applied gains from Section :ref:`control-reactive` are shown in the figure 
+below. Mechanical power maximization requires significant net input electrical power 
+(signified by red bar - negative power is power harvested from the system). Thus, 
+instead of harvesting electrical power, power would need to be used to achieve mechanical
```
nathanmtom, line 1271: Editorial comment: Change test to read, "..., power would need to be taken from the grid or energy storage component to achieve...

docs/_static/images/reactiveWithPTOCC.png
nathanmtom, line 1: Editorial Comment: Request that rather than time averaged power, can we plot as capture width. Understand the point here is to see the relative difference between the green and red bars, but myself forgetting the size of the sphere and the wave conditions you have used makes the power numbers difficult to relate too. I think plotting against available wave power would also highlight how much power is needed to drive the motor relative to what is available in the waves.

docs/_static/images/reactiveWithPTOOpt.png
nathanmtom, line 1: Editorial Comment: Similar to the reactiveWithPTOCC.png, if changing from power to capture width request for this image as well.

docs/user/advanced_features.rst
```diff
@@ -1243,6 +1244,50 @@ computation time and complex setup.
...
+desired force which is fed into the direct drive PTO. The current in the generator is then 
+used to control the applied force. 
+
+.. figure:: /_static/images/piPTOSimulink.png
+   :width: 500pt 
+   :align: center
+
+The PTO parameters used for this example are defined in the wecSimInputFile and correspond to 
+the Allied Motion Megaflux Frameless Brushless Torque Motors–MF0310 :cite:`allied`. The 
+results for the applied gains from Section :ref:`control-reactive` are shown in the figure
```
nathanmtom, line 1268: Editorial/minor: Can we add another sentence or two to clarify how the different power levels are calculated. The image piPTOSimulink.png has blue and dashed boxes that are I believe supposed to indicate each step in the power conversion chain; however, clarity on what signals are being used in the calculation would minimize confusion and likely an issue/clarification request. If the piPTOSimulink.png is editable, perhaps putting stars or another indicator on where in the PCC power is being calculated would add clarity.

docs/_static/images/reactiveWithPTOSweep.png
nathanmtom, line 1: Editorial comment: Can the sea state properties be added for clarity?

docs/user/advanced_features.rst
```diff
@@ -1243,6 +1244,50 @@ computation time and complex setup.
...
+instead of harvesting electrical power, power would need to be used to achieve mechanical 
+power maximization. 
+
+.. figure:: /_static/images/reactiveWithPTOCC.png
+   :width: 300pt 
+   :align: center
+
+On the other hand, by testing different controller gains, the gains which optimize for 
+maximum electrical power can be found as shown below. Increasing the proportional gain 
+and decreasing the integral gain magnitude leads to a maximum electrical power harvest
```
nathanmtom, line 1280: Minor comment: Agree with how the changing gains leads to better electrical power, but feel the statement would be more impactful if you can add a sentence or two on why those modifications led to the improved electrical power. For example was the electric generator able to run under more efficient conditions or more that the generally efficiency curve of the electric generator results in a greater proportion of energy input than harvested.

docs/user/advanced_features.rst
jtgrasb: First part resolved. Leaving open to add link to point to applications case when possible.

docs/_static/images/reactiveWithPTOCC.png
jtgrasb: Changed plots to capture width instead of power. I am not sure if it is better or not, but I do think it is good to see the results relative to wave power. The one thing that may be confusing is that all the other controllers documentation refers to the average power, so it may be confusing to change to capture width for only this example.

docs/_static/images/reactiveWithPTOOpt.png
jtgrasb: Updated image to capture width. There is no efficiency map included in the motor datasheet. The only generator losses accounted for here are I^2R losses which is proportional to the square of applied torque. I think this is better explained in the documentation with the addition of power level explanation and explanation of why electrical power was increased.

docs/user/advanced_features.rst
jtgrasb: I added a few sentences. I decided not to indicate in the png because the different power levels are combinations of multiple power sources (ex: Mechanical is inertial, damping, torque), so I thought it may cause more confusion.

docs/_static/images/reactiveWithPTOSweep.png
jtgrasb: Sea state specified and colorbars removed.

docs/user/advanced_features.rst
nathanmtom: Thanks.

docs/_static/images/reactiveWithPTOCC.png
nathanmtom: Apologies as I should have been also reviewing the other controller documentation for consistency. Then perhaps can we include both dimensional power plots and capture width for this example? The capture width plot I think is really helpful as in the unoptimized case the electrical input is 35 times the power hitting the sphere. So my request would be to have just two figures, dimensional to be consistent with other controller examples and the capture width to relate to wave climate.

docs/_static/images/reactiveWithPTOOpt.png
nathanmtom: Thanks Jeff, for the plot see my response to the other plot that was changed to capture width. If preferred to have units in W's then suggest just having both plots. 

Thanks for the additional description of the powers being calculated and how the change in PI gains are impacting the dynamic response.

docs/user/advanced_features.rst
nathanmtom: Thanks! Think it is a great addition.

nathanmtom:
@jtgrasb Thanks for addressing my comments and requests. I'm good with this pull request, but will wait to see how you wanted to address the power figure question, power vs capture width. Once you have a decision and let me know I'll merge this PR along with the WEC-Sim Applications PR.

docs/_static/images/reactiveWithPTOCC.png
jtgrasb: No problem, I agree that having both figures provides more understanding and context, so added back the power plots. Thanks!

nathanmtom:
@jtgrasb Thanks for confirming there were no other updates to the PR. I've merged into Dev.

PR #1141, opened 2023-10-16
Nonlinear Restoring Force Implementation; wecSim Model Error Displaying Feature

AlixHaider:
In this pull request, I suggest addressing the following two issues:

1. ► Nonlinear Restoring Force Implementation:

I’m currently using the **_nonlinearHydro=2_** option to simulate nonlinear hydrodynamics. With reference to issues #886  and  #910 for _**nonLinearBuoyancy.m**_ function, I noticed that the developer team switched from 

```
% Zero out regions above the mean free surface
z=center(:,3);
idx=find((z-elv)>0);

% Calculate the hydrostatic pressure at each triangle center
pressureVect = -rho*g.*[(z-elv) (z-elv) (z-elv)].*-av;
p = rho*g.*-(z-elv);
pressureVect(idx,:)=0;
p(idx)=0;
```

to:

```
% Zero out regions above the mean free surface
z=center(:,3); z((z-elv)>0)=0;

% Calculate the hydrostatic pressure at each triangle center
pressureVect = rho*g.*[-z -z -z].*-av;
p = rho*g.*-z;
```

This switching is problematic because the exerted hydrostatic pressure should consider z-points with respect to wave elevation, i.e., ‘**(z-elv).**’ For the ellipsoid example, consider the example body displacement in heave at -1m for a given wave elevation profile. For two example points P1 and P2, it is elucidated that correct hydrostatic pressure should consider **‘(z-elv)’** instead of **‘z’**:

image

image

_Side Note:_ It is worth mentioning that the suggested change is the same technique that is implemented in nonFKForce.m calculations, where, with the change of variables definitions, we are using the same **‘(z-elv)’** to calculate pressures and then zeroes out force for panels above **‘(z-elv).’**

image

2. ► Model Error Displaying Feature in wecSim.m file:

Previously, in issue #1012, I addressed the problem of moorDyn crashing the Matlab session. I included an error handle **_‘e’_** in **_wecSim.m_** file to catch error exceptions in Simulink and close the moorDyn console without terminating the Matlab session.
I noticed that my merged change misses displaying some of the Simulink Model error messages. The cause is that I was printing _**‘e.cause{i}.messages,’**_ but sometimes Simulink can directly generate error messages via **_‘e.message’_** with an empty ‘e.cause’ field. So, I have included a code segment in **_wecSim.m_** to display not only  **_‘e.cause{i}.messages’_** but also _**‘e.message’**_ so we do not miss displaying any error messages in the Matlab command window. I also changed the name of variable **_‘e’_** to **_‘errorHandle’_**

image


akeeste:
Hi @AlixHaider 

For now I will let others address point 1 on nonlinear buoyancy, but on point 2:
I made an update to the dev branch which should fix the error handling. The new solution still catches any errors during simulation so that MoorDyn always closes, but it uses ``rethrow(e)`` to rethrow the error. This is a more robust way of getting the full error message, cause, stack, etc than a custom solution can. This change will migrate to the master branch within the next week for our upcoming release.

dforbush2:
Hi @AlixHaider 

On the nonlinear hydrostatics calculation, we have actually gone back and forth on this and the prevailing knowledge for now is best captured in #886 (please see my last comment on this issue ). The subtlety has to do with how WEC-Sim calculates non-linear hydrodynamics in a wave field (simu.nonLinearHydro ==2) by summing the nonlinear hydro function with the non-linear froude krylov function. I believe the current implementation is correct, if a bit confusing since it is divided up differently than a lot of texts do it. The only time the nonLinearHydro function is considered in isolation is if simu.nonLinearHydro==1, which should only be used in absence of wave excitation (i.e., the free surface is at 0 everywhere). 

Please review my last 2 comments on #886 and let me know if you disagree. My July 11, 2022 comment I think aligns with your reasoning, but as I later learned, this is incorrect, as summarized in my March 22, 2023 comment,

AlixHaider:
@dforbush2 I agree with you on your March 22, 2023 comment. I did a thorough literature review that agrees with your comment; in particular, equation 7 in the following reference confirms that your present implementation is correct.
[https://doi-org.proxy.hil.unb.ca/10.2478/IJNAOE-2013-0052](https://doi-org.proxy.hil.unb.ca/10.2478/IJNAOE-2013-0052)
I would retract my push request on this issue. Thank you, @dforbush2, for correcting my understanding of the nonlinear FK forces.

akeeste:
@AlixHaider the fix for correct error messages should now be pulled into the master branch. I will close this PR since the nonlinear hydro implementation has been resolved.

PR #1144, opened 2023-10-17
Path fix for BEMIO example

akeeste:
A very minor bug, part of the compareBEMIO example would break on non-windows OS due to the file seperators used. Related to a fix required in https://github.com/WEC-Sim/WEC-Sim_Applications/pull/49


kmruehl:
Thanks @akeeste

Issue #76, opened 2015-12-12
Morison Element water particle velocity z component


bradling:
Hello,

I think the z-component of the water particle velocity used in the morison element calculation has the incorrect sign. I am going to do some more investigation into this over the weekend and see if I can demonstrate this with a simple model.

bradling:
I did some more research into this, and I think the issue is not the sign of the z-component of fluid velocity, but rather the sign of the entire morison force term.

I tested this by creating a simple 1 body model, with a prescribed motion heave only PTO. I then included a morison drag term at the CG of the body, and simulated with no input waves. This should be equivalent to using a viscous drag term, as the orbital fluid velocity is zero. However, the drag force has the opposite sign from a model using a viscous drag term, see plot below. If needed I can upload the model and input file I used to generate these results.

![image](https://cloud.githubusercontent.com/assets/10558137/11794025/f91b1232-a263-11e5-9ccd-bc2ae456c4d7.png)

bradling:
Attached is my test case. The h5 file is a dummy file with zeros for hydro force coefficients.





[morison_test_case.zip](https://github.com/WEC-Sim/WEC-Sim/files/66872/morison_test_case.zip)

cmichelen:
Hi Brad,

I haven't had a chance to look at the Morrison Element implementation in detail, but I played with the test case you shared and switching the sign in the addition block seems to solve the problem, like you suggest. The best way to solve it would have been to go into the ME blocks and have it output force in the correct direction, but for now this should be good enough.  I made this change and pushed. 

This are the results before the change:
![heave_org](https://cloud.githubusercontent.com/assets/5296181/11904208/bfb17ef8-a57b-11e5-9422-9e0d88bf7044.png)
![surge_new](https://cloud.githubusercontent.com/assets/5296181/11904209/c55d283e-a57b-11e5-867e-c92691b6bc5b.png)

And this is after:
![heave_new](https://cloud.githubusercontent.com/assets/5296181/11904215/cddb0814-a57b-11e5-9625-846889421a9f.png)
![surge_new](https://cloud.githubusercontent.com/assets/5296181/11904216/d241baa6-a57b-11e5-91de-b7af149c3aab.png)

cmichelen:
Let us know if you find any other bugs with the morison elements. We just implemented it very recently.

PR #1166, opened 2023-11-13
Exclude R2020b and R2021a from Windows test runners

H0R5E:
This PR is to exclude R2020b and R2021a from the Windows test runners while they seem to be broken.

Specific changes are:
+ Merged windows and linux jobs, using the matrix to vary version differences
+ Excluded R2020b and R2021a from Windows jobs
+ Include latest in the Linux jobs

EDIT: Opened bug report in `setup-matlab` [#76](https://github.com/matlab-actions/setup-matlab/issues/76).

### DIFF
--- a/.github/workflows/run-tests-dev.yml
+++ b/.github/workflows/run-tests-dev.yml
@@ -13,13 +13,22 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
   cancel-in-progress: true
 jobs:
-  run_windows_tests:
-    runs-on: windows-latest
+  run_tests:
+    runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
+        os: [ubuntu-latest, windows-latest]
         release: [R2020b, R2021a, R2021b, R2022a, R2022b, R2023a]
-    name: MATLAB ${{ matrix.release }} on Windows
+        exclude:
+          - os: windows-latest
+            release: R2020b
+          - os: windows-latest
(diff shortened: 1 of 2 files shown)


akeeste (approved):
Approved. Thanks @H0R5E

H0R5E:
@akeeste, not sure that using the caches are worth it. If you have a look [here](https://github.com/H0R5E/WEC-Sim/actions/caches), the storage used goes well over the 10GB limit with just 3 versions of MATLAB cached. Shall I remove the option before this is merged?

akeeste:
@H0R5E what is being cached by setup-matlab? Is it saving a specific MATLAB installation for a future test?

H0R5E:
> @H0R5E what is being cached by setup-matlab? Is it saving a specific MATLAB installation for a future test?

Yeah, that's the idea. It would reduce the installation time on subsequent runs. The caches are saved per repository, but the images are 4Gb-ish each, which means only 2 versions can be realistically cached, whereas we test about 10.

H0R5E:
@akeeste, you can see the same effect on the WEC-Sim GH Actions cache [here](https://github.com/WEC-Sim/WEC-Sim/actions/caches) after running this PR.

akeeste:
@H0R5E thanks for confirming. I agree; I don't see MATLAB caching being very useful right now, given their size.

H0R5E:
OK, I've removed the caching. Mathworks are looking at the issue with the broken versions already in [#76](https://github.com/matlab-actions/setup-matlab/issues/76).

Issue #1171, opened 2023-11-22
[BUG] Moordyn Issue with V2 Matlab 2023b 


timmaflan:
**Note: italicized text below is include as an example and should be updated before submission. If you feel any section is not applicable to your request, please replace with 'N/A' rather than delete a section.**

**Describe the bug**
_A clear and concise description of what the perceived bug is._
I am experiencing a similar to issue #1164. When I run wecSim v6 with matlab 2023b I get the following error:

---------------Starting MoorDyn-----------
Error using wecSim
Error evaluating 'InitFcn' callback of SubSystem block 'RM3MoorDyn/MoorDyn/MoorDyn/MoorDyn'.
Callback string is '% Initialize MoorDyn Lib (Windows:dll or OSX:dylib)
disp('---------------Starting MoorDyn-----------')

if libisloaded('libmoordyn')
    calllib('libmoordyn', 'MoorDynClose');
    unloadlibrary libmoordyn;
end

if ismac
    loadlibrary('libmoordyn.dylib','MoorDyn.h');
elseif ispc
    loadlibrary('libmoordyn.dll','MoorDyn.h');
elseif isunix
    loadlibrary('libmoordyn.so','MoorDyn.h');
else
    disp('Cannot run MoorDyn in this platform');
end

calllib('libmoordyn', 'MoorDynInit', mooring.orientation, [0 0 0 0 0 0], mooring.moorDynInputFile);
disp('MoorDyn Initialized. Now time stepping...')
'
Caused by:
    Error using wecSim
    There was an error loading the library
    "...\Documents\WEC-Sim\source\functions\moorDyn\libmoordyn.dll"
    The specified module could not be found.

        Error using wecSim
        The specified module could not be found.

**WEC-Sim file(s)**
_If known, please identify the the WEC-Sim files causing the bug or error:_
_Please upload relevant case files if possible to speed debugging._
It seems for some reason, that matlab is unable to find libmoordyn.dll, or that there is an error with libmoordyn.dll. 

**To Reproduce**
_Steps to reproduce the behavior:_
1. place moordyn in the correct folder & ensure it is in the path. (Documents\WEC-Sim\source\functions\moorDyn)
![image](https://github.com/WEC-Sim/WEC-Sim/assets/105233738/dbe4ee16-29f4-4d0c-a867-e88281cf32a4)
2. set up the moodyn example in wecSim applications so that it has a geometry and hydro folder from wecSim's RM3 example. 
[...]

PR #1177, opened 2023-12-05
Update wecSim.m to new MoorDyn dll

jtgrasb:
This PR updates wecSim.m to reference the new MoorDyn library (`libmoordyn`) instead of the old one (`Lines`). This is an update that should have been included in the previous MoorDyn updates but was left out by mistake. Discovered in Issue #1171 

This is a bug and should be pushed to both Master and Dev.

### DIFF
--- a/source/functions/postProcessWecSim.m
+++ b/source/functions/postProcessWecSim.m
@@ -128,9 +128,9 @@ clear bodiesOutput ptosOutput constraintsOutput ptosimOutput cablesOutput moorin
 for iMoor = 1:simu.numMoorings
     if mooring(iMoor).moorDyn==1
         % Ensure that Lines library is closed before reading MoorDyn output
-        if libisloaded('Lines')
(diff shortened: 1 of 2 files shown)


hfchen20:
@jtgrasb Same thing in postProcessWecSim.m.
https://github.com/WEC-Sim/WEC-Sim/blob/master/source/functions/postProcessWecSim.m#L131-L134

FYI.  I'm able to run MoorDyn related cases, even with these `Lines` statements present.

akeeste (approved):
Thanks @jtgrasb . I reviewed and ran things locally. Tests are passing. Can you pull master into dev after this is merged?

jtgrasb:
Thanks for merging. Yes, I can do that. One note: PR #1176 is also in Master but not dev. I assume we want to also pull this into dev, correct?

akeeste:
@jtgrasb yes, #1176 and any other updates to master since the release should merge into dev too

PR #1190, opened 2023-12-26
Correct the phase of excitation force for bodies  not situated at the coordinate origin

meatballgithub:
I was recently doing a WEC-array simulation with four bodies being at distinct locations. There should be phase differences in excitation forces, while the simulation indicated that excitation forces have the same phase. The issue lies in that the frequency-domain excitation force data represents the amplitude-phase-frequency characteristics with respect to the wave elevation at the body's COG, while in WEC-Sim excitation force is always calculated using wave elevation at the coordinate origin. This will result in phase errors if the body is not situated at the origin. Therefore, I did this commit to make the frequency-domain excitation force data matching with the origin. A brief readme and a simulation test project are attached. Due to the file format and size restriction, please delete '.zip' in 'test.z01' and 'test.z02' for unzipping the simulation files.
[readme.pdf](https://github.com/WEC-Sim/WEC-Sim/files/13770319/readme.pdf)
[four-body sim test.zip](https://github.com/WEC-Sim/WEC-Sim/files/13770286/four-body.sim.test.zip)
[four-body sim test.z01.zip](https://github.com/WEC-Sim/WEC-Sim/files/13770283/four-body.sim.test.z01.zip)
[four-body sim test.z02.zip](https://github.com/WEC-Sim/WEC-Sim/files/13770284/four-body.sim.test.z02.zip)

### DIFF
--- a/source/functions/initializeWecSim.m
+++ b/source/functions/initializeWecSim.m
@@ -281,7 +281,7 @@ idx = find(hydroBodLogic==1);
 if ~isempty(idx)
     for kk = 1:length(idx)
         it = idx(kk);
-        body(it).hydroForcePre(waves(1).omega,waves(1).direction,simu.cicTime,waves(1).bem.count,simu.dt,...
+        body(it).hydroForcePre(waves(1).omega,waves(1).direction,waves(1).wavenumber,simu.cicTime,waves(1).bem.count,simu.dt,...
             simu.rho,simu.gravity,waves(1).type,waves(1).waveAmpTime,simu.stateSpace,simu.b2b);
     end; clear kk idx
 end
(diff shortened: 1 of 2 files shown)


meatballgithub:
The commit contains a bug that requires replacing 'body(idx)' with 'obj' on line 383 of 'bodyclass.m'

meatballgithub:
The commit contains a bug that requires replacing 'body(idx)' with 'obj' on line 383 of 'bodyclass.m'

meatballgithub:
The commit contains a bug that requires replacing 'body(idx)' with 'obj' on line 383 of 'bodyclass.m'

PR #1191, opened 2023-12-26
Correct the phase of excitation force for bodies not situated at the coordinate origin

meatballgithub:
I was recently doing a WEC-array simulation with four bodies being at distinct locations. There should be phase differences in excitation forces, while the simulation indicated that excitation forces have the same phase. The issue lies in that the frequency-domain excitation force data represents the amplitude-phase-frequency characteristics with respect to the wave elevation at the body's COG, while in WEC-Sim excitation force is always calculated using wave elevation at the coordinate origin. This will result in phase errors if the body is not situated at the origin. Therefore, I did this commit to make the frequency-domain excitation force data matching with the origin. A brief readme and a simulation test project are attached. Due to the file format and size restriction, please delete '.zip' in 'test.z01' and 'test.z02' for unzipping the simulation files.
[readme.pdf](https://github.com/WEC-Sim/WEC-Sim/files/13770680/readme.pdf)
[four-body sim test.z01.zip](https://github.com/WEC-Sim/WEC-Sim/files/13770682/four-body.sim.test.z01.zip)
[four-body sim test.z02.zip](https://github.com/WEC-Sim/WEC-Sim/files/13770683/four-body.sim.test.z02.zip)
[four-body sim test.zip](https://github.com/WEC-Sim/WEC-Sim/files/13770686/four-body.sim.test.zip)

### DIFF
--- a/source/functions/initializeWecSim.m
+++ b/source/functions/initializeWecSim.m
@@ -281,7 +281,7 @@ idx = find(hydroBodLogic==1);
 if ~isempty(idx)
     for kk = 1:length(idx)
         it = idx(kk);
-        body(it).hydroForcePre(waves(1).omega,waves(1).direction,simu.cicTime,waves(1).bem.count,simu.dt,...
+        body(it).hydroForcePre(waves(1).omega,waves(1).direction,waves(1).wavenumber,simu.cicTime,waves(1).bem.count,simu.dt,...
             simu.rho,simu.gravity,waves(1).type,waves(1).waveAmpTime,simu.stateSpace,simu.b2b);
     end; clear kk idx
 end
--- a/source/objects/bodyClass.m
+++ b/source/objects/bodyClass.m
@@ -299,10 +299,11 @@ classdef bodyClass<handle
             obj.dof = length(obj.quadDrag.drag);
         end
         
-        function hydroForcePre(obj,w,direction,cicTime,bemCount,dt,rho,g,waveType,waveAmpTime,stateSpace,B2B)
+        function hydroForcePre(obj,w,direction,wavenumber,cicTime,bemCount,dt,rho,g,waveType,waveAmpTime,stateSpace,B2B)
             % HydroForce Pre-processing calculations
             % 1. Set the linear hydrodynamic restoring coefficient, viscous drag, and linear damping matrices
             % 2. Set the wave excitation force
+            % 3. Correct the phase of excitation force caused by the body not being located at the coordinate origin
             obj.setMassMatrix(rho)
             if (obj.gbmDOF>0)
                 % obj.linearDamping = [obj.linearDamping zeros(1,obj.dof-length(obj.linearDamping))];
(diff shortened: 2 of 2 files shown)


hfchen20:
@meatballgithub You are using AQWA. The correction may not apply to other BEM codes such as NEMOH, in which you can specifically define the point location the forces are calculated at.

meatballgithub:
> You are using AQWA. The correction may not apply to other BEM codes such as NEMOH, in which you can specifically define the point location the forces are calculated at.

@hfchen20 Yes, I'm using AQWA with this issue. I'm not familiar with NEMOH or their BEM codes, and I have some questions:
As the frequency-domain data of excitation force,
$F_e(j\omega)=F_{e,\rm amp}(\omega)e^{jF_{e,\rm phase}(\omega)}=F_{e,\rm RE}(\omega)+jF_{e,\rm IM}(\omega)$,
represents the amplitude-phase characteristics subject to the waves acting on the body, the force in time-domain should be calculated based on wave elevation at the body's origin (COG), however in WEC-sim, the wave elevation used for excitation force calculation is always the wave elevation at the coordinate origin. You mentioned that it can be defined in NEMOH, does that imply that NEMOH automatically modifies $F_e(j\omega)$ to match waves at coordinate origin rather than body's COG?

hfchen20:
> You mentioned that it can be defined in NEMOH, does that imply that NEMOH automatically modifies 
 to match waves at coordinate origin rather than body's COG?

I think in NEMOH, if the reference location and body CoG are consistently defined, the excitation force's phase angles are calculated correctly. I do see a phase difference between excitation forces on two bodies spaced apart. If the reference locations for all bodies are defined as the global origin, then probably there exists a need to correct the phase angle.

I don't have experience with other BEM codes.

meatballgithub:
> > You mentioned that it can be defined in NEMOH, does that imply that NEMOH automatically modifies
> > to match waves at coordinate origin rather than body's COG?
> 
> I think in NEMOH, if the reference location and body CoG are consistently defined, the excitation force's phase angles are calculated correctly. I do see a phase difference between excitation forces on two bodies spaced apart. If the reference locations for all bodies are defined as the global origin, then probably there exists a need to correct the phase angle.
> 
> I don't have experience with other BEM codes.

I wonder if the phase differences are in WEC-Sim or NEMOH. AQWA can calculate excitation force itself without using WEC-Sim, and there will be correct phase differences for multi bodies, however, once importing the BEM data into WEC-Sim, there is no phase difference.

hfchen20:
I didn't delve into the [readBEM](https://github.com/WEC-Sim/WEC-Sim/tree/master/source/functions/BEMIO) code so do not have a definite answer. It seems to me that wec-sim only reads the force magnitude/phase data, for AQWA or NEMOH, even WAMIT, and does not do anything to correct the phase angle.

It may be worth to read carefully the[ code](https://github.com/WEC-Sim/WEC-Sim/blob/master/source/functions/BEMIO/readAQWA.m) to compare what you have in AQWA and what is read into MATLAB. If there is a need to correct the phase, it may make more sense to correct it in 'readAQWA.m' before you have a full picture about other BEM codes.

meatballgithub:
> I didn't delve into the [readBEM](https://github.com/WEC-Sim/WEC-Sim/tree/master/source/functions/BEMIO) code so do not have a definite answer. It seems to me that wec-sim only reads the force magnitude/phase data, for AQWA or NEMOH, even WAMIT, and does not do anything to correct the phase angle.
> 
> It may be worth to read carefully the[ code](https://github.com/WEC-Sim/WEC-Sim/blob/master/source/functions/BEMIO/readAQWA.m) to compare what you have in AQWA and what is read into MATLAB. If there is a need to correct the phase, it may make more sense to correct it in 'readAQWA.m' before you have a full picture about other BEM codes.

Indeed, WEC-Sim only reads the force mag/phase data for all BEM codes, my concern is that WEC-Sim calculates the force using the wave data at the coordinate origin rather than the body's origin, what I did in this commit is actually to compute the excitation force using the wave data at the body's origin. According to my understanding, the force mag/phase data in frequency-domain is defined as the mag/phase characteristics related to the wave at the body's origin. This can cause phase issues for bodies not situated at the coordinate origin. If this discrepancy doesn't occur in NEMOH, it implies that the frequency-domain data is defined inconsistently in different BEM software?. I'm confused and I hope someone can provide clarification. Moreover, it would be better if someone could share multi-body BEM data from NEMOH, WAMIT, or other sources for me to conduct a more comprehensive analysis of this phase issue.

jtgrasb:
@meatballgithub Thank you for working on this PR. Since the issue is present specifically for AQWA inputs, I created a new PR #1253 to resolve the issues in our readAQWA.m script, and I am closing this PR. I appreciate your example case here and explanation as it helped me understand the issue better. Please let us know if the new PR does not resolve your issues.

Issue #1184, opened 2023-12-15
[Bug] Errors when using WEC-Sim GUI and Run From Simulink features


Dadidal:
**Is your question request related to a problem? Please describe.**
When I am trying to run my model from Simulink GUI I get the following error:

```
Variant control 'sv_visualizationOFF' of variant block 'Ship_T14H02h16_WD00_L160W500_flex_cable4_RL/Global Reference Frame/waveVis' must be a condition expression or name of a Simulink.Variant object in the global scope workspace of the model. Ensure that the condition does not match the model name.
Component:Simulink | Category:Model error

Variant control 'sv_visualizationON' of variant block 'Ship_T14H02h16_WD00_L160W500_flex_cable4_RL/Global Reference Frame/waveVis' must be a condition expression or name of a Simulink.Variant object in the global scope workspace of the model. Ensure that the condition does not match the model name.
Component:Simulink | Category:Model error
```
Also I want to know how to setup the viscous drag matrix in hydrodynamic block mask? I don't see such input field: 
![image](https://github.com/WEC-Sim/WEC-Sim/assets/32517441/45592161-5ba2-48f1-94a8-70441825ac09)

**Describe the theory or implementation approach that you have a question**
I successfully run my model from wecSimInputFile.m by typing wecSim to command line. But in this case the opened Simulink model closes. I tried to run the model from Simulink as described in [](https://wec-sim.github.io/WEC-Sim/master/developer/library.html#run-from-simulink) but got the abovementioned error. I tried both Parameter Input Methods: "Input file" and "Custom parameters" and got this errors in every case. Also when I am trying to "Load parameters from Input File" I receive another error: 
![image](https://github.com/WEC-Sim/WEC-Sim/assets/32517441/468df65c-da74-4164-b37a-6517d98d62cb)

**Personal project context**
My end goal is to use the Reinforcement Learning Toolbox to create an RL Agent, which will drive some parts of my model. The training of RL agent includes a lot of simulations, which are carried out by Simulink one by one in automatic mode. In order to do this I want to run my Simulink model without typing wecSim to command line. I guess if I can run WEC-Sim from Simulink I can reach my goal.

**Describe the type of conclusion or resolution you are looking for**
I want to know how to properly run Simulink model with WEC-Sim blocks without typing wecSim to command line. Is it possible to run such model during training the RL agent?

akeeste:
Hi @Dadidal 

[...]

PR #1195, opened 2024-01-03
Bug fixes for WEC-Sim GUI and Run From Simulink features

akeeste:
This PR fixes two bugs brought up in #1184:
- [x] ``loadInputFileCallback.m`` broke when the top level of a model had non-WEC-Sim library blocks
- [x] ``writeInputFromBlocks`` was not writing strings to the input file robustly. The command to write a new input file would break when paths contained "\U" or other special character combos, which was typical when a user selected a file of some kind (e.g. body(1).geometryFile = "C:\Users\name\...\geometry\file.stl")

### DIFF
--- a/source/functions/initializeWecSim.m
+++ b/source/functions/initializeWecSim.m
@@ -62,8 +62,11 @@ if exist('runWecSimCML','var') && runWecSimCML==1
     run('wecSimInputFile');
 else
     % Get global reference frame parameters
-    values = get_param([bdroot,'/Global Reference Frame'],'MaskValues');    % Cell array containing all Masked Parameter values
-    names = get_param([bdroot,'/Global Reference Frame'],'MaskNames');      % Cell array containing all Masked Parameter names
+    blocks = find_system(bdroot,'Type','Block');
+    mask = contains(blocks,'Global Reference Frame');
+    referenceFramePath = blocks{mask};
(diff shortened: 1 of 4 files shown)


akeeste:
``writeInputFromBlocks`` has many lines changed because I updated how the ``writeLineFromVar`` function works. This changed each call to ``writeLineFromVar`` within ``writeInputFromBlocks``. Each line of the input file must be written using the ``'%s'`` notation in fprintf so that strings are written correctly and the line break can still be added.

akeeste:
@kmruehl There was another problem identified with my solution in #1184 . I'll work to make this bugfix more robust this week.

akeeste:
@kmruehl This is now ready for review

kmruehl:
@MShabara  and @jleonqu  this is ready for review

jleonqu:
@akeeste I was able to test this PR and it works well on my end. I tested this PR in Matlab 2022b and Matlab 2023b.
I tested this PR in a similar way than you in issue #1184

jleonqu (approved):
@akeeste  I was able to test this PR and it works well on my end. I tested this PR in Matlab 2022b and Matlab 2023b.
I tested this PR in a similar way than you in issue #1184

MShabara:
@akeeste I tested the modifications using MATLAB 2023b and the model works perfectly, 
good job!

PR #1204, opened 2024-01-17
Nemoh v3.0.2 Examples

MShabara:
1. Split NEMOH examples into v2.3 and v3.0.2.
   2. Added two examples for v3.0.2: Cylinder and Sphere.
   3. Redirected the **testBEMIO.m** file to the new Sphere example of v3.0.2 instead of v2.3.


dav-og:
Thanks for updating these, @MShabara !

The main solver seems to be running fine :)

I couldn't get the QTFSolver to run, but I'm not sure if thats actually available on this example case?

dav-og (approved):
Tests all passing, Nemoh v3 examples working, great addition 👍

PR #1207, opened 2024-01-22
Bugfix for plotBEMIO bodies

akeeste:
The BEMIO plotting functions were not handling DOFs correctly when not all bodies were being input. 
e.g. ``hydro.plotBodies = 2``, then the first body was still being plotted

This PR fixes the logic in each plotting function to accurately plot the hydrodata of bodies called out in hydro.plotBodies, regardless of body order. I also updated some of the looping variables to be more meaningful so that the functions are a bit easier to follow. e.g. ``iB`` for body index, ``iDof`` for DOF index, etc

### DIFF
--- a/source/functions/BEMIO/addDefaultPlotVars.m
+++ b/source/functions/BEMIO/addDefaultPlotVars.m
@@ -18,9 +18,12 @@ function hydro = addDefaultPlotVars(hydro)
 %         Hydro data with default plotting variables.
 % 
 
-hydro.plotDofs = [1,1; 3,3; 5,5];
-hydro.plotBodies = 1:hydro.Nb;
-hydro.plotDirections = 1;
+[a,b] = size(hydro);  % Last data set in
+F = b;
+
+hydro(F).plotDofs = [1,1; 3,3; 5,5];
+hydro(F).plotBodies = 1:hydro(F).Nb;
+hydro(F).plotDirections = 1;
 
 end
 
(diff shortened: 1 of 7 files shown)


akeeste:
Updated my branch to be based on master

jtgrasb:
Hi Adam, thanks for catching this bug and fixing it. Your fix looks good and works well, and I appreciate the updates to make it more readable as well. 

One question I have is for the `addDefaultPlotVars.m`, is there a point to just adding the plot variables to the last dataset? I don't see an instance where the user would pass in multiple hydro structs, because the readX.m scripts already apply the default plotting variables. Your implementation here works fine with just 1 so no issues with it, I'm just curious.

akeeste:
Hi @jtgrasb Thanks for the review! On ``addDefaultPlotVars`` -- This will break when hydro is an array of structures. For example if I wanted to read multiple output files to compare or combine them:
```
hydro = readCapytaine(hydro, 'file1.nc', ...)
hydro = readCapytaine(hydro, 'file2.nc', ...)
hydro = combineBEM(hydro)
plotBEMIO(hydro)
```
Then the second readCapytaine call will fail because ``addDefaultPlotVars`` tries to add ``plotDofs``, etc to an array instead of a structure. I was doing this because I was processing a lot of hydro files at once and comparing them.

I think we previously had a set-up like this in compareBEMIO, but it looks like it was changed to save computational time: https://github.com/WEC-Sim/WEC-Sim/blob/0af8e28352f0ce2cd376b98fd5d5109c591a1541/examples/BEMIO/compareBEMIO.m#L22-L25

jtgrasb:
@akeeste Thanks for the explanation, makes sense. Everything looks good then, I will merge this today.

PR #1134, opened 2023-10-02
Update to MoorDyn v2

RyanDavies19:
This updated the coupling of WECSim and MoorDyn to MoorDyn-C v2 using the MoorDyn v1 API. Additional updates to the MoorDyn libraries and input files will be added separately. 

The only new feature added is the ability to specify the path and file name of the MoorDyn input file. This will be the same path where outputs are written. The figures below show that this update performs as well as the MoorDyn v1 coupling. This update opens the door to future utilization of the MoorDyn v2 features. 

Fairten_comp
float_heave_response
float_surge_response
spar_heave_response
spar_surge_response

PR #1212, opened 2024-01-30
Enable WEC-Sim to use MoorDyn v2 capabilities

jtgrasb:
This PR expands upon #1134 to enable WEC-Sim's use of MoorDyn's new capabilities, notably the ability to handle mooring lines between multiple dynamic bodies. This PR includes the following changes:
- Mooring Library 
  - Changed the MoorDyn block to a `MoorDynConnection` block, which has two connections to detect the relative motion between two bodies/frames, send the displacement and velocity to the `MoorDynCaller` block, and receive back the forces calculates by MoorDyn
  - Added a `MoorDynCaller` block that programmatically receives the displacement and velocity from the `MoorDynConnection` block(s) and calls MoorDyn to calculate the forces. 
  - There can only be one `MoorDynCaller` block, but it allows for multiple connection blocks to be used.
- Updated the mooring class to initialize and close MoorDyn instead of doing it within the Simulink block. 
- Updated documentation to reflect the new process for using MoorDyn.
  - I added a note to indicate that the MoorDyn webinar is out of date, but need to consider updating this in the future.

A few other notes about the implementation:
- MoorDyn handles the kinematics between the body's center of gravity and the mooring connection location (based on lines.txt). Thus, as I understand it, the mooring(i).location should always be `[0, 0, 0]` so that the motion passed into MoorDyn is the exact motion of the body.
- The initial displacement factors into the orientation of the mooring connection and is added to the displacement before sending to MoorDyn. Any initial body displacement should also be applied to the moordyn connection.
- Updates to the MoorDyn examples in WEC-Sim Applications are in [PR#57](https://github.com/WEC-Sim/WEC-Sim_Applications/pull/57)


docs/_include/mooring.rst
```diff
@@ -92,17 +94,27 @@ The lookup table dataset should contain one structure that contains fields for e
...
-defined in MoorDyn (``mooring(1).moorDynLines = <Number of mooring lines>``) 
-
-A mooring folder that includes a moorDyn input file (``lines.txt``) is required 
-in the simulation folder. 
+file (wecSimInputFile.m), followed by setting :code:`mooring(i).moorDyn = 1` to 
+initialize a MoorDyn connection. Each MoorDyn connection can consist of multiple 
+lines and each line may have multiple nodes. The number of MoorDyn lines and nodes in 
+each line should be defined as (``mooring(i).moorDynLines = <Number of mooring lines>``) 
+and (``mooring(i).moorDynNodes(iLine) = <Number of mooring nodes in line>``), 
+respectively.
```
RyanDavies19, line 104: @jtgrasb Does this need to match what is defined in the MD input file? If so maybe we note that

docs/_include/mooring.rst
```diff
@@ -92,17 +94,27 @@ The lookup table dataset should contain one structure that contains fields for e
...
+lines and each line may have multiple nodes. The number of MoorDyn lines and nodes in 
+each line should be defined as (``mooring(i).moorDynLines = <Number of mooring lines>``) 
+and (``mooring(i).moorDynNodes(iLine) = <Number of mooring nodes in line>``), 
+respectively. 
+
+A mooring folder that includes a MoorDyn input file (``lines.txt``) is required 
+in the simulation folder. The body and corresponding mooring attachment points are 
+defined in the MoorDyn input file. MoorDyn handles the kinematic transform to 
+convert the forces from the attachment points to the 6 degree of freedom force 
+acting on the body's center of gravity. Thus, the mooring location should generally
```
RyanDavies19, line 110: @jtgrasb If you are referring to a MD body here, you should note that the forces act on the body's location rather than center of gravity. The body is a bit like a point in space which when things are attached the forces and motions are controlled rigidly through that point

docs/_include/mooring.rst
```diff
@@ -135,8 +148,9 @@ Applications <https://github.com/WEC-Sim/WEC-Sim_Applications>`_ repository.
 
 * **MoorDyn Input File**: A mooring folder that includes a moorDyn input file 
   (``lines.txt``) is created. The moorDyn input file (``lines.txt``) is shown 
-  in the figure below. More details on how to setup the MooDyn input file are 
-  described in the MoorDyn User Guide :cite:`Hall2015MoorDynGuide`.
+  in the figure below. More details on how to set up the MooDyn input file are
```
RyanDavies19, line 151: @jtgrasb Does the current configuration always require a coupled body on the MD side? If so I can make that change in the MD documentation, or we could put it on this end

docs/theory/theory.rst
```diff
@@ -782,14 +782,14 @@ MoorDyn discretizes each mooring line in a mooring system into evenly-sized
...
 gravitational and buoyancy forces, hydrodynamic loads, and reactions from 
-contact with the seabed. Hydrodynamic drag and added mass are calculated based 
-on Morison's equation. A mooring line's axial stiffness is modeled by applying 
-a linear stiffness to each line segment in tension only. A damping term is also 
-applied in each segment to dampen non-physical resonance caused by the 
-lumped-mass discretization. Bending and torsional stiffnesses are neglected. 
-Bottom contact is represented by vertical stiffness and damping forces applied 
-at the nodes when a node is located below the seabed. 
-:cite:`Hall2015ValidationData`. 
+contact with the seabed or other bodies. Hydrodynamic drag and added mass
```
RyanDavies19, line 785: @jtgrasb Have you seen this contact with other bodies effect? I don't think that there are any effects between objects in terms of contact aside from the seafloor.

docs/_include/mooring.rst
jtgrasb: Yes this does - I added this to the docs.

docs/_include/mooring.rst
jtgrasb: Thanks for clarifying. If I'm understanding correctly, you're saying that the forces are converted to the Body location specified in lines.txt which is not necessarily the CG? WEC-Sim applies the forces to the body at the center of gravity, so the way I understand it the location specified in lines.txt should be consistent with the initial location of the center of gravity.

docs/_include/mooring.rst
jtgrasb: Technically, I believe another body could be added to lines.txt and simulated in MoorDyn. It would be a weird way to do it but possible. As for the WEC-Sim body, it should always be specified as Coupled, so I have added this to the docs here.

docs/_include/mooring.rst
RyanDavies19: You are correct. If you want the returned forces to be the forces on the CG, then the MD body needs to be initialized at the location of the CG of the WECSim Body.

dav-og:
minor thing: 

image

MoorDynCaller layout seems to be messed up?

jtgrasb:
The MoorDynCaller block programatically adds and removes the From and Goto blocks according to the number of MoorDyn connections specified. I tried to [arrange the model automatically](https://www.mathworks.com/help/simulink/slref/simulink.blockdiagram.arrangesystem.html) but it seemed to give errors when doing it from the mask. I could also change the location of these placements, but it does not affect the functionality so I didn't feel it was worth it.

jtgrasb:
Hey @dav-og, do you have any other comments or suggestions? Or is it ready to be merged?

dav-og:
Hi @jtgrasb it seems to be running fine, I just have a couple of points:

1. Do you have a comparison between v2 and v1 to verify that the results aren't being affected?
2. Would this update be backwards-compatible with MoorDyn v1?

RyanDavies19:
> Hi @jtgrasb it seems to be running fine, I just have a couple of points:
> 
> 1. Do you have a comparison between v2 and v1 to verify that the results aren't being affected?
> 2. Would this update be backwards-compatible with MoorDyn v1?

@dav-og For your second question, no this is not backwards compatible with MD V1. Function names have changes, as well as input file structure, and I/O between MoorDyn and the calling script (WECSim in this case). However MD V1 is no longer in development and V2 has improved a lot of the behavior of the code, and from the MoorDyn side we encourage all users to use the most up-to-date version of v2.

dav-og:
@RyanDavies19 cheers, I assumed so from MoorDyn side, but wasn't sure if on the WEC-Sim side we'd be able to do `if v1 -> use old approach, elif v2 -> use new approach` ... might make life easier for someone with a complex mooring model already implemented with v1, but as you say - people should really just be upgrading to v2 👍

RyanDavies19:
@dav-og Thats a good point, hopefully with the MD docs site it should be relatively simple (but potentially tedious) to convert from a v1 input file to a v2 input file. The format of both the v1 and the v2 input files are in there.

RyanDavies19:
Noting that @jtgrasb and I discussed the pending reviews above and they can be closed

jtgrasb:
@dav-og Thanks for the review. 
1. The zip file for comparison can be found on [PR#57](https://github.com/WEC-Sim/WEC-Sim_Applications/pull/57) of the Applications Repo. Plotting the results for comparison yields the following figure. The notable difference is due to the fact that MD v2 has a more effective initialization. Both results converge to very similar dynamics after about 20 seconds, with a slight offset due to the updated moordyn input file (lines.txt).
md_compare

2. The updates are not backwards compatible, as Ryan said. The main reason for this is because the MoorDyn function names have changed. For example, when WEC-Sim previously called MoorDyn, it would use LinesInit() but now must use MoorDynInit(). As far as I can tell, there is no way to make the changes backwards compatible while still updating to MD v2.

dav-og:
Looks like there is a persistent offset in the v2 results, but it probably makes sense to trust the v2 results more.

@RyanDavies19 is there a verification reference for v2 that we can point to in our docs?

RyanDavies19:
@dav-og for now point to here: https://moordyn.readthedocs.io/en/latest/theory.html#id2

I am working on adding more v2 theory papers to that site. In general, for the past couple years work using MoorDyn-C and/or MoorDyn-F in OpenFAST have been using MoorDyn v2.

jtgrasb:
Thanks @RyanDavies19 and @dav-og. The documentation added here points to the updated MoorDyn Documentation. 
@dav-og Do you have any further review comments or is this PR and the [WEC-Sim Applications PR](https://github.com/WEC-Sim/WEC-Sim_Applications/pull/57) ready for merge?

dav-og:
ready to merge 👍

Issue #1217, opened 2024-02-05
Calculation of the nonlinear FK force


guangyaow:
I am unclear on how the nonlinear FK force is calculated for a 3D irregular wave field. 

I can see that in waveElevation.m, the direction for each wave mode (the varaibale in the script is 'direction')  are the x and y coordinates (center(:, 1) and center(:, 2)) considered when calculating the wave elevation for each spatial in a 2D (x-y) space. 

Then the calculated elevation (elv) is used as input of nonFKForce.m. While in the function pDis (within nonFKForce.m), the wave direction and the y coordinate for surface (center(:,2)) are not considered to calculate pressure (only center(:,1) is used in the calculation), as shown below:

function f=pDis(center,elv,AH,w,dw,wDepth,deepWater,t,k,phaseRand,typeNum,rho,g)
% Function to calculate pressure distribution
f = zeros(length(center(:,3)),1);
z=zeros(length(center(:,1)),1);
if typeNum <10
elseif typeNum <20
    f = rho.*g.*AH(1).*cos(k(1).*center(:,1)-w(1)*t);
    if deepWater == 0
        z=(center(:,3)-elv).*wDepth./(wDepth+elv);
        f = f.*(cosh(k(1).*(z+wDepth))./cosh(k(1)*wDepth));
    else
        z=(center(:,3)-elv);
        f = f.*exp(k(1).*z);
    end
elseif typeNum <30
    for i=1:length(AH)
        if deepWater == 0 && wDepth <= 0.5*pi/k(i)
            z=(center(:,3)-elv).*wDepth./(wDepth+elv);
            f_tmp = rho.*g.*sqrt(AH(i)*dw(i)).*cos(k(i).*center(:,1)-w(i)*t-phaseRand(i));
            f = f + f_tmp.*(cosh(k(i).*(z+wDepth))./cosh(k(i).*wDepth));
        else
            z=(center(:,3)-elv);
            f_tmp = rho.*g.*sqrt(AH(i)*dw(i)).*cos(k(i).*center(:,1)-w(i)*t-phaseRand(i));
            f = f + f_tmp.*exp(k(i).*z);
        end
    end
end
f(z>0)=0;
end

May I ask why the pressure is calculated without considering wave direction and y coordinate? Your help and support is much apprecaited.

Best,
Guangyao

dforbush2:
Hi @guangyaow 

It is possible that #886 is also relevant here.
[...]

PR #1220, opened 2024-02-08
fix to #1217, nonlinFK and body block changed

dforbush2:
Credit to @guangyaow on catching the error and proposing the fix. 
This fix makes nonlinear froude krylov work for **plane waves** with **non-zero** directions. If the waves are not plane waves (e.g, there are a sum of wave trains from different directions) this will not work correctly. Capabilities will be expanded in a future PR.

Hotfix merge into master is intentional


dforbush2:
This is ready to go. To avoid confusion for the reviewer note that:
1) elv input is from waveElevation and is a vector that is equal in length to the center(:,3) from the STL file. This updates every time step. This elv already accounts for a non-zero wave direction to solve for wave elevation at [center(:,1), center(:,2)]. So Z= center(:,3)- elv accurately reflects the difference in height between the body panel in Z and the free surface at that same (x,y) point. 
2) For regular waves it is possible to use elv as the input since it is monochromatic. For irregular waves, the portion of the hydrostatic pressure equation for elevation must be recalculated because the check for deep water conditions happens every frequency. Recall that elv is elevation due to all frequency components at the points (center(:,1), center(:,2)): hence although it is a bit redundant to repeat this calculation here it is necessary. 
3) The whole nonFKForce only works with planar waves. UserDefinedEta and the multiple wave train feature added recently is the only way currently that a user could input non-planar waves. Please update documentation related to this latter feature to indicate that non-linear forcing will not work

guangyaow:
Hello, thanks so much for your response and clarification. If I want to use
UserDefinedEta and the multiple wave train feature (without considering the
nonlinear FK force), I need to define the waveclass as " waves = waveClass(
'elevationImport');", is my understanding correct? In addition, may I ask
what's the standard format for waves.elevationFile ='<elevationFile>.mat'
(how the elevation, which is a function of (x,y) and time. should be
stored)? Thanks.

Best,
Guangyao

salhus:
Hi @guangyaow,

Yes, you are right.
For a template for the .mat file please refer to the examples and you would find .mat files for an elevation import, and a spectra import.

Cheers,
sal

akeeste:
@salhus @dforbush2 Was there any change to the mask in this PR? I'm having issues merging updates into my variableHydro branch that I believe are ultimately coming from this PR into master.

Also *please* only squash and merge PRs into master/main/dev. The commit history on main is pretty messy now.

dforbush2:
To my knowledge there is not a mask change but there is a change to the noted library block to pull in wave direction to the function. This MAY have resulted in a subsequent mask change when multiple wave trains I.e. multiple possible wave directions were added, but I don’t think that had to happen here.

akeeste:
Thanks @dforbush2. I'll keep digging around for the issue on #1248

akeeste:
Made some progress--my issue on #1248 was coming from a mask and Run From Simulink + WEC-Sim GUI which is hard to track, but is unrelated to anything done here.

Issue #1218, opened 2024-02-07
[WEC-Sim Applications] Power calculation in PTO-Sim


fanofdw:
**Note: italicized text below is include as an example and should be updated before submission. If you feel any section is not applicable to your request, please replace with 'N/A' rather than delete a section.**

**Cite the [WEC-Sim applications](https://github.com/WEC-Sim/WEC-Sim_Applications) case in question**
https://github.com/WEC-Sim/WEC-Sim_Applications/tree/master/PTO-Sim/RM3/RM3_cHydraulic_PTO

**Describe the bug or issue**
Hello, WEC-Sim team. 

I have run the PTO-Sim/RM3/RM3_cHydraulic_PTO example code, and I found that the power obtained by voltage * current of the Electric Generator Equivalent Circuit is greater than the power obtained by torque * shaftspeed of the hydraulic motor.

Question 1: Is it reasonable that the electrical power is bigger than the motor power? The trend of the two power curves seems similar.

Question 2: The power obtained by voltage * current of the Electric Generator Equivalent Circuit is the final power generated by RM3 WEC, right?

Thank you for your time. 

**To Reproduce**
1. Add the code below into userDefinedFunctions.m.
```matlab
figure();
plot(output.ptoSim(6).time,-output.ptoSim(6).voltage.*output.ptoSim(6).current/1000)    % U*I is expected to be negative for generating electricity, and the "-" here is just for a better view in comparison.
set(findall(gcf,'type','axes'),'fontsize',16)
xlabel('Time (s)')
ylabel('Power (kW)')
title('Power')
grid on

figure();
plot(output.ptoSim(5).time,output.ptoSim(5).torque.*output.ptoSim(5).shaftSpeed*2*3.14/60/1000)
set(findall(gcf,'type','axes'),'fontsize',16)
xlabel('Time (s)')
ylabel('Power(kW)')
title('motor Power')
grid on
``` 

2. Modify the simulation duration time:
```matlab
simu.endTime= 2000;  
``` 

3. Run wecSim.
4. Check the output figures.

**Expected behavior**
According to the energy transfer path, the final power should be lower than the motor power.

**Screenshots**
![power comparison](https://github.com/WEC-Sim/WEC-Sim/assets/143822308/74df37ea-e097-405f-a1d3-f367585b330d)

**Workstation Information (this section MUST BE completed):**
 - OS: Windows 10
[...]

PR #1221, opened 2024-02-08
Fixing an inconsistent sign in the 'Electric Generator Equivalent Circuit' block in PTO-Sim

jleonqu:
An issue regarding the power calculated with the RM3 example using PTO-Sim was reported in #1218 

There are two inconsistent signs in the 'Electric Generator Equivalent Circuit'. The first inconsistent sign is in the voltage equation of the equivalent circuit. The correct equation for the voltage in the circuit is:

$` e - L_a \frac{di}{dt} - R_ai-R_{load}i = 0 `$
$` e = k_e \omega  `$

For the dynamic equation in the shaft, the torque input should be considered as a positive input, which is accelerating the motor shaft, and the electric torque should be negative because it is used to generate the electric power and it is decelerating the electric motor shaft. The correct dynamic equation is:

$` J_{EM} \ddot{\theta} = T_L - b_{shaft}\dot{\theta}-k_t i `$

This PR fix the signs in the 'Electric Generator Equivalent Circuit' block.


dforbush2:
The signs seem to agree now and the dynamics make sense. One problem here is that it is assumed that Rload can go negative: this amounts to current injection (i.e., reactive power), and can be a bit misleading represented in this way. Recommend eventually revising to a command current at the inverter. 

Ready to merge.

kmruehl:
@jleonqu should this go in dev or master?

kmruehl:
@dforbush2  we usually recommend bugfix PRs into main (formerly master), and then we pull main into dev. That ensures that all bug fixes get propagated into both branches. @jleonqu can you update this PR to go into main, and then once merged we'll pull it into dev.

jleonqu:
Done. I've changed the PR to 'main' and I've solved the conflicts. It is ready to merge

kmruehl:
Thanks @jleonqu

akeeste:
@jleonqu when this PR was changed from dev into main, did you rebase it to be based on the main branch? It looks like there are 185 files changed, and that all of dev was pulled into main, not just this fix

akeeste:
This PR was force overwritten to remove the commit that included items from dev. See new PR

PR #1226, opened 2024-02-21
Add new examples for updated version of NEMOH (NEMOHv3.0.2)

ashleynchong:
Updated example files for NEMOHv3.0.2


dav-og:
It takes a long time to run >200 freqs for 7 cases. I propose a test option, where we just run ~20 freqs for each case and check the results?

I can work on this - just want to check if this approach makes sense to others first.

dav-og:
@ashleynchong @MShabara 

I just pushed a script to more easily edit the number of frequencies and run Nemoh (either v2 or v3) and a plotting script to compare results - please can you test these out for yourself and check that they are ok?

I did see some differences in the results of each code but I suppose that is to be expected...

Some of the v2 examples had a range of wave headings - I noticed the updated v3 examples lost this and only have one heading.

Issue #1215, opened 2024-02-01
[Developer Issue] Lets rename WEC-Sim 'master' branch to 'main'


dav-og:
**Project**
WEC-Sim

**Describe the issue**
In alignment with our ongoing efforts to foster diversity, equity, inclusion, and accessibility (DEIA) in our community, I propose that we rename our default branch from 'master' to 'main'. This change is a step towards removing language that may be considered exclusive or insensitive, and it reflects a broader shift in the tech community towards more inclusive terminology.

Several leading organizations and many open-source projects have already embraced this change, recognizing the importance of language in creating an inclusive environment. By renaming our default branch, we not only align ourselves with these best practices but also send a clear message about our values and commitment to DEIA.

For more information see: https://github.com/github/renaming

Cheers,
Dave

@akeeste:

I'm reopening this issue because in reviewing a PR I noticed there's a few outstanding things to complete:
- [x] CI tests are not running on the main branch, e.g. #1221 did not have any checks run when it went into main. All GitHub workflows (3) need to be updated to reference the 'main' branch instead of master #1235
- [x] Update WEC-Sim Applications repo branch name
- [x] Update WEC-Sim/MoorDyn branch name
- [x] Merge PR updating the WEC-Sim Applications CI workflows WEC-Sim/WEC-Sim_Applications#59
- [x] Update all references to the github repo trees, e.g. "https://github.com/WEC-Sim/WEC-Sim/tree/master" to "https://github.com/WEC-Sim/WEC-Sim/tree/main" (for applications repo too). Fortunately if unchanged these URLs DO redirect to the main branch automatically, but we should still update them  #1235
- [x] Update all references to the built documentation. This is a MAJOR change and will break any URL to WEC-Sim documentation that we've distributed in the past. Ideally our old URLs could redirect but I don't know how to do that currently. - MT: In #1235, a redirect was added from `master/index.html` to `main/index.html`. Note, this won't catch any of the old paths below `master/index.html`, they will just 404.
- [x] update organization setting to default future repos to main

PR is forthcoming to do 1, 4 above ✔️ (See #1235, #1240, #1241 and https://github.com/WEC-Sim/WEC-Sim_Applications/pull/59)

kmruehl:
@dav-og I agree, this is something I've wanted to do for a while. Making this change will have some downstream impacts that we need to address, so I'd like to make sure the rest of the team is aware prior to making this change. Thanks for the suggestion!

[...]

PR #1235, opened 2024-02-28
Update master to main

akeeste:
This PR addresses a few outstanding names changes described in #1215 on updating master to main.

It updates every reference to the master branch in our repo, except in ``release_notes.rst`` since the master branch was valid at the time of those releases.

Most of the lines changed come from correcting an old merge conflict in a LIS file


akeeste:
@H0R5E we renamed our "master" branch to "main". Can you review this PR and 
- see if there's anything else needed to get Actions up and running on the main branch again
- I believe these changes will start building our documentation at https://wec-sim.github.io/WEC-Sim/main/index.html instead of https://wec-sim.github.io/WEC-Sim/main/index.html. Can you confirm? And if so, is there any way we can get our old docs to redirect to the new URL?

H0R5E:
@akeeste, as I said in Applications, you might want to consider renaming `run-tests-master.yml` to `run-tests-main.yml` and then updating the README.

To redirect the docs, you're going to have to change [this line](https://github.com/WEC-Sim/WEC-Sim/blob/5e65f1f54c05a9656d1e5b15a1ca116833d1062c/docs/_assets/gh-pages-redirect.html#L6) so that the `url` field is now `url=./main/index.html"`.

H0R5E (changes requested):
See previous comments.

H0R5E:
@akeeste you will also need to update [this line](https://github.com/WEC-Sim/WEC-Sim/blob/5e65f1f54c05a9656d1e5b15a1ca116833d1062c/docs/conf.py#L76) to reference main rather than master. I only just remembered, sorry.

H0R5E:
@akeeste, I've added https://github.com/akeeste/WEC-Sim/pull/13 which makes the change I suggested above and also adds a second redirect from `/master/index.html` to `/main/index.html`. This should mean that any only links that went straight to `master` instead of the root index will be redirected.

H0R5E:
One last comment, Adam, is that you will need to sync the `dev` branch after merging this, otherwise it's docs builds won't create the `main` branch version.

akeeste:
Thanks for your help @H0R5E! Once this is merged, I'll pull this and a few other bugfixes from master into dev.

Since the docs will now redirect, I finished changing every reference to the old docs here and on WEC-Sim/WEC-Sim_Applications#59. The test to build the current branch docs is failing because it's trying to reference the main branch on the Applications repo, which does not exist yet. It will pass once the WEC-Sim_Applications branch name is renamed.

Unfortunately from the labs, only @kmruehl has access to the Applications repo to rename master to main, and is unable to for two more weeks. @H0R5E do you have access [here](https://github.com/WEC-Sim/WEC-Sim_Applications/settings) to do this?

H0R5E:
@akeeste, sorry for the delay in replying. I'll add a main branch to the repo now and we can check it's working OK, prior to changing the default.

akeeste:
I've only found one more reference to the master branch documentation. It's a comment within the WECSim_Lib_PTOSim.slx but I can't find it for the life of me. I suggest we let that go for now since it should redirect anyways, and it's very unlikely to be used or found by anyone

H0R5E (approved):

H0R5E:
> I suggest we let that go for now since it should redirect anyways, and it's very unlikely to be used or found by anyone

@akeeste, agreed. We can tidy up any little things in the future, and this is important to get merged ASAP. Thanks for the great work.

H0R5E:
@akeeste, I've applied this to the `dev` branch in #1240, FYI.

kmruehl:
@akeeste and @H0R5E thank you both for resolving these issues while I was OOO

Issue #58, opened 2015-11-03
Mooring questions


cmichelen:
@ahagmuller 
```
Hi Carlos , 

I haven't looked yet but is MoorDyn also open source? 

I had posted some questions regarding Mooring damping and stiffness and what I thought was the
 correct application of the matrix. 

I think it would be really helpful to the user to see a drawing of how the matrix is used base on the
 geometry of the device and how the equations are applied. I am still a little fuzzy on some I this 
myself. 

That being said, we are assuming that if we use damping on the spherical joint in the example we 
posted that this can be resolved to forces in x,y and z based on the connection points of the morning 
lines. Then, the rotational damping*rotational velocity in that axis is equal to torque, and that torque 
divided by the effective torque arm length to the mooring connection will be the magnitude of the 
catenary mooring force which can be resolved into its components. Does this sound correct? 

Thanks , 

Alex
```

cmichelen:
Yes MoorDyn is open source. Here is the site: http://www.matt-hall.ca/software/moordyn/

For WEC-Sim though, we distribute a compiled MoorDyn library (in ``source/functions/moorDyn``).

We are now also giving the option of using floating joints motion for the mooring matrices. This is specified in a similar way to as with the body as:
```matlab
mooring                 = struct('c',          zeros(6,6), ...          % Floating constraints only. Mooring damping, 6 x 6 matrix. 
                                         'k',          zeros(6,6), ...          % Floating constraints only. Mooring stiffness, 6 x 6 matrix.
                                         'preTension', [0 0 0 0 0 0])           % Floating constraints only. Mooring preTension, Vector length 6.
        initDisp                = 0                                             % Pitch constraint only. Initial angular displacement of joint [radians]. Use with caution: frame must be rotated back before attachement to next body.
    end
```

For backwards compatibility we are still going to leave the option of specifying the matrix on the body itself (eventually we will remove it though, we want mooring outside the body block).

The way the mooring force using the matrices works is:
Force = velocity(6x1) * c_matrix(6x6) + position(6x1) * k_matrix(6x6) - preTension(6x1)
You would have to be able to describe your mooring forces in such a way.

With this new setup the mooring force is applied at the location of your floating constraint, ``constraint(i).loc``, rather than the body's cg. 

By default (``simu.moorDyn=0``) the matrices formulation is used for mooring, with default values equal to zero. To use MoorDyn set ``simu.moorDyn=1``.

cmichelen:
[...]

PR #1236, opened 2024-03-04
Fix sign bug in the Generator Equivalent Circuit Block in PTO-Sim - Rebase PR to main

jleonqu:
This PR is based on PR #1221. This new PR is rebase on the 'main' branch of WEC-Sim


H0R5E:
@jleonqu, you'll need to push an update to trigger the CI. It should work now.

dforbush2:
I have reviewed and tested on 2022b and 2023b windows, tested using applications case PR #58. This is ready to merge. 
However, I would suggest that we consider the documentation's level of detail and/or advise users to look closely at the implementation of these PTO blocks prior to use. For instance, motor back EMF constant is implemented as equal to motor torque constant: this is true only under very specific conventions that may not align with device developer specs.

Issue #1201, opened 2024-01-15
Changing operational depth of a device


TheWaveMaster:
_Hi all,
First of all, I want to thank the team behind wecSim for your amazing work! 
I work for a wave energy developer, and just finalized a wecSim model for our device which we will heavily rely on in the final optimizations before we build and deploy soon!! 

The amount of features that I was able to add because of wecSim are ridiculous, and the model is showing excellent matching against wave tank data!  Big thanks and please keep up the good work!_ 

Now, the theory implementation I want to discuss and would like to hear a few opinions. 
I want to simulate a device which changes its operational submergence depth in sim (gives you a clue to which company I work for). I was wondering if you guys have any idea regarding this. Obviously, it's not possible to change the hydrodatabase in sim (I can't even begin to imagine what will happen to the math of the model lol). Perhaps with the nonlinear hydros, the issue will be entities like the added mass and the radiation damping as basically the device goes from operating submerged to floating at the surface. Wondering what you guys think.

salhus:
Hi @TheWaveMaster (cool name lol),

I think what you are describing is the need for the capability to swap out BEM coefficients for large amplitude motions. I would recommend you put in a feature request, and we have a formal format to follow when doing so...  Short answer, it is possible...  but may require some development.  
If you are willing, you could put in the PR !
Basically, you will have to,
0. Load multiple hydro objects and assign them different names, e.g. hydro_0 = load hydro for SWL, hydro_1 = load hydro for 1m draught and so on..
1. Introduce a conditional if statement that checks for the displacements in the main model,
2. Make a mask over the body blocks,
3. In the mask's code initialization, have it write the hydro variable based on the appropriate hydro object from the conditional if statement...

I might try to implement this but can't guarantee a timeline..  Hope that helps ..

Cheers,
sal

Salman Husain, MSc, PhD,
National Renewable Energy Laboratory.

dforbush2:
This sounds like something similar to our passive yaw implementation, where we look up the appropriate set of BEM coefficients from the present state of the device. In that case it is its yaw orientation, but we could do something similar with a lookup table based on z-position relative to the free surface.

TheWaveMaster:
> Hi @TheWaveMaster (cool name lol),
> 
[...]

PR #1248, opened 2024-04-10
Variable hydrodynamics

akeeste:
This PR addresses issue #1201 and the discussion in https://github.com/WEC-Sim/WEC-Sim_Team/discussions/89 by allowing users to define multiple hydrodynamic datasets and switch between them during simulation as the operational state of a device changes. 

Major changes in this PR:
- WECSim_Library_BodyElements.slx:
   - adds ``From`` flags that hold an index (``body1_hydroForceIndex``, etc) which the user can define in their switching logic.
   - creates a separate Variable Hydrodynamics block which controls importing body.hydroForce into Simulink. This is a variant subsystem so that in the case of 1 h5File (as was previously supported), the library behavior is simplified.
- bodyClass.m
   - adds an index for the hydroForce dataset at equilibrium. ``body.hydroForceIndexInitial``
   - Makes every instance of body.h5File a cell array of strings
   - Makes every instance of body.hydroData an array of structures, with each structure retaining the previous format
   - Makes every instance of body.hydroForce a structure containing the previous hydroForce structure format. This nested structure is required to import the entire dataset as a bus in Simulink.
- initializeWecSim.m
   - whenever functions are called on body.hydroData, loop through the entire array of hydroData structs
   - create buses for each hydro body
   - create variable hydro variant subsystems for each hydro body
- struct2bus.m
   - modified from answer on [this thread](https://www.mathworks.com/matlabcentral/answers/98993-how-do-i-create-a-bus-object-from-a-struct-in-my-workspace-using-simulink-7-4-r2009b)
   - creates a bus from nested structures so that body.hydroForce can be loaded into Simulink in its entirety.
   - Called for each body as the format of body.hydroForce can vary with GBM, mean drift, etc

Minor changes in this PR: 
- responseClass.m and postProcessWecSim.m
   - saves the hydroForceIndex to the responseClass for visualization
- addDefaultPlotVars.m
   - Minor bug fix
- wecSimMCR.m and initializeWecSim.m 
   - renames saved MCR hydroData to prevent conflicts

TODO list, roughly in suggested order
- [x] Allow variable hydrodynamics in bodyClass.m
- [x] Allow variable hydrodynamics in Body library
- [x] Add initial value for hydroForceIndex that does not change
- [x] Allow the hydroForceIndex to change in the library so that users can control switching
- [x] Update mask initialization to select the default hydroForce based on ``body.hydroForceIndexInitial``
- [x] Allow different bodies to contain different numbers of h5Files
- [x] Allow variableHydro.option to be body dependent. This seems to be breaking due to the bus set-up
- [x] Allow displaced volume and centerBuoyancy to vary with hydroForce. These are copied to hydroForce and applied from there during the simulation
- [x] Allow quadratic drag and linear damping to vary with hydroForce?
- [x] Adding switching logic example
- [x] Fix added mass post-processing
- [ ] Add global GoTo blocks for body forces, constraint forces, time
- [x] Evaluate the model instability when switching between different h5files --> see https://github.com/WEC-Sim/WEC-Sim_Applications/pull/65
- [x] Validate a case study (passive yaw can highlight benefits/drawbacks of VH vs another advanced feature) --> see https://github.com/WEC-Sim/WEC-Sim_Applications/pull/65
- [ ] Add documentation

Future work:
- [ ] Time varying mass and inertia
- [ ] Enable variable hydro with the state space and FIR filter blocks --> TBD, this may not be possible. Requires setting a mask variable through a port.
- [ ] Additional applications (large amplitude sphere, VG OSWEC)


kmruehl:
Based on @dav-og suggestion, consider including passive yaw as use case and deprecating the current passive yaw implementation

akeeste:
One follow-up on passive yaw-- A feature that passive yaw has that is not currently planned here is interpolation between data points. Passive yaw (I believe) will interpolate between directions if past the threshold. We were not planning that here since right now the user-defined switching parameter is not a dimension of each hydroForce array. That may be possible, but the implementation will definitely be more complicated. We were planning to force the user to include as more discretization for their state as is required

dav-og:
Thats correct, passive yaw interpolates hydrodynamic coefficients between headings - for a user-defined threshold. The results can change a lot depending on this threshold - but if you set it very low then the computation time explodes. User has to evaluate this speed-accuracy trade off.

For passive yaw, the hydrostatic stiffness matrix is rotated with the body in WEC-Sim

akeeste:
I made some nice progress on this PR lately by turning the body.hydroForce structure into a bus signal. All data nested within body.hydroForce can be loaded once as a constant in the simulation. The data to use at a given time can be indexed using a signal defined by the user. Some progress left to go, but this is a big step in moving this PR forward.

[test_v2.zip](https://github.com/user-attachments/files/15684044/test_v2.zip)

akeeste:
Excitingly, I got this feature working for a basic case now! I temporarily edited the RM3 example model and input files to run when importing the a set of 3 h5files into the same body. There a few more complex cases to test/implement next:
- [ ] import different h5files and evaluate the model instability when switching
- [ ] allow different bodies to contain different numbers of h5Files. I'm not sure this is working yet

akeeste:
Comment from @MShabara - we may want to reference the [buffered fourier spectral method](https://www.math.ualberta.ca/ijnam/Volume-9-2012/No-2-12/2012-02-25.pdf) to address unstable switching in this feature

akeeste:
Recording a few recent updates:

Right now the implementation mostly works but has a couple odd edge cases that should work but don't. I'm working through fixes for these edge cases.

Working scenarios:
- all bodies use variable hydro (``body.variableHydro.option=1) and each has multiple h5Files. 
- no bodies use variable hydro

Breaking scenarios:
- some but not all bodies use variable hydro
- a body with variable hydro has only one h5 file

akeeste:
> Breaking scenarios:
> 
> * some but not all bodies use variable hydro
> * a body with variable hydro has only one h5 file

The latest commit will fix the two issues noted immediately above. Something in the bus still breaks when there is only one h5 file, but that case should not need variable hydro so I added a check to prevent it.

akeeste:
Summarizing a discussion that @salhus and I had on this feature:

An important but currently missing aspect of this PR is how to treat the CG, mass, and inertia if a body state changes. E.g if a device is flooded, its cg, mass, and inertia all change. I do not think Simscape will allow the inertia tensor of the body itself to change during the simulation.

We could account for the fact that the gravity force changes because WEC-Sim calculates that manually but it will add complexity for the user. If they can account for a change to gravity through the cg and volume that would be best. It seems unlikely we could use this feature to model cases where the mass and inertia of a body change drastically. If so it will take a lot more scrutiny, effort and more development.

kmruehl:
@akeeste do you know why these tests are failing? And should there be merge conflicts?

kmruehl:
@akeeste the merge conflicts are resolved, but do you know why the tests are failing?

examples/RM3/wecSimInputFile_orig.m
kmruehl, line 1: Should this file be in the PR?

kmruehl (commented):
It looks like there is one file that should be removed from this PR

kmruehl:
> @akeeste the merge conflicts are resolved, but do you know why the tests are failing?

@akeeste the tests are throwing a warning when the Simulink files are copied, and the MATLAB version doesn't match. This happens in `C:\Users\kmruehl\Documents\GitHub\WEC-Sim\WEC-Sim\tests\runFromSimTest.m` lines 34-37

``` matlab
            copyfile(fullfile(testCase.wsDir,'/examples/RM3FromSimulink/RM3FromSimulink.slx'),...
                fullfile(testCase.runFromSimDir,'fromSimInput.slx'));
            copyfile(fullfile(testCase.wsDir,'/examples/RM3FromSimulink/RM3FromSimulink.slx'),...
                fullfile(testCase.runFromSimDir,'fromSimCustom.slx'));   
``` 

However, the error is due the following: 

``` matlab
    ---------
    Error ID:
    ---------
    'MATLAB:num2str:nonNumericInput'
    --------------
    Error Details:
    --------------
    Error using num2str (line 24)
    Input to num2str must be numeric.
    
    Error in writeBlocksFromInput (line 160)
            values{i,1} = num2str(maskVars.(names{i,1}));
    
    Error in loadInputFileCallback (line 40)
                writeBlocksFromInput(blockHandle,type,inputFilePath);
    
    Error in runFromSimTest/setupInputFiles (line 81)
                        loadInputFileCallback(grfBlockHandle);
```
`maskVars.(names{i,1})` is not numeric

examples/RM3/wecSimInputFile_orig.m
akeeste: No, I'll remove

akeeste:
@kmruehl I saw these tests failures yesterday. In short, no I don't yet know why. I did not touch any WEC-Sim GUI or Run From Simulink functionality. I'm guessing there was a problem coming from library changes in main that were pulled into dev, and then into my branch

akeeste:
@kmruehl The WEC-Sim GUI issue should be resolved pending the test completion. I thought the bug was coming from a library merge that I messed up because a non-variable hydro case was failing. But the issue was on my development efforts--related to body.h5File becoming a cell which is not allowed in a mask

kmruehl:
@akeeste thanks for the update, looks like all tests are passing now. Thanks!

kmruehl (approved):
@akeeste thanks again for your help resolving the tests. Everything looks good, but I'm going to go. I'm going to go through it myself one more time before merging.

Issue #1116, opened 2023-09-03
[BUG] Phase angle of wave excitation loads for offset body, and added mass and radiation damping matrix for multibodies in AQWA


hachikoi1:
**Note: italicized text below is include as an example and should be updated before submission. If you feel any section is not applicable to your request, please replace with 'N/A' rather than delete a section.**

**Describe the bug**
_A clear and concise description of what the perceived bug is._

Dear [salhus](https://github.com/salhus) and [dforbush2](https://github.com/dforbush2)

I open a new issue here to talk about the Phase angle of wave excitation loads for offset body, and added mass and radiation damping matrix for multibodies. The issue is was originally posed in **A designed two-body hinged raft wave energy converter: From experimental study to annual power prediction for the EMEC site using WEC-Sim** by Siya.

Let me briefly describe the problem. For an offset body with its COG away from the origin of global coordinate system in the XY plane (this configuration is usually found in multibody cases), the phase angle output by AQWA is defined as the local parameter phase angle. But the phase angle output by WAMIT is defined as the global phase angle, which is equal to the local parameter phase angle + global wave phase angle (global wave phase angle is defined as the phase angle of the local coordinate origin with relative to the global coordinate origin, and is related to the spatial location of the body)

AQWA
![image](https://github.com/WEC-Sim/WEC-Sim/assets/106663159/00389944-0ed2-48c1-a1a2-bb956f66b85c)

WAMIT
![image](https://github.com/WEC-Sim/WEC-Sim/assets/106663159/6fbe5197-301f-4ea1-89c6-c71bd7581f6b)
![image](https://github.com/WEC-Sim/WEC-Sim/assets/106663159/ed34bf9f-15ca-49d7-ae29-47ce3c25ff4e)

Based on previous discussions (https://github.com/WEC-Sim/WEC-Sim/issues/1090#issue-1828915005), for a body with its COG located at the origin of global coordinate system in the XY plane

 PhaseAngleWAMIT = -PhaseAngleAQWA

But the definition is not perfect, because it is given based on the premise that the wave phase angle is zero (COGx = COGy = 0). For an offset body, the correct phase angle should be 

PhaseAngleWAMIT = -PhaseAngleAQWA - k * [x * cos(beta) + y * sin(beta)]
where PhaseAngleAQWA is the phase angle ouput by AQWA, k is the wave number, x and y are locations of the body, and beta is the wave heading. The corrected code in **readAQWA.m** is:

[...]

PR #1252, opened 2024-05-03
[BUG] Update readAQWA to fix excitation phase and b2b loading issues

jtgrasb:
This PR fixes 2 bugs in the readAQWA.m script:
1. The excitation phase in Ansys AQWA is given relative to the center of gravity of each body and a positive phase angle represents the excitation lagging behind the wave.

On the other hand, WEC-Sim follows WAMIT's standard, which is to define the excitation phase relative to the global origin and a positive phase angle meaning the excitation leading the wave.

The current readAQWA.m script negates the phase correctly to adjust to the discrepancy in leading/lagging, but does not take into account the difference in reference frames. Thus, WEC-Sim does not apply the correct excitation phase when using AQWA and a body is not located at the global origin. I updated the readAQWA.m script to adjust the phase of the excitation coefficients upon loading them into the hydro structure. This approach uses the equation (k is wave number, x and y are the coordinates of the body's cg, and $\beta$ is the wave direction): 
$\Theta_{WAMIT} = -\Theta_{AQWA} - k  [x  cos(\beta) + y  sin(\beta)]$

To demonstrate the difference, I used the example from PR #1191 and ran a 4 cylinder (locations: [-8, -8], [-8, 8], [8, -8], [8, 8]) case in both AQWA (blue, hydro_1) and Capytaine (orange, hydro_2) and compared the results. Before the updates in this PR, the excitation phase was significantly different, while after the updates in this PR, the results show much better match:

Before:  After PR: 

2. The body-to-body elements of the added mass and radiation damping matrices were being read in backwards. When looking at the AH1 output file, each added mass and radiation damping matrix has 3 leading numbers (first 2 related to the body interactions and last number related to the frequency number). When outputting the AH1 descriptions, AQWA states that `"  1  1" (i3,i3) means the force on structure 1 caused by the motions of structure 1.` Thus, I interpret this as the 1st number representing the body on which the forces act and the 2nd number being the body causing the forces. When we load these values into WEC-Sim and turn on b2b, it is clear that the interactions are being read in backwards. From AQWA RM3 case, the force on body 1 from body 2 (frequency 1) should have elements: 

But, with simu.b2b = 1, the body-body interaction terms for the force on body 1 from body 2 (frequency 1) are slightly different: 

When looking at the opposite terms (force on body 2 from body 1) as well, it is clear that the interaction terms are backwards. 

AQWA output: 

WEC-Sim Forces: 

To put it more plainly, readAQWA originally loaded the added mass and radiation damping terms as in eq. 7 below, while this PR corrects it to be as eq. 8. 

image

Some additional useful references include [this paper on a 2 body hinged raft WEC](https://www.sciencedirect.com/science/article/pii/S0029801822025690?ref=pdf_download&fr=RR-2&rr=87e20e333c4be8fd#appsec1) which uses WEC-Sim but demonstrates the two readAQWA.m updates which were needed to correct their results and issue #1116 where both issues are explained as well.

### DIFF
--- a/source/functions/BEMIO/readAQWA.m
+++ b/source/functions/BEMIO/readAQWA.m
@@ -104,6 +104,14 @@ for ln = n:length(raw1)
             hydro(F).g      = tmp(3);   % Gravity
         end
         clear V182
+        % Set water depth conditions for calculating wavenumber later
+        if hydro(F).h == inf || hydro(F).h >= 200
+            deepWater = 1;
+            waterDepth = 200;
+        else
+            deepWater = 0;
+            waterDepth = hydro(F).h;
+        end
     end
     if isempty(strfind(raw1{ln},'COG'))==0
         for i=1:hydro(F).Nb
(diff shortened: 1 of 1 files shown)


jtgrasb:
Closing in favor of new PR into main branch to resolve the bug.

PR #1256, opened 2024-05-15
Fix direct drive bugs

jtgrasb:
The inertia and damping terms of the direct drive PTO had the wrong sign. This PR updates them to both have a negative sign.


dforbush2:
thanks for the fast fix @jtgrasb 
squash and merged.

PR #1263, opened 2024-06-08
Update readCAPYTAINE.m to read Khs from .nc

salhus:
Latest version of capytaine writes hydrostatics data in the .nc file. This PR adds some code to read the hydrostatics data directly from the .nc file and removes the need to have a separate function that first writes a .dat file in python.

### DIFF
--- a/source/functions/BEMIO/readCAPYTAINE.m
+++ b/source/functions/BEMIO/readCAPYTAINE.m
@@ -218,7 +218,7 @@ for m = 1:hydro(F).Nb
         raw = textscan(fileID,'%[^\n\r]');
         raw = raw{:};
         fclose(fileID);
-        for i=1:6
+        for i = 1:6
             tmp = textscan(raw{i},'%f');
             hydro(F).Khs(i,:,m) = tmp{1,1}(1:6);  % Linear restoring stiffness
         end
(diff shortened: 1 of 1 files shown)


akeeste:
Thanks @salhus. I pushed a few updates to this PR:
- [x] updated error handling for Khs
- [x] combined the code that searches the .nc and .dat files for Khs into one segment for more robust error handling
- [x] added the same Khs workflow for the other hydrostatics parameters that are read from Hydrostatics.dat

If a user properly adds all hydrostatics data to their Capytaine dataset (``hydrostatic_stiffness``, ``center_of_mass``, ``center_of_buoyancy``, ``volume``), users should not need any custom .dat files.

@mancellin It would be much more robust for BEMIO to read  ``center_of_mass``, ``center_of_buoyancy``, and ``volume`` if they were added to the output dataset by default.

Noting that right now the Khs will not read from the .nc file properly if bodies have <6 DOFs. This is an ongoing problem and I'm working on a broader fix for readCapytaine in this case.

akeeste:
I will merge when the tests pass

mancellin:
> It would be much more robust for BEMIO to read center_of_mass, center_of_buoyancy, and volume if they were added to the output dataset by default.

I was delaying this until I have a proper way to handle also the multiple body cases, but maybe we can implement something already for the single rigid body case.

akeeste:
thanks @mancellin, great to hear that it is in progress. No rush from my perspective. When this is added on the Capytaine side we will update the BEMIO function accordingly.

PR #1198, opened 2024-01-05
PTO Extension Feature

Allison-File:
The PTO Extension feature was developed to address the issues described within the ‘new initial displacement implementation’ task card. The current initial displacement feature doesn't actually displace bodies but instead changes their locations entirely. The PTO Extension feature corrects this issue by allowing for the defined PTO displacement to propagate throughout the system, resulting in an accurate reflection of the initial position of each body without requiring the user to individually calculate and define. This feature was implemented on several PTO library blocks including: Translational PTO, Rotational PTO, Rotational PTO Actuation Motion, and Translational PTO Actuation Force.

The PTO Extension Feature includes the following changes:

1. Established an ‘extension’ property within the PTO Class
        -This property defines the priority, value, and checkbox initialization options for the specify position target checkbox of 
           each PTO library block joint

2. Edited the PTO library blocks
        - Set the joint block position target value as pto.extension.PositionTargetValue, which can be defined in the 
           wecSimInputFile

        - Added an if statement in the PTO library block mask to allow the pto.extension.PositionTargetSpecify to 
          turn on the specify position target checkbox if defined as a value of 1 in the wecSimInputFile.

3. Documentation
        - Updated the pto.rst file to include a detailed explanation of the PTO Extension Feature and how to initialize within the 
         wecSimInputFile. This includes a visual of the feature being implemented within the WECCCOMP wavestar model.

        - Updated the applications.rst file to include an explanation of the applications case employing the PTO Extension 
         feature for the RM3 model.

[PTO Extension Feature.pdf](https://github.com/WEC-Sim/WEC-Sim_Applications/files/13892141/PTO.Extension.Feature.pdf)

PR #1264, opened 2024-06-10
Add missing mask initialization line in spherical constraint

akeeste:
#1198 added PTO/Constraint extension. I missed a line in the new mask initialization when I extended the PTO extension feature to the spherical constraint, causing the Cable Application to fail on dev. This PR fixes that line and the Cable example runs locally now.

#1198 is only in dev, so this bugfix is going straight into dev.

I double checked the other relevant constraint/PTO blocks and they should have the right mask commands already.


akeeste:
@MShabara Once these tests finish passing, I will merge this PR so that the Applications tests re-run and we can confirm that the Cable Application is working again.

PR #1266, opened 2024-06-25
Update readCAPYTAINE.m to keep 1st wave direction for multiple wave headings

cckimball:
When a custom Capytaine script is run with multiple wave headings (i.e. not using 'call_capytaine'), the added mass and radiation damping coefficients may create matrices with repeating values in the 4th matrix dimension (for wave direction). Since these are repeated, identical values regardless of wave direction, only the first set is required. The unnecessary addition(s) to the matrix causes issues in the remaining BEMIO function files (radiationIRF, excitationIRF, writeBEMIOH5) if left intact. 

These changes make space in the matrix for the 4th dimension, and then preserve the first case for both the added mass and radiation damping coefficients.


cckimball:
Sync issue, will open again.

Issue #574, opened 2021-05-03
Hard stops on OSWEC example


mmartini1987:
Hello
I am running the OSWEC example and trying to include hard-stops in the modelling. I have modified the wecSimInputFile.m adding the following rows:

pto(1).hardStops.upperLimitSpecify = 'on'; % 
pto(1).hardStops.upperLimitBound = 5;  % [deg] 
pto(1).hardStops.upperLimitStiffness = 0; % [Nm/deg]
pto(1).hardStops.upperLimitDamping = 1e8; % [Nm/(deg/s)]
pto(1).hardStops.upperLimitTransitionRegionWidth = 0.1; % [deg] 

pto(1).hardStops.lowerLimitSpecify = pto(1).hardStops.upperLimitSpecify;
pto(1).hardStops.lowerLimitBound = -pto(1).hardStops.upperLimitBound;
pto(1).hardStops.lowerLimitStiffness = pto(1).hardStops.upperLimitStiffness;
pto(1).hardStops.lowerLimitDamping = pto(1).hardStops.upperLimitDamping;
pto(1).hardStops.lowerLimitTransitionRegionWidth = pto(1).hardStops.upperLimitTransitionRegionWidth;

This however seems not to affect the flap motion. To see any hard-stop effect occurring, I had to navigate into OSWEC.slx->Rotational PTO->Pitch PCC, force the "Specify Upper Limit" and "Specify Lower Limit" to 1, save the model (neglecting the warnings related to having modified a WEC-Sim library block) and run it again.

Are you able to replicate this - or maybe am I mistaking something?

Many thanks
Michele

dforbush2:
Hi Michele,

What version of Matlab are you running? It doesn't sound like the mask initialization function is running properly on the pto's in the OSWEC model. I am able to replicate your error in 2020b...though. 

For whatever reason, the try/catch statement in the mask initialization is not running because it does not recognize the block name that sets the pto(1).hardStops.upperLimitSpecify and pto(1).hardStops.lowerLimitSpecify toggle switches. So you don't get an error, but your limits don't get set. Only problem is these block names are observably identical to what I have coded there, so there is something subtle happening. I will dig into this and push a fix.

In the mean time, your work-around will work. Just make sure that you do not push the changes made locally to your wec sim library or that may break some of your other models!

Best,
Dom

dforbush2:
[...]

PR #1270, opened 2024-07-01
Update readCAPYTAINE.m to keep 1st wave direction for multiple wave headings

cckimball:
When a custom Capytaine script is run with multiple wave headings (i.e. not using 'call_capytaine'), the added mass and radiation damping coefficients may create matrices with repeating values in the 4th matrix dimension (for wave direction). Since these are repeated, identical values regardless of wave direction, only the first set is required. The unnecessary addition(s) to the matrix causes issues in the remaining BEMIO function files (radiationIRF, excitationIRF, writeBEMIOH5) if left intact.

These changes make space in the matrix for the 4th dimension, and then preserve the first case for both the added mass and radiation damping coefficients.

### DIFF
--- a/source/functions/BEMIO/readCAPYTAINE.m
+++ b/source/functions/BEMIO/readCAPYTAINE.m
@@ -19,6 +19,7 @@ function hydro = readCAPYTAINE(hydro, filename, hydrostatics_sub_dir)
 %     hydro : struct
 %         Structure of hydro data with Capytaine data appended
 
+
 %% Check file for required variables
 [a,b] = size(hydro);  % Check on what is already there
 if b == 1 && ~isfield(hydro(b),'Nb')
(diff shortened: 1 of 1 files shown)


akeeste:
@cckimball Can you look at your git commit and see if you can revert it and recommit the changes? GitHub shows ever line in readCapytaine as changed and I can't tell what's going on.

Also, without seeing a specific case or set-up, this sounds like a bug on the Capytaine side instead of WEC-Sim. Is there any reason this shouldn't be fixed in Capytaine instead? @mancellin

akeeste:
Also there are some other updates going into readCapytaine right now. Please pull those changes to your branch so there is no merge conflict

akeeste:
I just saw there was another PR with some git syncing issues. Maybe before diving into the git issues further, we first decide if this is WEC-Sim or Capytaine bug

mancellin:
> When a custom Capytaine script is run with multiple wave headings (i.e. not using 'call_capytaine'), the added mass and radiation damping coefficients may create matrices with repeating values in the 4th matrix dimension (for wave direction).

Sounds like a bug in Capytaine.
Added mass and radiation damping are independent of the wave direction, except when the forward speed is non-zero.
If you indeed has a `wave_direction` dimension in your added mass and radiation damping, please open an issue on Capytaine repository with a code example.

cckimball:
Thanks @mancellin and @akeeste , I've posted in the Capytaine page as issue # #574.

akeeste:
Thanks @cckimball and @mancellin. From my brief read of capytaine/capytaine#574, it sounds like this won't require any changes to readCapytaine so I will close this issue. @cckimball if you think this is still worth checking for on the BEMIO side, please reopen and we can look at options.

PR #1273, opened 2024-07-12
Expand regression tests

akeeste:
This PR expands the regression tests to account for accuracy bugs in BEMIO functions.

Currently the regression tests run the RM3 model using the committed hydroData in ``examples/RM3/hydroData/rm3.h5``.

If a BEMIO function is updated and runs functionally, but is inaccurate, the regression test will still pass because BEMIO is not being re-run. e.g. the tests [here](https://github.com/WEC-Sim/WEC-Sim/pull/1263/checks?check_run_id=27383027089) should have failed because the body cg, cb and volume were written incorrectly. The regression tests passed however because BEMIO is not rerun.

This update will _not_ catch inaccuracies in:
- readCAPYTAINE
- readNEMOH
- readAQWA

but it will catch issues in readWAMIT, radiationIRF, radiationIRFSS, excitationIRF, and writeBEMIOH5

### DIFF
--- a/tests/regressionTest.m
+++ b/tests/regressionTest.m
@@ -38,6 +38,16 @@ classdef regressionTest < matlab.unittest.TestCase
     
     methods(TestClassSetup)
         
+        function runBEMIO(testCase)
+            cd(fullfile(testCase.testDir,...
+                        '..',...
+                        'examples',...
+                        'RM3',...
+                        'hydroData'));
+            bemio
+            cd(testCase.testDir);
+        end
+
         function runRegTest(testCase)            
             cd(fullfile(testCase.testDir,       ...
                         'RegressionTests',      ...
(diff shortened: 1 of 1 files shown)


tests/regressionTest.m
```diff
@@ -38,95 +38,105 @@ function killPlots (~)
...
+        function runBEMIO(testCase)
+            cd(fullfile(testCase.testDir,...
+                        '..',...
+                        'examples',...
+                        'RM3',...
+                        'hydroData'));
+            bemio
+            cd(testCase.testDir);
+        end
+
```
kmruehl, line 50: I wonder if this test is better suited for `bemioTest.m`, instead of the RM3 regression test. If there is an issue with BEMIO, it seems like we should test it in `bemioTest.m`. That also allows us to compare across BEM solvers, and to a prior solution. 

Right now, we run the sphere cases in all BEM solvers, but it doesn't look like we compare the solutions. The sphere was chosen because it's a simple/fast case, but we could run the RM3 and compare the results to the uploaded h5, in the RM3 example. 

@akeeste what do you think about this suggestion?

kmruehl:
Thanks @akeeste. Let's go ahead and merge this PR to resolve the immediate issue. BEMIO is not currently run for the regression tests, so if there are any updates to `readWAMIT` they do not trick down to the regression test. This PR is a patch that resolve this issue. However, long-term we would like to add BEMIO regression tests that compare a prior BEMIO run for each BEM solver to a new run. This long-term solution has been added as a new project: https://github.com/orgs/WEC-Sim/projects/12/views/1?pane=issue&itemId=78629973/

PR #1274, opened 2024-07-12
readCapytaine - fix reading multiple bodies with <6 dofs

akeeste:
This is a bugfix PR to read Capytaine data when there are multiple bodies, each containing <6 DOF of hydrodynamic data

@jniffene see this one-line bugfix for our TEAMER project

### DIFF
--- a/source/functions/BEMIO/readCAPYTAINE.m
+++ b/source/functions/BEMIO/readCAPYTAINE.m
@@ -476,7 +476,7 @@ for k=1:length(split_body_names)
 end
 
 % set the indices that sort the old dofs/variables into the correct order
-inds = zeros(1,max([6*length(nDofs_per_body),sum(nDofs_per_body)]));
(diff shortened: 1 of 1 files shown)


akeeste:
@jniffene please test this out and let me know if it fixes the issue with our 3-body, 1-dof Capytaine case.

jniffene:
> @jniffene please test this out and let me know if it fixes the issue with our 3-body, 1-dof Capytaine case.

Yea it works now! Thanks!

akeeste:
Hey @MShabara, let me know if this fix still works for you. Building off of your comment to me, I think using the length of ``sorted_dofs`` will be the most robust because it will always contain the correct number of DOFS, even in the case of GBM modes on some but not all bodies.

MShabara:
Hi @akeeste, the modification works for me, I tested the modifications and they work fine. I will go ahead and merge the pull request.

PR #1290, opened 2024-08-05
update waveSpread calculation in irregular waves 

### LINKED ISSUE #1288: [BUG] sqrt(spreadFunction) does not conserve energy

**Describe the bug**
When wave spreading function is used, we enforce the condition 
`sum(waves.spread) ==1`

Described in literature, this constraint is imposed on the spreading function applied to the wave spectrum as a power spectral density.

 However, when applied in the irregular wave force calculation (line 11 of irregExcF) the the _amplitude_ spectra of the wave, we have
`C1 = sqrt(bsxfun(@times,A*spread(ii),dw));`

Thus implying that the square root of the spread function (elements of which are larger than those specified for the spread function) is what is utilized when calculating excitation force, noting that here A is the _amplitude_ spectra of the wave. 

**WEC-Sim file(s)**
The quickest fix I believe is irregExcF, but the proper specification of all irregular wave amplitudes should be reviewed. 

**To Reproduce**
A key example case of this is to use an axisymmetric WEC and inspect the heave amplitude in 1) a plane wave and 2) the same amplitude/period plane wave with spread applied over a few directions. Heave is insensitive to direction, so presuming that these directional waves are uncorrelated we should see identical RMS values of heave displacement in each. In fact, the heave RMS in case 2 will be significantly larger than 1).  

**Expected behavior**
Energy should be conserved over directionally spread waves.

**Additional context**
Error seems to have been introduced with the wave spread formulation: spectral wave formulation appears correct provided spread is not used.

dforbush2:
Fixes #1288


jtgrasb:
Thanks for the fix Dom!

PR #1301, opened 2024-08-29
Speed up writeBEMIOH5

akeeste:
``writeBEMIOH5`` can take a very long time. Almost all of the BEM coefficients are written twice, in the "components" groups, but never used in WEC-Sim. These groups are never read in by ``readBEMIOH5``

In my testing with the [directional OSWEC data in the Applications repo](https://github.com/WEC-Sim/WEC-Sim_Applications/tree/main/_Common_Input_Files/OSWEC/hydroData), this duplicate data takes approximately 98.9% of the CPU time of ``writeBEMIOH5`` and increases the H5 file size by almost 4 times.

This PR eliminates the duplicate writing of BEM data into the "component" groups and significantly reduces the computational expense of ``writeBEMIOH5``. This will critically enable the use of variable hydro.

### DIFF
--- a/source/functions/BEMIO/writeBEMIOH5.m
+++ b/source/functions/BEMIO/writeBEMIOH5.m
@@ -1,4 +1,4 @@
-function hydro = writeBEMIOH5(hydro)
+function writeBEMIOH5(hydro)
 % Writes the hydro data structure to a .h5 file.
 %
 % See ``WEC-Sim\tutorials\BEMIO`` for examples of usage.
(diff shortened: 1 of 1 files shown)


kmruehl:
Thanks @akeeste, this is a great suggestion. I certainly see value in having the component information in the h5 files for easy viewing/plotting of this data. However, that benefit is minimal when compared to the computation expense and size of the resulting h5 file. Plus, as you mentioned, the data is already contained within the h5, it just requires a bit more (or different) post-processing to plot each component. In short, I'm going to go ahead and merge this PR. If we decide we don't like it, we can revert.

kmruehl (approved):

PR #1242, opened 2024-03-25
Adds the QTFs to WEC-Sim

MShabara:
#### BEMIO:
1. Implemented functionality in the `readWAMIT()` function to read WAMIT sum and diff output files, organizing data into matrices and vectors.
2. Added capability to write QTFs data to .h5 files.

#### WEC-Sim:
1. Updated the `readh5()` function to incorporate reading QTFs data.
2. Adjusted `initializeWecSim.m` to define system variable parameters.
3. Integrated full QTF calculations in the frequency domain into the `bodyClass`.
4. Expanded `WECSim_Lib_Body_Elements.xls` to include the Second Order Excitation Forces Subsystem Variable.
5. Included QTFs in the `Run from Simulink` functionality.

These changes adds the full QTF functionality to the codebase, particularly in handling WAMIT output files and integrating QTF calculations seamlessly within WEC-Sim.

Please review and provide any feedback or suggestions. Thank you!

PR #1312, opened 2024-09-06
QTF - Variable Hydro compatibility

akeeste:
This PR finishes #1242 and makes the QTF feature compatible with variable hydro functionality which was done in parallel. Another upcoming PR will resolve the variable hydro's likely conflict with GBM case (even non variable hydro GBM cases)

@kmruehl As we've already discussed this live, I will merge once the tests pass.


kmruehl:
Thank you @akeeste!

PR #1322, opened 2024-09-12
hopefully fixes bad PR merge

dforbush2:
This is working locally but not rigorously tested with new variable hydro cases. Merges to bodyClass are intensive, search for (iH) and (hfName) is good debug check


kmruehl:
@dforbush2 thank you! Merging now

PR #1328, opened 2024-09-19
Full dir spectra2

dforbush2:
this is full directional wave spectra and may require manual merging due to different binary files. 
All applications cases are running (except maybe MOST weirdness) with this implementation.


jtgrasb:
@dforbush2 I resolved the conflicts for this PR a while ago with PR #1332, but it looks like that PR also has conflicts now due to the changes Adam made to the bodyClass. Let me know when you're ready to work on this PR again and we can talk about a path forward so I can review it.

PR #1335, opened 2024-10-05
User Defined Body Propterties

MShabara:
Allows users to set user defined cg and cb and prints the non hydro bodies info to the command window.

### DIFF
--- a/source/functions/BEMIO/readWAMIT.m
+++ b/source/functions/BEMIO/readWAMIT.m
@@ -98,7 +98,8 @@ for n = 1:N
         if isempty(strfind(raw{n},'Wave period (sec)'))==0
             T = 2;
             hydro(F).Nf = hydro(F).Nf+1;
-            tmp = textscan(raw{n}(find(raw{n}=='=')+1:end),'%f');
+            rawInd = find(raw{n} == '=');
+            tmp = textscan(raw{n}(rawInd(1)+1:end),'%f');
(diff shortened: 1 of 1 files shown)


jtgrasb:
@MShabara I confirmed that your changes work as intended and let the user specify the cg and cb, but I don't think they should be added as is. The center of gravity in BEM runs is used as the center of rotation and impacts the hydrodynamic coefficients in rotational dofs. The center of buoyancy changes based on displaced volume, which also impacts hydrodynamic coefficients. In both of these cases, BEM should be rerun. If the user has a good reason to modify them without rerunning BEM, it can still be done in BEMIO, but I don't think it should be made any easier.

jtgrasb:
@MShabara We talked about this on the Sandia side and here are the two main points to summarize our discussion:
1. Allowing users to modify the cg and cb in WEC-Sim is dangerous. Changing either of these variables should require rerunning BEM except for fringe cases. Although a warning message is helpful, it is easy to overlook and will likely lead to users mistakenly changing the cg or cb in cases where BEM should be rerun. We would like to avoid making this easier.
2. There are fringe cases where changing the cg and/or cb may be applicable without needing to rerun BEM. For these cases, we suggest a discussion thread which details why changing cg or cb is an error by default, the cases where it may be applicable without rerunning BEM, and best practice for getting around the error (e.g., modifying the h5 or editing the source code).

jtgrasb:
@MShabara See discussion #1354. Feel free to modify this PR to only include the nonhydro body print statement, and I can review it.

PR #1343, opened 2024-10-16
Variable hydro documentation

akeeste:
TODO:
- [ ]  user advanced feature section
- [ ] developer advanced feature section

### DIFF
--- a/docs/_include/variable_hydro.rst
+++ b/docs/_include/variable_hydro.rst
@@ -1,0 +1,4 @@
+.. _user-advanced-features-variable-hydro:
+
+Variable hydrodynamics is an advanced feature that...
+
--- a/docs/_include/variable_hydro.rst
+++ b/docs/_include/variable_hydro.rst
@@ -1,4 +1,75 @@
 .. _user-advanced-features-variable-hydro:
 
-Variable hydrodynamics is an advanced feature that...
+Overview
+""""""""""
+
+Variable Hydrodynamics is an advanced feature that enables users to change the 
+*state* of their device during simulation. In this context, a device *state* 
+could be determined by any number of scenarios, such as a variable geometry 
+changing shape, a flooding body, a change in operational depth, load shedding 
+capabilties, time-dependent changes to a device, etc. These states are 
+controlled by a signal, typically kinematics, dynamics, or time, which is used 
+to change the state of the device during simulation. This feature expands 
+WEC-Sim's simulation capabilities and enables
+modeling a new breadth of scenarios.
+
+Different states of a device are represented by different sets of boundary 
+element method data. User defined logic and user selected signals determine
+when the state changes. The Variable Hydrodynamics feature does not determine
+a specific scenario, state, signal, or discretization required. This is the 
+user's responsibility to implement a specific variable hydrodynamics case
+using best practices and validate with higher fidelity data.
+
+Example
+""""""""
+
+For example, take the scenario where a device submerges to shed loads. Once the
+total force on the device hits some threshold, a tether is reeled into submerge
+the device, decreasing its operational depth by three meters. This could be
+accomplished by defining a custom force on the body that is applied once the 
+total force is large enough. 
+
+Without variable hydrodynamics, as the device depth gets farther from it's 
+initial position, the hydrodynamic loading becomes increasingly inaccurate.
+Using variable hydrodynamics, the hydrodynamic loading can remain accurate 
+updating it to be being based on the device's instantaneous depth.
+
+In this example, the state of the device is its operational depth 
+(heave position). The total force and heave position are both signals that 
+dictate the varying hydrodynamics. The total force trigger a custom PTO
+force, and the heave position determines what BEM dataset is used as the body
+submerges.
+
+User defined logic dictates how quickly the operational depth changes, and 
+selects the corresponding BEM data at the new state.
+In this case, the user could supply BEM datasets of the device at 
+30 different depths over the 3m range. User defined logic then selects 
+the BEM dataset most closely corresponding to the instantaneous 
+heave position of the device. That new BEM dataset is then used to 
+calculate the hydrodynamic forcing on the device until updated again.
+
+Implementation
+""""""""""""""
+Variable hydrodynamics is implemented by allowing users to initialize a body
+with as many h5 files as required. Each i-th h5 file is then pre-processed into
+``body.hydroData(i)`` and ``body.hydroForce.hf{i}`` in the normal manner.
+The entirety of body.hydroForce is loaded into Simulink using custom buses. 
+Users write the index ``body1_hydroForceIndex`` in Simulink, which indexes 
+``body.hydroForce.hf{i}`` and selects the hydrodynamic force coefficients used
+in the hydrodynamic force calculations at a given time.
+
+
+Tips
+""""
+- investigate other features that can accomplish your modeling goals (passive yaw, large XY displacements, etc) more effectively
+- keep state range small
+- input BEM data to cover the entire range of the state 
+- a very fine state discretization may be required. study and iterate to fine the best state discretization
+- if a very high state discretization is required, use BEMIO to interpolate between points instead of running 1000 different BEM simulations
+- validate using high fidelity data
+- the hydroData directory may become very large
+- all BEM data is saved to the ``body`` variable. Pre-processing remains very fast, so it's not recommended to save ``body`` to an output file or the size increases drastically.
+- variable masses are not yet implemented.
+
+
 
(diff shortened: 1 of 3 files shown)


akeeste:
This is now ready for an initial review. As I was the one who created the feature, I don't have a good sense of what users will need detailed documentation on. I could use a review and suggestions about additional content to cover.

akeeste:
Doc tests should pass once I merge https://github.com/WEC-Sim/WEC-Sim/pull/1348 (pulling bug fixes into dev) and then pull those commits into this branch

docs/_include/variable_hydro.rst
```diff
@@ -0,0 +1,113 @@
...
+during simulation. This feature expands WEC-Sim's simulation capabilities and enables
+modeling a new breadth of scenarios in the time domain.
+
+Different states of a device are represented by different sets of boundary 
+element method data. User defined logic and user selected signals determine
+when the state changes. The Variable Hydrodynamics feature does not determine
+a specific scenario, state, signal, or discretization required. It is the 
+user's responsibility to implement a specific variable hydrodynamics case
+using best practices and validate the results with higher fidelity data.
+
```
kmruehl, line 21: This section is that first one that users will read. I recommend simplifying it to clearly describe what/how variable hydro works, and then provide an example. Here's my initial attempt:

Variable hydrodynamics allows WEC-Sim bodies to simulate bodies with hydrodynamic data from multiple BEM runs for a single body. Hydrodynamic data from each BEM run is parsed into H5 files, which are then initialized in the instance of the body class, for example `body(1) = bodyClass({'H5FILE_1.h5','H5FILE_2.h5',...);`. This feature allows users to simulate bodies with different initial conditions, for example body that is floating or submerged. Users then specify the condition to switch when each H5 is used. The condition must be based on a WEC-Sim state...

docs/_include/variable_hydro.rst
```diff
@@ -0,0 +1,113 @@
...
+should represent the body at the start of a simulation and likely its equilibrium 
+position. The initial dataset will be used to define the mass, center of gravity, 
+and center of buoyancy for a simulation.
+
+This parameter is flexible because an initial index of zero is not always convenient
+and can complicate indexing logic. For example, consider a flap-based device with
+multiple hydrodynamic datasets at various pitch angles (-50:10:50). It is most convenient
+to treat these angles in numerical order in BEM simulations, indexing logic, and other 
+data processing. In this case the initial position is at a pitch angle of 0 so ``body.variableHydro.hydroForceIndexInitial=6;``.
+
```
kmruehl, line 88: This level of detail seems more appropriate for the developer documentation. I recommend simplifying the user docs to only what the user needs to know to implement the feature, which is that they need to specify which index to start with. I don't even know how critical it is for a user to understand the bus signals.

docs/_include/variable_hydro.rst
```diff
@@ -0,0 +1,113 @@
...
+Compatibility
+"""""""""""""
+Variable hydrodynamics is not compatible with certain WEC-Sim features. It cannot be used concurrently with:
+
+* State-space radiation calculations
+* FIR Filter radiation calculations
+* Generalized body modes within the same body. Seperate bodies could each have generalized body modes and variable hydrodynamics however.
+* Non-hydrodynamic and drag bodies
+
```
kmruehl, line 97: Please convert this to a note. That's how we typically denote if a feature is not compatible with another. For example,  see PCT note here: https://wec-sim.github.io/WEC-Sim/main/user/advanced_features.html#parallel-computing-toolbox-pct

docs/_include/variable_hydro.rst
```diff
@@ -0,0 +1,113 @@
...
+""""
+* Investigate which advanced feature (variable hydrodynamics, passive yaw, large XY displacements, etc) will accomplish your modeling goals most effectively
+* Keep the state's range small
+* Input BEM data to cover the entire range of the state
+* A very fine state discretization may be required. Conduct a state discretization study to find the resolution required for accurate results.
+* If a very fine state discretization is required, use may be able to use BEMIO to create new datasets by interpolating between BEM simulations instead of simulating thousands of distinct BEM solutions. See the :ref:`user-applications-variable-hydro` Application for an example.
+* Validate using high fidelity data
+* The hydroData directory may become very large
+* All BEM data is contained within the ``body`` variable. Pre-processing remains very fast, so it's not recommended to save ``body`` to an output file or the file size may increase drastically.
+* Conditions that require a varying mass, center of gravity, or center of buoyancy are not yet implemented.
```
kmruehl, line 113: Similar comment, I this this is better suited as a note that this feature should be used with extreme caution. I also recommend moving a lot of these recommendations to the README in the variable hydro application case.

kmruehl (changes requested):
@akeeste I reviewed this content, and I have several recommendations. I'll put my commends in the review, and then we can discuss how to rework the documentation.  I'm happy to help rewrite the text.

docs/_include/variable_hydro.rst
kmruehl: @akeeste I used ChatGPT to refine the overview and here's the suggestion:

## Overview

Variable Hydrodynamics enables WEC-Sim bodies to utilize hydrodynamic data from multiple BEM runs within a single body instance. Each BEM run's data is stored in H5 files, which are initialized in the body class, for example:
  
`body(1) = bodyClass({'H5FILE_1.h5', 'H5FILE_2.h5', ...});`
  
 This feature allows users to simulate different initial conditions, such as floating or submerged states. Users specify the conditions that trigger the use of each H5 file, based on a WEC-Sim state.

docs/_include/variable_hydro.rst
kmruehl: Here is the ChatGPT recommendation for Implementation

## Implementation
To implement Variable Hydrodynamics, follow these steps:

1. **Initialize the Body**: Create a body instance with a cell array of H5 files containing the hydrodynamic data. For example:

   `body(1) = bodyClass({'H5FILE_1.h5', 'H5FILE_2.h5', 'H5FILE_3.h5', ...});`

2. **Enable Variable Hydrodynamics**: Set the variableHydro.option flag to enable the feature:

   `body(1).variableHydro.option = 1;`

3. **Set Initial Hydrodynamic Coefficient Index**: Use the hydroForceIndexInitial flag to define the default index for hydrodynamic coefficients, typically representing the body’s starting condition:

   `body(1).variableHydro.hydroForceIndexInitial = <initial_index>;`

4. **Simulink Integration**: Load the hydrodynamic force data into Simulink using a custom bus. Create a GoTo block named body1_hydroForceIndex (or similar) to allow the body block to read the index signal for selecting the appropriate hydrodynamic coefficients.

5. **User-Defined Logic**: Implement your own logic in Simulink to determine when to switch between H5 files based on the specified conditions.

_Note: Variable hydrodynamics can be applied to individual bodies in a simulation, and it is not necessary for all bodies._

kmruehl (commented):
I recommend using ChatGPT to improve our documentation. I fed it your text, my text and asked for it to the docs clearer and more concise. I think it did a pretty good job.

docs/_include/variable_hydro.rst
akeeste: I can change this to a note or paragraph, but I don't think the notes will be read if moved to the Application readme. Those documents tend to contain very basic information about a specific case, whereas these tips will be applicable to any simulation with variable hydrodynamics

docs/_include/variable_hydro.rst
akeeste: I like the suggested numbering of steps for a user's implementation. I combined this with the detailed paragraphs to walk a user through set-up.

docs/_include/variable_hydro.rst
akeeste: I agree this text could be more concise, I'll revise this text. 

The ChatGPT response is certainly concise, bit its inaccurate when for some of the details (H5 files aren't being initialized, the feature is not about initial conditions)

akeeste:
@kmruehl I made some revisions to the documentation based on your suggest. See above comments. Given the challenges of using this feature, I think it is worth leaving more detail and explanation for users than less. Let me know if you think there is any topic missing.

kmruehl:
@akeeste thanks, I'll review it again. Also, do you know why the documentation build isn't passing?

akeeste:
@kmruehl Any final comments on this PR before the release?

kmruehl:
@akeeste sorry for the delay, I've been OOO. I'll review this today.

docs/_include/variable_hydro.rst
```diff
@@ -0,0 +1,134 @@
...
+    * Generalized body modes
+    * Non-hydrodynamic and drag bodies
+    * Conditions that require a variable mass, center of gravity, or center of buoyancy
+
+Application
+""""""""""""
+See the :ref:`user-applications-variable-hydro` WEC-Sim_Application for a demonstration of setting up and using variable hydrodynamics.
+
+Additional Considerations
+""""""""""""""""""""""""""
```
kmruehl, line 118: I still don't love this section here, but I see its value and I'm okay with it for now. We should consider if this is how we want to do a feature disclaimers, or whether they belong with the specific application.

kmruehl (approved):
@akeeste thank you for the updates. I'll approve and merge

PR #1330, opened 2024-09-20
Fixing documentation inconsistency for PTO-Sim blocks

jleonqu:
This PR is related to issue #1314 
The comments for the PTO-Sim class have been revised and updated

PR #1342, opened 2024-10-16
Update Google analytics implementation in docs 

akeeste:
This PR updates the Google analytics implementation in our docs per the deprecation of ``analytics_id``. See example of failing docs build [here](https://github.com/WEC-Sim/WEC-Sim/actions/runs/11368439659/job/31623533133?pr=1340). Per that deprecation warning, I utilize [sphinxcontrib.googleanalytics](https://pypi.org/project/sphinxcontrib-googleanalytics/) in the build environment

PR #1341, opened 2024-10-16
Fixing documentation inconsistency for PTO-Sim blocks (#1330)

kmruehl:
* Fixing documentation inconsistency for PTO-Sim blocks

* Adding comments to nondimensional values for clarification

PR #1348, opened 2024-10-30
Pull bug fixes #1330 and #1342 into dev

akeeste:
Pull bug fixes #1330 and #1342 into dev


kmruehl:
@akeeste #1330 should already be in dev, but I'll review this PR to confirm

akeeste:
@kmruehl agreed, I missed that. Looks like https://github.com/WEC-Sim/WEC-Sim/pull/1341 pulled both these commits into dev and I see them there. Maybe we squashed that PR and that's why it looks like we're still bringing them in?

PR #1352, opened 2024-11-01
WEC-Sim Quadratic Transfer Functions compatability with NEMOH

MShabara:
This PR adds the two QTF features:
1- Reading the NEMOH QTF outputs to BIMIO.
2- Does the contour plots of the mod of the sum and diff.

With this update WEC-Sim should be able to compute the second order forces based on the NEMOH outputs.


MShabara:
Replaced with another PR done based on the dev branch

PR #1355, opened 2024-11-01
 User Defined Body Propterties v2

MShabara:
Replaced the PR #1335 
Allows the user to change the name of the hydro bodies.
Allows printing the non-hydro bodies data in the command window.

### DIFF
--- a/source/functions/initializeWecSim.m
+++ b/source/functions/initializeWecSim.m
@@ -513,21 +513,17 @@ for iW = 1:length(waves)
 end; clear iW
 fprintf('\nList of Body:\n ');
 fprintf('Number of Hydro Bodies = %u \n',simu.numHydroBodies)
-i = 1; 
-while i <= simu.numHydroBodies
+for i = 1:simu.numHydroBodies
     if body(i).nonHydro == 0
-        body(i).listInfo(i)
+        body(i).listInfo(i,'0')
     end
-    i = i+1;
 end;  clear i
 if numNonHydroBodies ~= 0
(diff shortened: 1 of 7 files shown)


source/functions/initializeWecSim.m
```diff
@@ -511,11 +511,25 @@
...
     waves(iW).listInfo();
 end; clear iW
-fprintf('\nList of Body: ');
-fprintf('Number of Bodies = %u \n',simu.numHydroBodies)
-for i = 1:simu.numHydroBodies
-    body(i).listInfo
+fprintf('\nList of Body:\n ');
+fprintf('Number of Hydro Bodies = %u \n',simu.numHydroBodies)
+i = 1; 
+while i <= simu.numHydroBodies
```
jtgrasb, line 517: To be consistent with the rest of the source code, I think we should use for loops. "for i = 1:simu.numHydroBodies"

akeeste:
FYI @jtgrasb @MShabara all can ignore my two test commits. They did not end up changing anything. I was testing a github issue and mshabara's is the only repo I don't currently have collaborator access to, so I used this PR.

source/functions/initializeWecSim.m
MShabara: I applied your recommendations.
I also made the code print the non-hydro body details same as the hydro body

jtgrasb:
@MShabara Thanks for the updates. The last thing is to discard the changes to `rm3.h5` from the PR, then it will be ready for merge.

MShabara:
I discarded the changes to rm3.h5 from the PR. If things look good to you, feel free to merge this PR.

PR #1359, opened 2024-11-04
Minor bug fix related to PR #1353

MShabara:
Unifies the mod naming across BIMIO to be `MOD_F_ij` instead of `MOD_ij`


akeeste:
This makes the naming of MOD_F_ij in QTF functions consistent. Initial bug occurred [here](https://github.com/WEC-Sim/WEC-Sim/pull/1353#discussion_r1828039112), but we'll keep the naming to  be consistent with other QTF parameters

PR #1371, opened 2024-11-18
Multiple wave instances bug fix 

MShabara:
When multiple wave instances are defined by the user, the body library duplicates the hydrodynamic calculations and combines their outputs. However, in the current version, the "From" blocks remain linked exclusively to Body(1) and are not updated dynamically for additional bodies.

To address this, changes were reverted to align the body library behavior with version 6.0. The key modifications include:

1-    Updates to the initialization functions of the following blocks:
        "Rigid Body"
        "Hydrodynamic Body"

2-    Adjustments to the following wave excitation force blocks:
        "Regular Wave Excitation Force"
        "Nonlinear Yaw Excitation Force"
        "Irregular Wave Excitation Force"
        "Irregular Wave Nonlinear Yaw"


akeeste:
@MShabara We can remove the From blocks to fix this issue for multiple waves, but they need to be replaced with lines that manually route from the output of ``Hydrodynamic Body/Variable Hydrodynamics Control`` to the ``/Hydrodynamic Body/Wave Diffraction and Excitation Force Calculation/Linear Wave Excitation Force Variant Subsystem``. We can't hardcode .hf1, etc in constant blocks with variable hydro.

MShabara:
@akeeste 

Thanks for the message. I made a replacement PR  #1373  that solves this bug using another method. If the new PR looks good for you please cancel this one.

akeeste:
Closing in favor of #1373

PR #1378, opened 2024-12-02
MoorDyn variable path support

RyanDavies19:
This PR fixes some MoorDyn path things to support V2 (input files can have any name or path). Previously there were instances of MoorDyn v1 file naming that was hard coded. Now, the input file name and path is used to derive the output file names and paths. This requires the MoorDyn filename in the wecSimInputFile to only have a single "." character, I added tests accordingly that check this and that the file exists. 

If there are formatting things that need changing here please let me know! I tried to follow how other parts of the code do things as closely as possible.

### DIFF
--- a/source/objects/responseClass.m
+++ b/source/objects/responseClass.m
@@ -321,7 +321,7 @@ classdef responseClass<handle
             end
         end
         
-        function obj = loadMoorDyn(obj,linesNum)            
+        function obj = loadMoorDyn(obj,linesNum, InputFile)            
             % This method reads MoorDyn outputs for each instance of the
             % ``mooringClass``
             %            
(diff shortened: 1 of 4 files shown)


source/objects/responseClass.m
```diff
@@ -321,23 +321,33 @@
             end
         end
         
-        function obj = loadMoorDyn(obj,linesNum)            
+        function obj = loadMoorDyn(obj,linesNum, InputFile)
```
jtgrasb, line 324: Use "inputFile" with camel case to be consistent. Same with "inputStrings" and "fileRootPath" below.

jtgrasb:
@RyanDavies19 I left one minor comment. Other than that, I tested it out and this looks great and will be ready for merge!

source/objects/responseClass.m
RyanDavies19: Done, thanks for the review!

PR #1379, opened 2024-12-09
Update flex body (GBM) to allow for nonlinear hydrodynamics (on first 6 DOF)

jtgrasb:
This PR updates the flex body to allow for nonlinear hydrodynamics on the first 6 dofs (non-GBM dofs). The GBM dofs are treated as linear, which is what was done for the Mondragon TEAMER award.

In the future, it may be valuable to also apply nonlinear hydrodynamics to the GBM dofs themselves. I am working on this as well, but I wanted to first enable basic compatibility with this PR.

This PR can be tested by going to the Generalized_Body_Modes folder on WEC-Sim Applications and setting `body(1).nonlinearHydro = 2;`.


jleonqu:
This PR works well on my end. It's ready to merge.

PR #1382, opened 2024-12-13
Fix mooring path check

jtgrasb:
Bug fix to dev so that we only check the mooring file path when using MoorDyn.


akeeste (approved):
Makes sense and looks good to me.

akeeste:
Approved review and tests are passing -- merging.

PR #1380, opened 2024-12-11
Pull #1373 into dev

akeeste:
This is a draft PR to pull #1373 from main into dev. The merge will likely be done locally due to the difficulty of merging the slx files.

PR #1384, opened 2024-12-13
Enable multiple wave trains with and add full directional wave block to flex body

jtgrasb:
@dforbush2  I noticed that we forgot to add the full directional wave block to the flex body. This PR adds the full directional wave to the flex body, but I am not able to test it because we don’t have an example that has both multi-directional BEM and GBM dofs. I don’t think anything should be different for the flex body though, right?


akeeste:
@jtgrasb to check that the case runs without error, you could manually modify the hydrostruct, duplicating the excitation coefficients to create a frequency direction. The data would be the same, but still allow you to test that the implementation works

dforbush2:
Woops good catch. I also don’t have anything easily testable. The dimension of the excitation force will equal 6+ flex DOF…but otherwise should work as advertised.
________________________________
From: jtgrasb ***@***.***>
Sent: Friday, December 13, 2024 1:22:52 PM
To: WEC-Sim/WEC-Sim ***@***.***>
Cc: Forbush, Dominic Dean ***@***.***>; Mention ***@***.***>
Subject: [EXTERNAL] [WEC-Sim/WEC-Sim] Add full directional wave block to flex body excitation (PR #1384)


@dforbush2<https://github.com/dforbush2> I noticed that we forgot to add the full directional wave block to the flex body. This PR adds the full directional wave to the flex body, but I am not able to test it because we don’t have an example that has both multi-directional BEM and GBM dofs. I don’t think anything should be different for the flex body though, right?

________________________________
You can view, comment on, or merge this pull request online at:

  https://github.com/WEC-Sim/WEC-Sim/pull/1384

Commit Summary

  *   39821a1<https://github.com/WEC-Sim/WEC-Sim/pull/1384/commits/39821a12b11d314cc320738909e8d28526d80bc8> Add full directional wave block to flex body excitation

File Changes

(1 file<https://github.com/WEC-Sim/WEC-Sim/pull/1384/files>)

  *   M source/lib/WEC-Sim/WECSim_Lib_Body_Elements.slx<https://github.com/WEC-Sim/WEC-Sim/pull/1384/files#diff-844e7ca775ba61c832f3d3bb16bea0742a95eaec26684288098fdd4dba57622f> (0)

Patch Links:

  *   https://github.com/WEC-Sim/WEC-Sim/pull/1384.patch
  *   https://github.com/WEC-Sim/WEC-Sim/pull/1384.diff

jtgrasb:
@akeeste Good point, I'll test this now.

jtgrasb:
Unfortunately, the same block cannot be used for the flex body because the multiple waves feature (and corresponding reformatting of the wave variable for simulink) was never made applicable for flex bodies. I'll come back to this next week and see if its reasonable to resolve quickly. It seems like it might be a bit of an effort though.

dforbush2:
To test this I recommend using an OWC example with the wave field as specified by Pioneer. There won't be proper coupling between the internal free surface and the rigid body heave, but should be enough to verify it works.

@salhus can you offer any insight on this? Do you anticipate a big roadblock to adding multiple waves to flex body?

That said, I don't think we'd ever use fullDirSpectra and multiple waves at the same time...so I'm kind of missing the dependency? Unless flex body just doesn't work at all with the revised wave class... 

I see that the mask functions related to multiple wave cases are missing for flex body. I've tried to pull this in with the naming adjusted as appropriate with the different flex body block levels...commented out for now though

jtgrasb:
@dforbush2 Thanks for taking a look. Yeah, the issue is that the flex body just doesn't work at all with the revised wave class. I tried just adding the mask to the excitation block, but was still running into some errors. I don't think it'll be too difficult but gotta take some time to debug the additional errors.

akeeste:
@jtgrasb @dforbush2 I haven't follow this too closely. Will this PR go in before the release? I need to complete #1380 which will cause library conflicts (though it should not alter the flex body block).

dforbush2:
@akeeste I think the answer is yes it has to because the revised wave class doesn't work with flexible bodies. not sure we've drilled down exactly where and when this started happening yet.

dforbush2:
@jtgrasb None of the variant subsystems initialize correctly. This PROBABLY originates in initializeWecSim and is currently a problem on the main branch too. To test: pull down MAIN of WecSim and MAIN of WEC-Sim Applications and run the GBM case. 
[WSErrorMessage.txt](https://github.com/user-attachments/files/18159913/WSErrorMessage.txt)

jtgrasb:
@dforbush2 @akeeste Yes, I am planning to get this done in time for the release. I've made some progress and getting closer to having this working. 

@akeeste if your PR is ready to merge before this one, that is fine. I can just switch out the flex body block to resolve conflicts.

jtgrasb:
I've updated the flex body excitation block(s) so they are now consistent with the way the rigid body handles the wave excitation. Now, the flex body can handle multiple wave trains as well as the full directional waves. 

I set up a test case which can comment on/off the different wave types using GBM.  [GBM_tester.zip](https://github.com/user-attachments/files/18169107/GBM_tester.zip)

I haven't tested it, but I do not think that passive/nonlinear yaw will work with the flex body (even though the Regular Wave Nonlinear Yaw excitation subsystem is in the flex body, it doesn't look like its set up for >6 dofs). But, this seems like a separate issue.

PR #1396, opened 2025-01-14
Qtf bug time interpolation

MShabara:
Fixes a bug with the time interpolation function of the QTFs.


akeeste (approved):
@MShabara I switched the PR to go into main instead of dev per our recent discussion. Otherwise this looks good to go

PR #1397, opened 2025-01-14
Applied recent commits to main into dev

MShabara:
Copy commits related to the new DOI for version 6.1.2 and apply the QTF time interpolation bug fix from main to dev.


akeeste (approved):
@MShabara Thanks, approved and merging

PR #1401, opened 2025-01-16
Bug fixes for applications

### LINKED ISSUE #1400: [BUG] Applications breaking.
Some application cases are breaking in v6.1.2. Given our current release is buggy, this should be a high priority to resolve. I just reran all Applications tests using both main branches and some are still failing. Results [here](https://github.com/WEC-Sim/WEC-Sim_Applications/actions/runs/12776912600)

- [x] [MOST](https://github.com/WEC-Sim/WEC-Sim_Applications/actions/runs/12776912600/job/35731020055)- these regression tests have also failed previous. It's unclear why an identical case results in some slightly different outputs. We can reset the regression data or try to test against a case without rerunning any processes with optimization or randomness. We've had prior discussions on this point, I'll try to find and link to them.

- [x] [Mooring/MoorDyn (and Paraview RM3)](https://github.com/WEC-Sim/WEC-Sim_Applications/actions/runs/12776912600/job/35731021137) - some drastic crash, crashing MATLAB entirely. It's hard to tell what's going on from the log, but it has failed on a 2nd retry now. This runs locally with R2023b and R2024a on windows. --> this is more or less working now. MoorDyn on MATLAB via GitHub Actions is strangely inconsistent, but it runs locally and there's a few cases of it running fine on the Paraview case or jtgrasb/WEC-Sim_Applications

- [x] [Morison element](https://github.com/WEC-Sim/WEC-Sim_Applications/actions/runs/12776912600/job/35731022001#step:9:37) - @dforbush2 @jtgrasb The full directional wave spectra #1332 seemed to break other tests running with wave spreading due to the change [here](https://github.com/WEC-Sim/WEC-Sim/blame/6a143ce063b6f4613831f5664018d62a6ec44dd2/source/objects/waveClass.m#L819). --> will be resolved by #1403

- [x] [Multiple wave headings](https://github.com/WEC-Sim/WEC-Sim_Applications/actions/runs/12776912600/job/35731023609) - failing while processing a bug signal. @akeeste will investigate --> This was related to a previous bugfix in main that got lost. Reimplemented in #1401

- [x] [OWC](https://github.com/WEC-Sim/WEC-Sim_Applications/actions/runs/12776912600/job/35731025744#step:9:186) keeps getting cancelled and failing. --> The run time is about 35 minutes, the full test time including set-up/teardown is over 45 minutes, which is over the job timeout limit. We can increase the job timeout or remove MCR runs from the OWC test. I prefer and have implemented the latter in WEC-Sim/WEC-Sim_Applications#80

- [x] [Passive Yaw](https://github.com/WEC-Sim/WEC-Sim_Applications/actions/runs/12776912600/job/35731027609#step:9:271) - the regression test is failing due to differences in results. --> My guess is that this related to the fixes for wave spreading in #1289 (also see issues #1367 #1377). If this is the case and the current simulations are more accurate, we need to update the regression data being tested against on the Applications repo. See thread below

akeeste:
PR to address the bugs appearing in our applications cases, as described in #1400. Closes #1400


akeeste:
@kmruehl this is ready for review

akeeste:
@kmruehl any updates on the review for this PR?

kmruehl:
@akeeste can you explain the differences made to the Body Library?

.github/workflows/run-tests-dev.yml
```diff
@@ -18,8 +18,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, windows-latest]
-        release: [R2022a, R2022b, R2023a, R2023b]
+        os: [ubuntu-22.04, windows-latest]
+        release: [R2022b, R2023a, R2023b, R2024a, R2024b]
     name: ${{ matrix.release }} on ${{ matrix.os }}
```
kmruehl, line 23: @akeeste is there a reason for the increase in the number of MATLAB versions being tested? We previously tested the last four versions, and now it's five.

.github/workflows/run-tests-main.yml
```diff
@@ -18,8 +18,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, windows-latest]
-        release: [R2022a, R2022b, R2023a, R2023b]
+        os: [ubuntu-22.04, windows-latest]
+        release: [R2022b, R2023a, R2023b, R2024a, R2024b]
     name: ${{ matrix.release }} on ${{ matrix.os }}
```
kmruehl, line 23: Same question here, why the increase from the last four versions of MATLAB, to five?

source/lib/WEC-Sim/WECSim_Lib_Body_Elements.slx
kmruehl, line 1: @akeeste can you explain the differences made to the Body Library?

kmruehl (commented):
@akeeste can you respond to my comments? Thanks

source/lib/WEC-Sim/WECSim_Lib_Body_Elements.slx
dforbush2: @akeeste correct me but I think I have it:
The only difference is here, in the mask initialization functions of "Hydrodynamic Body/Wave Diffraction and Excitation Force Calculation"
 
image

To my eye, this is a necessary change and all should be merged. The only reason I suspect it works as-is is that the default behavior of the blocks will carry them in all cases where there is a single set of hydrobody data (i.e., the only extant subfield is hf1). 

Short answer: this is an intentional and good change to the body library block that I suggest is merged.

kmruehl: Thank you @dforbush2

kmruehl:
I'm going to amend the test cases to revert to the last four versions of MATLAB and then merge the PR

akeeste:
@dforbush2 @kmruehl yes that change to the body library mask was in a previous bugfix and must have gotten dropped at some point during all the library changes in the latest release. The mask line is necessary for the multiple wave definitions to work with the variable hydro changes

PR #1332, opened 2024-10-02
Full dir spectra (resolved conflicts)

jtgrasb:
This PR is a for #1328 and attempts to resolve the merge conflicts.

PR #1403, opened 2025-01-24
Fix irregular wave elevation with multiple directions

jtgrasb:
This PR fixes the irregular wave elevation function, which was failing for multiple directions. This was causing the [Morison Element applications case to fail](https://github.com/WEC-Sim/WEC-Sim_Applications/actions/runs/12776912600/job/36069012775) - see #1400. The OSWEC case was also erroring because it had multiple wave directions.

@dforbush2 I traced this change back to our PR #1332, but I don't think it was intentional since you created a new function `waves.waveElevFullDir()` for the imported full directional wave elevation.

### DIFF
--- a/.github/workflows/run-tests-dev.yml
+++ b/.github/workflows/run-tests-dev.yml
@@ -18,8 +18,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, windows-latest]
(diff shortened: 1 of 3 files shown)


kmruehl:
This is related to https://github.com/WEC-Sim/WEC-Sim/issues/1406, MATLAB 2022a and 2022b ubunto tests are failing

Issue #1394, opened 2025-01-13
Improve variable hydro radiation force calculation


akeeste:
some comments/questions today from Giorgio and @ryancoe got me thinking about our radiation force with variable hydro more. I think we can improve how the convolution integral method of the radiation force is calculated:

Currently what we do (with variable hydro and fixed time-step to simplify writing the math) is:

$$F_{rad}(t_i) = \int^{t_i}_0 v(\tau) \cdot K_r(t-\tau) \cdot d\tau$$

$$F_{rad}(t_i) \approx dt \cdot \left( 0.5 \cdot v(t_i) \cdot K_r(0) + v(t_{i-1}) \cdot K_r(dt) + v(t_{i-2}) \cdot K_r(2*dt) + ... \right)$$ 

But in the above implementation, the time history of the IRF coefficient ($$K_r$$) doesn't reflect the state of the device when that contribution to the radiation force was made. So if we switched from state B at $$t_{i-2}$$ to state A at $$t_{i-1}$$, the radiation force should be:

$$F_{rad}(t_i) \approx dt \cdot \left( v(t_i) \cdot K_{r,A}(0) + v(t_{i-1}) \cdot K_{r,A}(dt) + v(t_{i-2}) \cdot K_{r,B}(2*dt) + ... \right)$$ 

@dforbush2, @MShabara thoughts?

akeeste:
The necessity of this change is dependent on how much the varying state changes over ``simu.cicTime``. Even with fine discretization, the state could change significantly over the convolution time and necessitate updated coefficients.

If the last equation above is the most accurate way to approach this issue, it shouldn't be too hard to implement. In the convolution integral calculation, we need to record the IRF coefficient history like we do the velocity history

dforbush2:
pulling from our other dialog, the issue doesn't "go away" if we have a fine discretization between hydrodynamic data sets because the data buffer for the convolution integral is (or should be) sufficiently long that we may see a significant change in the device geometry over this time. 

Practically though, how do we do this? As a straw man: I propose an additional structure probably below the (hf) substructure of the body that contains an IRF "surface" that we calculate during the pre-processing stage, basically we calculate each IRF function over CIC-time, where the additional axis is the state variable that we presently use to switch between the (hf) substructures. We then revise the IRF calculation to pull the Kr coefficients from a (probably tabular gridded interpolant) representation of this table.

akeeste:
[...]

PR #1411, opened 2025-02-06
Variable hydro improvements

akeeste:
This PR adds 
- an alternate convolution integral calculation for the radiation force and user defined excitation force when variable hydrodynamics are used. These alternate calculations use an IRF surface, created in pre-processing, for each force. The new functions do not replace the old ones in simulink because they are ~2x slower in the radiation case and significantly slower in the user defined excitation case. See #1394
   - [x] refactor `Convolutionintegral_interp.m`
   - [x] create `convolutionIntegralSurface.m`
   - [x] test to compare `convolutionIntegralSurface.m` to `ConvolutionIntegral_interp.m`
   - [x] create a radiation IRF surface in the body class when variable hydro is used
   - [x] create an excitation IRF surface in the body class when variable hydro is used
   - [x] update library with variant subsystems for the alternate radiation and excitation IRFs
- [x] `GoTo` flags for different signals of interest in variable hydro (and controls): excitation force, total force. Also resolves the need for broken library links in most WEC-Sim applications
- [x] relocates the mass and inertia utilized during the simulation to body.hydroForce. This prevents user confusion around the mass definition, especially when simulations are stopped prematurely, and will enable future work to define a variable mass

Closes issue #1394.


akeeste:
I just need to test this functionality over a variety of cases (variable hydro on/off x radiation IRF and excitation IRF calculations) and it is ready for review

akeeste:
The most recent commit changed `excitationConvolutionIntegralSurface` to use a persistent amplitude which in CML testing was actually ~1.5x faster than loading in the entire wave amplitude time history (since its known ahead of time) and indexing the relevant portion

akeeste:
This PR is ready for review after #1401 is merged and those library changes are brought in

akeeste:
I created a surface an alternate calculation for the excitation IRF without thinking clearly enough. Unlike the radiation IRF, the excitation IRF is acausal. Since the varying state is generally not known ahead of time, we can't create an IRF surface that is used during the simulation. This only affects the `elevationImport` wave with user-defined wave excitation. 

I'll remove the library and bodyClass changes I made for the excitation IRF surface

akeeste:
I removed the excitation IRF and fixed a couple bugs. Once I update the documentation re this feather then the PR will be ready for review

akeeste:
@MShabara This PR is ready for review. I did finish moving the adjusted mass for use during simulation to `body.hydroForce`. I updated the summary up top to reflect that.

akeeste:
@MShabara between #1461 and #1411 I suggest we just merge whatever is ready first and go from there. The library change in #1461 is very small and easy to recreate in either case. I intend to work on this today before I am out

akeeste:
Tests pass and the implementation works on a test case that uses an IRF surface. Merging! Sorry for the delay, I could've gotten this merge done awhile ago but I assumed it would be harder than it was

PR #1419, opened 2025-02-13
Add function to make sure BEM directions span 360 degrees for passive yaw and full directional spectra cases

jtgrasb:
This PR adds a function that makes sure BEM directions span 360 degrees for passive yaw and full directional spectra cases. This prevents extrapolation from occurring.

As of now, if users are using a full directional spectrum with MCR, simu.reloadH5Data needs to be set to 1. This is because I am adding elements to the hydroData structure and it will try to add them again if it is not reloading H5. Open to suggestions on this if there is a better way.


akeeste (changes requested):
@jtgrasb 
To make this compatible with `simu.reloadH5Data=1`, `bodyClass.makeSpanNeg180To180` could update the hydroData direction instead of hydroForce direction, then that change is retained for subsequent MCR runs.

Given that ``makeSpanNeg180To180`` now contains error messages related to full 360 directional coverage, I removed a duplicate message.

see https://github.com/jtgrasb/WEC-Sim/pull/18

jtgrasb:
@akeeste Thanks for the comments and changes! Changing the hydroData directions makes sense, and I merged your PR into this one and tested everything locally to confirm it is still working as expected. Ready to merge once the tests pass as I see it.

akeeste (approved):

akeeste:
@jtgrasb great! tests as passing. merging...

PR #1420, opened 2025-02-17
Updates the Ubuntu and Matlab versions

MShabara:
Updates the Matlab and the Ubuntu versions in the workflow.

Limits the tests to ubuntu-22.04 rather than ubuntu-latest (ubuntu-24.04). Only Matlab 2024x is compatible with ubuntu-24.04.

Noting that the Ubuntu 20.04 runner image will be fully unsupported by github on April 1, 2025


salhus:
Hi @MShabara,

This seems to be a good solution to maintain compatibility. This can be merged.

PR #1423, opened 2025-02-21
elevationToSpectrum.m Function Enhancements

Gusmano-2-OSU:
Makes slight enhancements to the elevationToSpectrum.m function

Now coerces `data` and `time` vectors into the correct shape.

Adds an option for users to remove spectral leakage from the function. Default is `false`.

Fixes miscellaneous warnings.


Gusmano-2-OSU:
I realized I accidentally turned data and time from a row vector to a column vector.

kmruehl:
Thank you @Gusmano-2-OSU. This looks like a great addition. However, can you revise this PR to go into dev instead of main?

kmruehl:
Thank you for changing the PR to dev

MShabara (approved):
@Gusmano-2-OSU  Thank you for submitting this PR, the new feature is useful and coded nicely.

PR #1425, opened 2025-02-25
Fix treatment of the inertia products in body.adjustMassMatrix

akeeste:
This PR fixes a minor bug in the treatment of the inertia products in `body.adjustMassMatrix()`. See https://wec-sim.github.io/WEC-Sim/main/developer/advanced_features.html#theoretical-implementation for the documented reasoning which this update aims to match.


akeeste:
Changing back to draft while I investigate the implementation of added mass adjustments with body interactions

akeeste:
Ready for review again. I confirmed the implementation of our added mass adjustments in b2b cases and added to our documentation explaining it.

kmruehl:
@akeeste after reviewing the proposed documentation revisions, the `bodyClass` revisions in this PR appears to be good, but it would be helpful if you could verbally explain the issue and proposed solution within the PR for future reference/clarity. Thank you!

akeeste:
Correcting my above comment, the `bodyClass` implementation was correct, but our developer manual section on WEC-Sim's added mass treatment was incorrectly writing the inertia tensor as if it was a skew-symmetric matrix. it is not, the inertia tensor is a symmetric matrix. Our source was right, the docs were wrong.

Docs are now corrected.

akeeste:
@kmruehl FYI, see above comment for explanation. The docs build fine so this should be good to merge

kmruehl:
@akeeste thank you for the clarification and for reverting the revisions to the `bodyClass`. Once the tests pass, I will merge.

Issue #1438, opened 2025-03-12
Bug in paraviewVisualization.m


hamzaJayy:
**Note: Please complete all required fields. Italicized text describes each field.**

**Describe the bug (REQUIRED)**
The properties for simu.paraview to set the start, timestep and endtime are being defined inside the for loop of body. Thats means the "if isempty(simu.paraview.startTime) || isempty(simu.paraview.dt) || isempty(simu.paraview.endTime)"  logic only works for first body, never for other bodies. Then "writeParaviewBody" is being called with different "TimeBodyParav". 

![Image](https://github.com/user-attachments/assets/6399559b-a70b-4407-b8ac-cf43702b0905)

I thik it should be like this

![Image](https://github.com/user-attachments/assets/3de8bfac-9926-403e-9cea-5a5b94618baa)

**Workstation information (REQUIRED):**
_Describe the workstation:_
 - Operating System: [e.g., iOS8.1, Windows 10, Linux 5.12.1]
 - [MATLAB/Simulink Release](https://www.mathworks.com/help/matlab/release-notes.html):  2024a
 - [WEC-Sim Release](http://wec-sim.github.io/WEC-Sim/man/release_notes.html):  v6.1

**WEC-Sim file(s):**
paraviewVisualization.m

**How to reproduce:**
set simu.paraview.dt > simu.dt and run rm3 model with moordyn. Then in paraview, load the paraviewfiles, check the information of each file. First body will have different time vector than the rest. 

**Expected behavior**
in paraviewfiles, every entity has the same time vector

**Screenshots:**

![Image](https://github.com/user-attachments/assets/7725e911-1189-4a44-9da7-0476cbf78b1e)

![Image](https://github.com/user-attachments/assets/cddda749-4283-4bbf-80e1-0eb9235f5b0c)

**Additional information:**
same issue in wecsim 6.0

jtgrasb:
@hamzaJayy Thanks for catching this issue. I agree with your proposed solution. Are you able to create a pull request to the main branch to fix this bug? Please let me know and I could review it. If not, I can apply the fix myself.

hamzaJayy:
@jtgrasb I haven't done this before. I tried through dev branch, but it says I dont have authority to do so. Please go ahead and apply the fix. Thanks

jtgrasb:
No problem, I created PR #1441.

jtgrasb:
Resolved by PR #1441

PR #1441, opened 2025-03-13
Fix paraview output function

jtgrasb:
This PR addresses #1438 which found a bug in the paraview output function. If `simu.paraview.dt` is larger than `simu.dt`, the paraview timestep will only be applied to the first body. To fix this, I moved the if statement to before the bodies are looped through and moved the time vector selection outside of the if statement so it is applied for all bodies.

Thanks for finding this issue and suggesting a fix @hamzaJayy.

### DIFF
--- a/source/functions/paraview/paraviewVisualization.m
+++ b/source/functions/paraview/paraviewVisualization.m
@@ -18,31 +18,34 @@ if simu.paraview.option == 1
             end
         end
     end
-   % bodies
+    if isempty(simu.paraview.startTime) || isempty(simu.paraview.dt) || isempty(simu.paraview.endTime)
+        if isempty(simu.paraview.startTime)
+            simu.paraview.startTime = simu.startTime;
+        end
+        if isempty(simu.paraview.dt)
+            simu.paraview.dt= simu.dt;
+        end
+        if isempty(simu.paraview.endTime)
+            simu.paraview.endTime = simu.endTime;
+        end
+        NewTimeParaview(:,1) = simu.paraview.startTime:simu.paraview.dt:simu.paraview.endTime;
+    end
+    % bodies
     filename = [simu.paraview.path filesep 'bodies.txt'];
     mkdir([simu.paraview.path])
     [fid ,errmsg] = fopen(filename, 'w');
     vtkbodiesii = 1;
+   
     for ii = 1:length(body(1,:))
         if body(ii).paraview == 1
             bodyname = output.bodies(ii).name;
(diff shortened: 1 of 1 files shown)


jtgrasb:
@kmruehl I also amended our post-processing for the mooring class and the Paraview functions to account for multiple mooring connections. This should be working now with all data included in the output structure and the Paraview folder when turned on. 

One thing that is not set up is loading multiple mooring connections into Paraview. The `WEC-Sim.py` script is set up for the RM3_MoorDyn example which only has 1 MoorDyn connection, so it only reads in the one. I propose that we leave `WEC-Sim.py` as an example for a Paraview input script and suggest that users create their own as needed, since we don't have anyone on the team that actively uses Paraview.

kmruehl:
Based on conversation with @jtgrasb I reverted this PR to only resolve the bug in #1438, and moved the MoorDyn feature development to https://github.com/WEC-Sim/WEC-Sim/pull/1455.

kmruehl:
To very this feature take the RM3 case and set `simu.paraview.option = 1` and `simu.paraview.dt > simu.dt` and checked the resulting timestep in the paraview output files

kmruehl:
@jtgrasb I'm getting the following error when I run the RM3_MoorDyn_Viz applications case with `simu.paraview.dt = 0.5`. The case ran without issue when I did not specify `simu.paraview.dt`

``` matlab
Index exceeds the number of array elements. Index must not exceed 161.

Error in writeParaviewMooring (line 60)
            pt = [moorDyn.(['Line' num2str(iline)]).(['Node' num2str(inode) 'px'])(it), moorDyn.(['Line' num2str(iline)]).(['Node' num2str(inode) 'py'])(it), moorDyn.(['Line' num2str(iline)]).(['Node' num2str(inode) 'pz'])(it)];

Error in paraviewVisualization (line 62)
        writeParaviewMooring(output.moorDyn, modelName, output.moorDyn.Lines.Time, datestr(simu.date), mooring.moorDynLines, mooring.moorDynNodes,simu.paraview.path,TimeBodyParav,NewTimeParaview)

Error in stopWecSim (line 38)
paraviewVisualization

Error in run (line 112)
evalin('caller', strcat(scriptStem, ';'));

Error in wecSim (line 56)
run('stopWecSim');
 ```

kmruehl (approved):
This is ready for a merge one @jtgrasb confirms the remove vtk dir

jtgrasb:
Also added a couple variables to the clear statement to avoid errors if you run a model with Paraview twice in a row with different settings. Merging once tests pass.

PR #1447, opened 2025-03-20
Dev: fix badBEMIOFix_fcn

dforbush2:
this adds warnings if a peak is detected on the edge of the data set (which would otherwise error) and references the deSpike structure rather than hardcoded defaults.


akeeste:
@dforbush2 can we have this function also remove the peaks on dataset edges? I'm not familiar with it's internal working, but maybe it could truncate data to remove the edge peaks and then extrapolate the smoothed data back to the original frequency range? I'm using this update to fix some bad BEM data right now and am left with sharp peaks at the edges of my data.

dforbush2:
can do...the difficulty internally is that "peak" on the edge sometimes is just the "edge", but I can cook something up.

kmruehl:
@akeeste do you want to review this PR?

dforbush2:
converted to draft: @akeeste suggested fix is easy to implement and incoming soon

dforbush2:
@akeeste try that one. that will despike your endpoints

akeeste:
Thanks @dforbush2 I'll test it out today.
@kmruehl Yes, I can review

PR #1461, opened 2025-04-15
QTF Standing (Newman) Approximation

MShabara:
This PR adds the Newman approximation for the slowly varying component for the second order excitation forces.
The option is enables by enabling `body(i).QTFs = 2`

The calculation was validated against openFAST and OrcaFlex for a spar model related to a FOWT as shown in the figure below:
image


docs/user/advanced_features.rst
```diff
@@ -734,6 +734,68 @@ Applications <https://github.com/WEC-Sim/WEC-Sim_Applications>`_ repository
...
+
+Second-Order Excitation Forces and Quadratic Transfer Functions (QTFs)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+For large offshore structures, second-order wave excitation forces can significantly influence system behavior—particularly in surge, pitch, and other low-frequency motions. These forces arise from interactions between pairs of wave frequencies, :math:`\omega_i` and :math:`\omega_j`, within the wave spectrum.
+
+There are two types of second-order forces:
+
+- **Sum-frequency forces** (:math:`\omega_i + \omega_j`): High-frequency components that can excite fast body motions.
+- **Difference-frequency forces** (:math:`\omega_i - \omega_j`): Low-frequency components that can induce slow-drift responses such as surge or pitch.
```
akeeste, line 747:
```suggestion
- **Difference-frequency forces** (:math:`\omega_i - \omega_j`): Low-frequency components that can induce slow-drift responses, commonly in surge or pitch.

File: docs/user/advanced_features.rst [hunk 2]
```diff
@@ -734,6 +734,68 @@ Applications <https://github.com/WEC-Sim/WEC-Sim_Applications>`_ repository
...
+
+WEC-Sim supports **Quadratic Transfer Functions (QTFs)** to model second-order excitation forces. These can be imported from hydrodynamic solvers such as **WAMIT** and **NEMOH**, provided via `bemio` in an `.h5` file.
```
akeeste, line 749:
```suggestion
WEC-Sim supports **Quadratic Transfer Functions (QTFs)** to model second-order excitation forces. These can be imported from hydrodynamic solvers such as **WAMIT** and **NEMOH**. BEMIO functions will automatically read, process, and write QTF coefficients from WAMIT and NEMOH to the ``*.h5`` file.

File: docs/user/advanced_features.rst [hunk 3]
```diff
@@ -734,6 +734,68 @@ Applications <https://github.com/WEC-Sim/WEC-Sim_Applications>`_ repository
...
+
+----
+
+**File Naming and Conventions**
```
akeeste, line 753: add subsection characters
```suggestion
File Naming and Conventions
"""""""""""""""""""""""""""""""""

File: docs/user/advanced_features.rst [hunk 4]
```diff
@@ -734,6 +734,68 @@ Applications <https://github.com/WEC-Sim/WEC-Sim_Applications>`_ repository
...
+  ::
+  
+      OUT_QTFP_N_<body#>.dat   (for sum-frequency QTFs)
+      OUT_QTFM_N_<body#>.dat   (for difference-frequency QTFs)
+
+- NEMOH uses **0-based indexing** for body numbers.
+
+----
+
+**Activating QTFs in WEC-Sim**
```
akeeste, line 784: add subsection characters
```suggestion
Activating QTFs in WEC-Sim
""""""""""""""""""""""""""""""""

File: docs/user/advanced_features.rst [hunk 5]
```diff
@@ -734,6 +734,68 @@ Applications <https://github.com/WEC-Sim/WEC-Sim_Applications>`_ repository
...
+
+- ``body(i).QTFs = 1`` — Enables **full QTF-based** second-order excitation.
+- ``body(i).QTFs = 2`` — Enables **Newman approximation** for **difference-frequency** forces.
+  - In this case, sum-frequency components are still computed using full QTFs.
+
+.. note::
+
+   Do **not** use ``body(i).QTFs`` and ``body(i).meanDrift`` together.  
+   This will result in **double-counting** of mean drift forces—once from the QTFs and once from linear mean drift.
+
```
akeeste, line 796: add onto the note about double counting
```suggestion
When attempting to use ``body(i).QTFs`` and ``body(i).meanDrift`` together, a warning is thrown and the standalone mean drift flag is turned off. Only the QTFs will be used to avoid double-counting the mean drift forces.

File: source/objects/bodyClass.m [hunk 6]
```diff
@@ -781,92 +786,116 @@ function qtfExcitation(obj, wavesObj, iH)
...
+                        % Sanity check: match sizes before assigning
+                        assert(isequal(size(ampFreqExtended), size(QTF.diffDiag)), 'QTF: Size mismatch in Newman Approx. lines');
+
+                        Lp = zeros(N,1);
+                        Ln = zeros(N,1);
+                        Lp(1:N/2) = ampFreqExtended(1:N/2) .* sqrt( max(QTF.diffDiag(1:N/2), 0));
+                        Ln(1:N/2) = ampFreqExtended(1:N/2) .* sqrt(-min(QTF.diffDiag(1:N/2), 0));
+
+                        % Compute slow-varying forces
+                        fSlowNewman = abs(2*ifft(Lp)).^2 - abs(2*ifft(Ln)).^2;
```
akeeste, line 888: Should this also include a term that is the order of magnitude of the frequency difference?

source/objects/bodyClass.m
```diff
@@ -781,92 +786,116 @@ function qtfExcitation(obj, wavesObj, iH)
...
+
+                        QTF.diffDiag = interp1(omegaCoarse, real(diag(QTF.diffExtended(:,:,n), 0)), omegaFineExtended)';
+
+                        % Sanity check: match sizes before assigning
+                        assert(isequal(size(ampFreqExtended), size(QTF.diffDiag)), 'QTF: Size mismatch in Newman Approx. lines');
+
+                        Lp = zeros(N,1);
+                        Ln = zeros(N,1);
+                        Lp(1:N/2) = ampFreqExtended(1:N/2) .* sqrt( max(QTF.diffDiag(1:N/2), 0));
+                        Ln(1:N/2) = ampFreqExtended(1:N/2) .* sqrt(-min(QTF.diffDiag(1:N/2), 0));
```
akeeste, line 885: Under the square root, why is the maximum and minimum of ``QTF.diffDiag`` used? Shouldn't there also be a factor of 2 in the square root?

akeeste (changes requested):
Thanks @MShabara 

I added a few minor suggestions on the documentation updates.

Primarily, I have a couple of questions on the calculations added in the bodyClass.m. I'm referencing slides you presented last year on the Newman approximation but see some differences.

Additionally:
- Yes, the warning to catch mean drift and QTF at the same time is included.
- Simulink GUI is not yet updated for QTF=2. It should be a quick addition to the drop-down menu

source/functions/BEMIO/QTFs/triToFullMatrix.m
```diff
@@ -42,7 +42,7 @@
                 F_ij(dof_data(i, end-1), dof_data(i, end)) = elements(i);
                 F_ij(dof_data(i, end), dof_data(i, end-1)) = conj(F_ij(dof_data(i, end-1), dof_data(i, end)));
             else
-                elements = dof_data(:, col) - 1i .* dof_data(:, col + 1);
+                elements = dof_data(:, col) + 1i .* dof_data(:, col + 1);
```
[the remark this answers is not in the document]
MShabara: The `dof_data` array stores the QTF coefficients and other parameters generated by the BEM solvers, with the excitation coefficient real part in one column and the imaginary part in the next. Line `45` combines these two columns to form the full complex coefficient for the QTF `sum` components. 
I believe this error in line `45` was introduced when the NEMOH QTF feature was added in BEMIO earlier this year.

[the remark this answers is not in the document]
MShabara: The Simulink GUI is updated for `QTF=2`.

source/objects/bodyClass.m
MShabara: This is part of the Newman Approximation, which is summarized in the figure below:
image

Note that the equations implemented in lines 884 and 885 are in the frequency domain, while the equations in the figure represent their equivalent form in the time domain.

In the Newman approximation, if the term under the square root is negative, multiply it by -1 before taking the square root. If it is positive, proceed with the square root as usual and apply it directly in the time-domain operation.

akeeste: @MShabara 
On the factor of 2 -- I see this in line 888. In line 888, the factor of 2 should be outside of the exponent. Right now a factor of 4 is applied in line 888

On the positive/negative under the square root -- I am following how this is applied differently to get L+ and L-. However why do you also take the maximum and minimum across the frequency dimension?

MShabara: That's a really good catch @akeeste and it took me some time to figure out.

1- There's an additional multiplication by an extra 2 that comes from using the `ifft` function. By definition, the Fourier transform is two-sided and symmetric around zero frequency, with energy equally distributed across the positive and negative halves of the spectrum (each carrying half the signal energy to preserve total energy).

When we switch to a one-sided Fourier representation—as is often done for real-valued signals—we need to multiply the positive-frequency amplitudes by 2 to account for the energy from the negative side. This ensures that the reconstructed signal amplitude remains accurate.

akeeste: Thanks @MShabara. I think I'm following, given the differences in MATLAB fft/iffts, and the frequency-domain time-domain implementations. We should eventually add a section to our theory manual that ties the external references to our implementation. For now, can you add link to the reference you are using?

Given that the current use of MATLAB's ``ifft`` function requires a factor of two, aren't we then missing a factor of 2 that is removed from lines 884-885?

```
% incorporate factor of 2 under the sqrt() as in the derivation
Lp(1:N/2) = ampFreqExtended(1:N/2) .* sqrt( 2*max(QTF.diffDiag(1:N/2), 0)); 
Ln(1:N/2) = ampFreqExtended(1:N/2) .* sqrt(-2*min(QTF.diffDiag(1:N/2), 0));

% multiple each ifft output by 2 to account for the one-sided signal
fSlowNewman = abs( 2*ifft(Lp) ).^2 - abs( 2*ifft(Ln) ).^2;
```

MShabara: Sorry for the confusion, the current implementation is Standing approximation. 
**Standing approximation** proposed a modified form of **Newman's approximation** that, while mathematically equivalent to the original, is expressed differently and tailored to accommodate multidirectional wave environments. Unlike Newman's original formulation, Standing's version does not necessitate high-frequency filtering after application.

image
 
 More information can be found in line `1130` in [OpenFast QTF implementation code ](https://github.com/OpenFAST/openfast/blob/main/modules/hydrodyn/src/WAMIT2.f90)

akeeste (approved):
thanks for clarifying @MShabara

PR #1464, opened 2025-05-06
Bug fix for drag and nonhydro body block mass 

akeeste:
This PR fixes a minor issue (only in dev) from the library merge in #1411. It reverts the nonhydro and drag body blocks back to using ``body.mass`` instead of ``body.hydroForce.hf1.mass``. body.hydroforce is limited for those body types and doesn't contain all of the necessary information. Also variable hydro is not used for those bodies, so we may as well use body.mass directly


akeeste:
Tagging you @jtgrasb since this builds on #1411 which you reviewed

akeeste:
Will wait on this pending @jtgrasb's upcoming PR on allowing variable masses which will likely absorb this on dev

jtgrasb:
Upcoming PR is completely different. I checked this one with the nonhydro body Applications case and it worked as expected. Merging now.

PR #1466, opened 2025-05-14
Universal PTO joint

akeeste:
This PR adds a universal joint to the PTO library. There are no changes to the other PTO library blocks.

This is required for a TEAMER award. The PTO hard stop and extension parameters were updated to allow for inputs of size [1 n] in this case where multiple joint DOFs have motion.


akeeste:
@jniffene I looked into this problem further and don't have a solution yet. I don't think the discrepancies described above are any of our doing. The position, velocity, acceleration, have appropriate values for the universal PTO. I still see that forceConstraint has a very small non-zero value in pitch (~1e-6), and forceInternalMechanics has the same in yaw (~1e+1).

Maybe this is due to how the multiple body system is solved numerically and it retains some very small discrepancies in those quantities? I have some matlab office hours on thursday this week that I will attend and bring this up

akeeste:
I have a meeting with Mathworks next week to better understand this universal joint block. I will update more after that meeting.

akeeste:
Attaching a zip file of the test case for reference:
[RM3_test.zip](https://github.com/user-attachments/files/20981672/RM3_test.zip)

MShabara:
@akeeste Do you think we need to include the Run from Simulink feature for this block?
Also, should we add a description of the new block and its related inputs to the user guide for clarity?

akeeste:
The constraint force and internal mechanics forces here require special handling due to how they're output, but for this particular joint I think those are likely of lower importance for a user than having the joint that allows the correct motion. I suggest we merge this with that understanding so that the joint is available to use. It can be updated in the future.

kmruehl:
Thank you @akeeste!

akeeste:
@jniffene I believe in our last reviews this PR was otherwise ready to go aside from understanding the force output. Since we more information on that and can confirm that the block is working correctly, I will merge this PR.

kmruehl:
@akeeste we discussed this PR during our morning checkin, it's ready for a merge.

Issue #1329, opened 2024-09-19
[BUG] Irregular waves PM spectrum generation in parallel


dododododo21:
**Note: Please complete all required fields. Italicized text describes each field.**

**Describe the bug (REQUIRED)**
When I generate irregular waves with the PM spectrum with a given phaseSeed (in my case = 2), I get two different wave elevations if I am running wecSim and wecSimPCT.

**Workstation information (REQUIRED):**
_Describe the workstation:_
 - Operating System: Windows 11
 - [MATLAB/Simulink Release](https://www.mathworks.com/help/matlab/release-notes.html): R2022b
 - [WEC-Sim Release](http://wec-sim.github.io/WEC-Sim/man/release_notes.html): v6.0

**WEC-Sim file(s):**
I don't know.

**How to reproduce:**
I tried to write a very simple code, which should keep all the constants as I described them in my InputFile:
clear all 
close all

%% run normal
wecSim

figure
hold on;
phaseSeed = waves.phaseSeed;
plot(waves.waveAmpTime(:,1),waves.waveAmpTime(:,2), 'DisplayName', ['Normal phaseSeed=', num2str(phaseSeed)])

%% create .mat file
mcr.header = ["waves.height", "waves.period"];

mcr.cases(1 , 1) = waves.height;
mcr.cases(1, 2) = waves.period;

save MCR_cases.mat mcr

%% run in parallel
wecSimPCT
load('savedData001.mat')
phaseSeed = waves.phaseSeed;
plot(waves.waveAmpTime(:,1),waves.waveAmpTime(:,2), 'DisplayName', ['Parallel phaseSeed=', num2str(phaseSeed)])
xlabel('time [s]')
ylabel('Generated waves [m]')
legend

**Expected behavior**
I expect the same waves.

**Screenshots:**
![grafik](https://github.com/user-attachments/assets/0b9bc989-8e80-4203-b072-9e95ee96243f)

**Additional information:**
Maybe I made another mistake or it's already resolved in the new version, but after a day of figuring out my differences of the results I calculated, I thought it was relevant to post.

dforbush2:
Hi @dododododo21 

Can you please confirm that if you run wecSim multiple times with phaseSeed =2, and if you run wecSimPCT with phaseSeed = 2, each of those runs are identical? In other words, the first run of wecSim is equal to the second run of wecSim, and the first run of wecSim PCT is equal to the second run of wecSimPCT?

[...]

PR #1467, opened 2025-05-14
Fix wave generation with phaseSeed in waveClass for PCT

jleonqu:
This PR updates the random number generation in `setWavePhase` in the `waveClass` to use `RandStream` with `Substream`, instead of `rng(phaseSeed)`. The change ensures deterministic and reproducible wave elevation profiles across both serial (`wecSim`) and parallel (`wecSimPCT`) simulations when `phaseSeed` is specified. This resolves the issue where parallel runs with the same seed could produce different results due to independent worker streams (see Issue #1329 ). The new approach uses the `Threefry` generator, which supports substreams and is recommended by MathWorks for parallel reproducibility.

This PR can be tested using the MCR example available here: WEC-Sim_Applications/Multiple_Condition_Runs/RM3_MCROPT3 running the attached script. 
[irregularWaveGenerationExample.zip](https://github.com/user-attachments/files/20213653/irregularWaveGenerationExample.zip)

The wave profiles generated with `wecSim` and `wecSimPCT` should be identical when using the same seed value.

### DIFF
--- a/source/objects/waveClass.m
+++ b/source/objects/waveClass.m
@@ -598,7 +598,9 @@ classdef waveClass<handle
             % Sets the irregular wave's random phase
             % used by: :meth:`waveClass.setup`.
             if obj.phaseSeed ~= 0
-                rng(obj.phaseSeed);     % Phase seed = 1,2,3,...,etc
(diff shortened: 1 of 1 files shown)


kmruehl:
This is great! Thank you @jleonqu!

PR #1465, opened 2025-05-13
Enable Variable Mass

jtgrasb:
This PR enables mass and buoyancy properties to vary using the variable hydrodynamics feature. Actually varying the mass requires replacing the File Solid block ("Body Properties") with the Simscape "General Variable Mass" block. This is not done in the source library and is demonstrated in the application case below. The plan is to merge this PR then finalize the applications case afterwards.

The body class is edited to include variable mass, inertia, and inertia products inputs and to include these factors (plus center of gravity) in the body.hydroforce structure. The library and docs are updated accordingly. 

As of now, variable mass is not verified for NL hydro and an error message has been added.

PR #1468, opened 2025-05-15
Fix mask for flex body

jtgrasb:
@akeeste This PR fixes the mask for the flex body to align with the changes from #1465 to the rigid body.


akeeste (approved):
approved, thanks @jtgrasb!

Issue #1471, opened 2025-05-15
[Developer Issue] Fix regression tests for RM3


jleonqu:
There are some issues related to the regression tests for the RM3 example.

1. The first issue is that the files that are being used as reference to compared and verify the wecSim outputs don't have the required data for the comparison.
2. The tests were run to compare results from different WEC-Sim versions against the current development branch. Many tests are currently failing. The versions compared were v4.4 and v6.1.2. The timeline of the versions that were compared and their possible cause for failing are shown in the next timeline:

![Image](https://github.com/user-attachments/assets/4c844c66-212b-4621-bb1d-e432107191ff)

We need to:

1. Check why the tests are failing. The issue may be related to very tight tolerances in the tests.
2. Update the reference files to ensure they include the required data for accurate comparison with current WEC-Sim outputs.

cornerzdc:
why my  regressionTest is error ？



wecSimTest
Running bemioTest
......
Completed bemioTest

Running regressionTest

WEC-Sim Input From Standard wecSimInputFile.m Of Case Directory...
WEC-Sim: An open-source code for simulating wave energy converters
Version: 6.1.2

Initializing the Simulation Class...
Case Dir: F:\academic\code\simiturbine_oc2025\WEC-Sim-main\WEC-Sim-main\tests\RegressionTests\RegularWaves\regular
Elapsed time: 0.272462 seconds.

WEC-Sim Pre-processing...
Infinite water depth specified in BEM and "waves.waterDepth" not specified in input file.
Set water depth to 200m for visualization.
Elapsed time: 0.220354 seconds.

WEC-Sim Simulation Settings:
Start Time (sec) = 0
End Time (sec) = 200
Time Step Size (sec) = 0.1
Ramp Function Time (sec) = 100
Number of Time Steps = 2000

Wave Environment:
Wave Type = Regular Waves (Constant Hydrodynamic Coefficients)
Wave Height, height (m) = 2.5
Wave Period, period (sec) = 8

List of Body:
Number of Hydro Bodies = 2

***** Body Number 1, Name: float *****  
Body CG                          (m) = [0,0,-0.72]  
Body Mass                       (kg) = 725833  
Body Moments of Inertia       (kgm2) = [2.09073E+07,2.13061E+07,3.70855E+07]  
Body Products of Inertia      (kgm2) = [0,0,0]  

***** Body Number 2, Name: spar *****  
Body CG                          (m) = [0,0,-21.29]  
[...]

PR #1473, opened 2025-05-16
Fixing Regression Tests for RM3

jleonqu:
This PR is solving Issue #1471 
The baseline files for the RM3 regression tests have been updated and verified.
The tolerance value of 1e-10 was too tight, which was causing the tests to fail. The tolerance value has been updated to 1e-5.
The printPlotRegular.m and printPlotIrregular.m scripts have been updated to load the WEC-Sim run data that contains the most recent run and the results from the baseline file for comparison.


kmruehl (changes requested):
@jleonqu Thank you for catching and resolving this error! My only recommendation is that the `_org.mat` files are very large. We only need `output.bodies`, so to save space I recommend resaving these files so that they only include the body class output. I'm also going to make some minor formatting revisions to these scripts.

kmruehl:
While reviewing this PR we noticed that the scripts were saving empty figures and that the comparison plots were not working properly. I pushed some revisions that resolve the issues with plotting the figures, but they may cause issues with the runner. I'll wait to see if the tests pass...

kmruehl:
The default appears to be that the figures are generating, which means that these tests should fail. This issue needs to be resolved.

kmruehl:
The test are passing, and this works for me. @jleonqu can you confirm? If it works for you, we can merge the PR

jleonqu:
All tests passed locally and the figures were generated correctly. The PR is ready to merge.

kmruehl:
Thanks @jleonqu I have a few minor edits that I'm going to push and then I'll merge this PR

Issue #1450, opened 2025-03-30
[Feature] On the extrapolation of radiation damping loads calculated by ConvolutionIntegral_interp.m


TianyuanWangi:
**Note: Please complete all required fields. Italicized text describes each field.**

**Describe the requested feature (REQUIRED):*
_Provide a clear and concise description of the requested feature._

Hi, there

I've recently been using **persistent** to store data in Simulink Matlab function. I found the radiation damping loads calculated by ConvolutionIntegral_interp.m remain unchanged during time updates (RK4 or others). I suggest using quadratic extrapolation to solve this problem. 

It need to save the previous **3 time step and load results** in ConvolutionIntegral_interp.m. I have tested it successfully on my model. If you think this is a reasonable suggestion, I'll be happy to share the modified model.

Best wishes,

Tianyuan

**Related issue:**
_Does this feature request relate to an issue? If so, provide a clear and concise description of the issue. For example, "I'm try to [...]", and (if applicable) reference the related issue or pull request._

**Describe alternatives solutions:**
_Are there other possible solutions? If so, provide a clear and concise description of possible alternative solution(s)._

**Related  application:**
_Does this feature request relate to a specific application? If so, describe the application so the WEC-Sim team can better understand your goal(s)._

**Additional information:**
_Provide any other additional information._

MShabara:
@TianyuanWangi 

Thank you for the suggestion, but before we discuss creating a PR I would like to draw your attention to the following.

The radiation force calculation involves two main integrations:
1️⃣ First Integration in BEMIO
The first integration is performed in BEMIO, which computes the radiation function `Kr​` based on hydrodynamic data. The result is a vector Kr that contains values for all time steps defined in the `radiationIRF` function.
<img src="https://github.com/user-attachments/assets/c9d34f14-5113-4ee4-98a2-0149c320cba4" width="25%">

2️⃣ Passing Data to WEC-Sim
The computed radiation function `Kr` is then passed into WEC-Sim, specifically into `bodyClass.m`, where it is used in the `convolutionIntegral_interp.m` function to calculate the radiation force.
<img src="https://github.com/user-attachments/assets/10d913d6-3a20-43c3-b2b1-f8eb3987c70b" width="20%">

3️⃣ Second Integration in `convolutionIntegral_interp.m`
1. Within convolutionIntegral_interp.m, a second integration takes place. 
2. At `t=0`: The variable `velocity_history` is initialized to store past velocities.
[...]

PR #1476, opened 2025-05-16
Radiation damping load extrapolation

TianyuanWangi:
Add a feature to extrapolate the radiation damping loads. Since convolutionIntegralInterp and convolutionIntegralSurface use persistent variables to track velocityHistory over time, radiation damping loads cannot be updated at each sub-time step.

See https://github.com/WEC-Sim/WEC-Sim/issues/1450


TianyuanWangi:
Hi @akeeste 
I hope my work helps. Sorry I changed some variable and file names, and now the inconsistent formatting is driving me nuts.

akeeste (changes requested):
Thanks @TianyuanWangi 

I pushed a couple minor formatting changes in extrapFrad and updated the name of ``convolutionIntegralInterp` in a couple other places.

My main request before merging:
`extrapFrad` could largely be simplified to call MATLAB's ``interp1``, instead of manually defining the interpolation. That will remove the if statements, making the function cleaner and faster, and enable different interpolation schemes in the future.

Can you update the function to do this? If `timeHistory` in `convolutionIntegralInterp` and `convolutionIntegralSurface` is initialized to `[-2*simu.cicDT -1*simu.cicDt 0]` instead of zeros(3,1), `interp1` can be called as is

TianyuanWangi:
Hi @akeeste 

I think you propose a better solution. Now the codes in extrapFrad.m is simplified to 
```
Frad =  interp1(timeExtrap,FradExtrap,time,'pchip','extrap')';
```

I used 'pchip' instead of 'spline', as the latter is is very unstable during the first few time steps when only limited data is available.

Since simulink doesn't have an interpolation function, I have to implement it via matlab function.

The timeHistory is initialized as 
```
    timeHistory = [-1*(cicTime(2)-cicTime(1));-2*(cicTime(2)-cicTime(1));-3*(cicTime(2)-cicTime(1))];
```
, following the principle that the updated data is the first element.

At the initial time step, it turns to

```
timeHistory = [0; -1*(cicTime(2)-cicTime(1));-2*(cicTime(2)-cicTime(1))];
```

akeeste (approved):
Thanks @TianyuanWangi for your contribution! My apologies for the long timeline on getting this into WEC-Sim. 

I pushed one last commit to add the extrapFrad function to the flex body and save the library to R2020b

Issue #1443, opened 2025-03-14
[Feature] Calculation reference point of QTF matrix


TianyuanWangi:
**Note: Please complete all required fields. Italicized text describes each field.**

**Describe the requested feature (REQUIRED):*
_Provide a clear and concise description of the requested feature._

Hi, developers

I find that the calculation reference point of QTF matrix is the COG of the floating body. Will you add the additional option to specify a separate reference point for the 2nd order wave exciation calculation?

For the floating offshore wind turbines, the calculation reference point of QTF matrix is generally set as the COG of the whole system, but not that of the floating platform. Therefore, is it possible to add a coordinate transformation function to the calculation of second-order wave loads to deal with inconsistent calculation reference points. This may involve a coordinate transformation of motions and loads.

Best regards,

Tianyuan

**Related issue:**
_Does this feature request relate to an issue? If so, provide a clear and concise description of the issue. For example, "I'm try to [...]", and (if applicable) reference the related issue or pull request._

**Describe alternatives solutions:**
_Are there other possible solutions? If so, provide a clear and concise description of possible alternative solution(s)._

**Related  application:**
_Does this feature request relate to a specific application? If so, describe the application so the WEC-Sim team can better understand your goal(s)._

**Additional information:**
_Provide any other additional information._

TianyuanWangi:
Yes, my pleasure. But this can involve a lot of changes, which I'll briefly list here

- In bodyClass.m, Add a variable to specify the new body reference point (BRP) of 2-nd excitation load
- In WECSim_Lib_Body_Elements.slx, transform the displacement, velocity and acceleration at COG to BRP
- In WECSim_Lib_Body_Elements.slx, transform the 2-nd excitation load from BRP to COG
- The coordinate transformations need to be implemented using eulXYZ2RotMat or more accurate methods

Again, I am happy to contribute to WEC-Sim to the best of my ability.

Best regards,

Tianyuan

MShabara:
@TianyuanWangi 

1. I believe simulink will define the body reference frame location during the simulation, so no need to define it in the `bodyClass`.
2. The phase shift in force can be introduced similarly to the first-order excitation force, as shown in [this link](https://wec-sim.github.io/WEC-Sim/dev/user/advanced_features.html#large-x-y-displacements) for the `largeXYDisplacement` feature.
[...]

PR #1477, opened 2025-05-17
QTF Rotation

MShabara:
This PR introduces modifications to both the body library and the `secondOrderExt.m` function in order to account for body rotation when computing second-order wave excitation forces. 

This ensures that the second-order force components are accurately aligned with the body's orientation as it evolves in time, improving the physical fidelity of the force application, especially in simulations with large rotational motions. 

This PR was requested by a user in Issue #1443


kmruehl:
@jtgrasb can you please review this PR and it's associated application, https://github.com/WEC-Sim/WEC-Sim_Applications/pull/86? thank you!

jtgrasb:
@MShabara The implementation makes sense, but I am getting some errors when I try to run it. I believe the body library hasn't been updated to account for the additional function input (`Displacement`) to the `secondOrderExt` function. 

image

It also looks like there is a conflict in the library file. What I would recommend (assuming the only change to the body library needed is to add the Displacement as an input to that function) is to accept the body library from the dev branch, then add the displacement as an input in Simulink.

PR #1478, opened 2025-05-27
Modified sign convention when a cable pretension is defined

ttran18:
@MShabara I did modify the cableClass.m to correct a cable pretension calculation. This modification is only for the cable tension input whereas the cable length input was good to calculate the cable tension output.


kmruehl:
If this is a bug it should go into main, instead of dev

kmruehl:
Thanks @ttran18! @dforbush2 has offered to review this PR. In the meantime can you amend this PR to go into `main`? You should just be able to modify the base, without opening a new PR

kmruehl:
Thank you very much @ttran18

PR #1489, opened 2025-06-23
Refactor badBemioFix_fcn

akeeste:
This PR refactors `badBemioFix_fcn`:
- renames function to `cleanBEM`
- modularizes - the peak smoothing is turned into a single function that is used for all hydro coefficients (A, B, Fex).
- simplifies the interpolation to use the built-in `interp1` for consistency and ease of changing the interpolation method
 - aligns the usage of the function with the rest of BEMIO
    - This is now a standalone function that can be called with any pre-existing hydro structure and it does not read in coefficients on its own
    - IRF and SS coefficients are removed and must be recreated after using this function to fix A, B, Fex
    - IRF and SS function calls are not repeated within the cleaning function, saving time
 - reorganizes the despike structure to fit the function-based method within `cleanBEM`

and fixes a couple bugs:
- the peak smoothing loop for added mass and damping only took the last value and did not iterate from 1:deSpike.N
-  the negative peak finder for imaginary excitation force component was set to "-.1" not "-1."

To compare the largely identical results after the refactor but before a few bug fixes, look at commit 381f8ba

TODO:
- [x] find the citation that this function was based on
- [x] upload comparison script and some sample data

### DIFF
--- a/docs/theory/theory.rst
+++ b/docs/theory/theory.rst
@@ -279,7 +279,7 @@ equations. In general, a linear system is desired such that:
     \dot{X}_{r} \left( t \right) =
         \mathbf{A_{r}} X_{r} \left( t \right) +
         \mathbf{B_{r}} \mathbf{u} (t);~~X_{r}\left( 0 \right) = 0~~ \nonumber \\
-    \int_{0}^{t} \mathbf{K_{r}} \left( t- \tau \right) d\tau \approx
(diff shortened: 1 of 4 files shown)


akeeste:
Note: the most recent version of BEMRosetta filters in a very similar way and can output an .h5 file that WEC-Sim can read

MShabara:
The PR works as expected, I will go ahead and merge this PR

PR #1494, opened 2025-07-01
Update variable mass implementation and add documentation

jtgrasb:
This PR updates the variable mass to follow the workflow:
1. Use variable mass inputs from user.
2. If no variable mass input but variable inertia is present in the input file, assume equilibrium and use the provided variable inertia.
3. If no variable mass or inertia is present, use the specified body mass and inertia for all variable hydro cases.

I also added some documentation for the updated variable mass example ([PR 90 on Applications](https://github.com/WEC-Sim/WEC-Sim_Applications/pull/90)).

### DIFF
--- a/docs/_include/variable_hydro.rst
+++ b/docs/_include/variable_hydro.rst
@@ -72,7 +72,8 @@ every body in a simulation. To implement variable hydrodynamics for a given body
 
     If no mass vector is specified but ``body(1).mass`` is set to equilibrium, the data from H5 
     files will be used to calculate the variable mass based on displaced volume and water density. 
-    However, the inertia vector will still need to be specified or else it will be assumed constant.
+    However, the full inertia vector (``body(1).variableHydro.inertia``) will still need to be 
+    specified or else it will be assumed to be constant and equal to ``body(1).inertia``.
 
 2. Enable Variable Hydrodynamics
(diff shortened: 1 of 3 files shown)


docs/_include/variable_hydro.rst
```diff
@@ -63,7 +63,16 @@ every body in a simulation. To implement variable hydrodynamics for a given body
...
+    will not be used, regardless of the ``option`` flag below. 
+
+    For variable mass cases, a vector can also be used for the variable hydrodynamics mass, 
+    inertia, and inertia products:
+
+        ``body(1).variableHydro.mass = [mass1, mass2, mass3, mass4, mass5];``
+
+    If no mass vector is specified but ``body(1).mass`` is set to equilibrium, the data from H5 
+    files will be used to calculate the variable mass based on displaced volume and water density. 
+    However, the inertia vector will still need to be specified or else it will be assumed constant.
```
akeeste, line 75: In the equilibrium case doesn't `inertia` have to be defined by the user? Otherwise it will be empty and the simulation won't run
```suggestion
    However, the inertia vector will still need to be specified.

akeeste (approved):
Thanks @jtgrasb. A couple minor suggestions to commit and then let's merge this

PR #1516, opened 2025-07-25
OSWEC Test Case - Irregular Waves

jleonqu:
There is an issue when comparing the results between WEC-Sim v4.0 and v6.1.2 for the OSWEC under irregular waves with wave directional spreading. The comparison is shown in the figure below:

image

The next step to determine the main reason for the difference in results is to compare the results for regular waves.


jleonqu:
Yes, we changed how the phase seed is handled. In v4.0, rng(obj.phaseSeed) was used, which is not reproducible in parallel runs because each worker has an independent RNG stream. The new implementation uses RandStream('Threefry') with substreams to ensure deterministic and reproducible wave profiles across both serial (wecSim) and parallel (wecSimPCT) simulations. The issue was identified in #1329 and it was solved in PR #1467

kmruehl:
Wonderful news, thanks @jtgrasb!

jtgrasb:
After the tests failed and more investigation, there seems to be a small difference between the irregular wave results in v4.0 and main/dev. The difference mainly occurs at the extents of motion. This difference is not observed for regular or regularCIC waves, but it is observed for both the RM3 and OSWEC. Below shows the difference for the RM3 response in irregular waves.

image

PR #1515, opened 2025-07-25
`waveClass` Constructor Improvements

pm-gusmano:
## Summary
Improves the `waveClass` constructor by replacing string-based type validation with a MATLAB enumeration. This implementation suggests the different wave types to users as they construct `waveClass` objects, and reduces boilerplate code.

## Changes
- **Modified**: `source/objects/waveClass.m` - Updated constructor to use `arguments` block with `internal.waveType` validation
- **Added**: `source/objects/+internal/waveType.m` - New enumeration defining all valid wave types

## Benefits
- **Type Safety**: Compile-time validation prevents invalid wave type strings
- **Better IntelliSense**: IDE can provide autocomplete for valid wave types  
- **Future-proof**: Easy to add new wave types by extending the enumeration
- **Backward Compatible**: Existing functionality unchanged

PR #1525, opened 2025-08-07
[BUG] Fix paraview outputs for new wave types

jtgrasb:
With PR #1515, the wave type is changed slightly from a string to an enumerator. When outputting to Paraview, this means that the wave type needs to be converted to a character array before printing.


MShabara (approved):
Thank you for catching this. The PR looks good to me. I will go ahead and merge it.

MShabara:
@jtgrasb I just noticed this, do we need to do the same changes to `writeParaviewResponse.m`?

PR #1526, opened 2025-08-07
Fix for the elevation plot when multiple wave classes are defined

MShabara:
This PR resolves an error that occurs when plotting the wave elevation with multiple wave class instances. The implementation now correctly sums the `waveAmpTime(:,2)` values from all wave objects before plotting, ensuring the total wave elevation is accurately represented.

### DIFF
--- a/source/objects/waveClass.m
+++ b/source/objects/waveClass.m
@@ -550,12 +550,16 @@ classdef waveClass<handle
                 rampTime double {mustBeReal, mustBeNonNan, mustBeFinite} = 0
             end
 
+            waveElevationSum = zeros(size(obj(1).waveAmpTime(:,2)));
+            for i = 1:numel(obj)
+                waveElevationSum = waveElevationSum + obj(i).waveAmpTime(:,2);
+            end
             figure
-            plot(obj.waveAmpTime(:,1),obj.waveAmpTime(:,2))
+            plot(obj(1).waveAmpTime(:,1),waveElevationSum)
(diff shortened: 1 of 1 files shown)


source/objects/waveClass.m
```diff
@@ -550,12 +550,16 @@ function plotElevation(obj,rampTime)
                 rampTime double {mustBeReal, mustBeNonNan, mustBeFinite} = 0
             end
 
+            waveElevationSum = zeros(size(obj(1).waveAmpTime(:,2)));
```
akeeste, line 553: This works but is quite surprising MATLAB behavior to me. I would have thought that calling an object's method on an array of objects would error, not result in the entire object array being available in the method.

source/functions/initializeWecSim.m
```diff
@@ -306,7 +306,7 @@
 
 % Check for all waves(#) are of the same type
 for iW = 2:length(waves)
-    if strcmp(waves(iW).type, waves(1).type) ~=1
+    if strcmp(char(waves(iW).type), char(waves(1).type)) ~=1
```
akeeste, line 309: suggested simplification to the logic:
```suggestion
    if ~isequal(waves(iW).type, waves(1).type)

akeeste (approved):
Thanks @MShabara. I added a minor simplification to the wave type comparison logic. I will also push a commit that extends your update for `waves.plotElevation()` to `waves.plotSpectrum()`

akeeste (approved):

PR #1528, opened 2025-08-08
Fix spread with nonlinear hydro

jtgrasb:
This PR fixes the nonlinear hydro calculation for cases with wave spreading. The spread term is moved to under the square root to be consistent with our [documentation](https://wec-sim.github.io/WEC-Sim/main/theory/theory.html#frequency-varying-direction-and-spread) and the other irregular excitation functions. 

I tested this to make sure everything runs, but I did not get a chance to test a nonlinear case with spread. A quick test would be to run the OSWEC case with nonlinear hydrodynamics and an irregular wave with spread in a few directions. Then, compare the result to the same irregular wave without spread. The result should show increased excitation for the case without spread (difference depends on the spread values).


MShabara (commented):
Thank you @jtgrasb, the code now matches the documentation for the wave spread.

PR #1566, opened 2025-10-02
Update mooring class, functions, library to MoorDyn v2 API

jtgrasb:
This PR goes with [PR#16 on the WEC-Sim/MoorDyn repo](https://github.com/WEC-Sim/MoorDyn/pull/16). The goal is to fix the issue where [MoorDyn is crashing MATLAB on GitHub actions](https://github.com/WEC-Sim/WEC-Sim_Applications/actions/runs/17800892973/job/50600275103), which seems to be caused by a [popup graphics window that is used by the v1 API](https://github.com/FloatingArrayDesign/MoorDyn/issues/327). This PR updates to the v2 API which has some different function names and uses a libstruct to define the MoorDyn instance.


akeeste (commented):
@jtgrasb this looks good. A question about whether we can work around using `evalin` statements and one in WEC-Sim/MoorDyn#16 about why we'd wait for a new release

PR #1590, opened 2025-11-11
Addition to #1546, catch Capytaine nan input data

akeeste:
Functionality to catch, warn, and remove low frequency NaN Capytaine data in `readCapytaine()`

The added check looks for when any of the added mass, radiation damping, real excitation, or the imaginary excitation coefficients are NaN in all degrees of freedom at a given frequency. Any frequency at which this happens is removed. 

@mancellin is there any case where NaNs are output at low frequencies that is not caught by the logic here?

### DIFF
--- a/tests/bemioTest.m
+++ b/tests/bemioTest.m
@@ -72,7 +72,6 @@ classdef bemioTest < matlab.unittest.TestCase
         function testCombineBEM(testCase)
             hydro(1) = testCase.wamitHydro;
             hydro(2) = testCase.nemohHydro;
-            hydro(3) = testCase.capytaineHydro;
             testCase.comboHydro = combineBEM(hydro);
         end
         function testIRF(testCase)


source/functions/BEMIO/readCAPYTAINE.m
```diff
@@ -505,6 +505,34 @@
 
 waitbar(8/8);
 
+%% Catch and remove nan values
+lowFrequencyNanMask = squeeze(all(isnan(hydro.A),[1 2])) | ...
```
akeeste: At this point the coefficients have dimensions of DOF, DOF (or size 1 for excitation), n frequencies.
So this line checks that a each frequency for data that is entirely NaN. If this happens in any of the hydrodynamic coefficients, those frequencies are removed.

mancellin:
I should document this somewhere, but here are all the possible occurrences of a NaN in the dataset (using a standard setup):

1. Low frequency / low depth ($kh < 0.1$), then added mass, radiation damping and diffraction force are NaN (not sure about Froude-Krylov, might be defined).
2. In the limit cases $\omega = 0$ or $\omega = \infty$:
    - the diffraction force and Froude-Krylov force are NaN
    - the added mass and radiation damping are always defined (not NaN) for $\omega = \infty$
    - the added mass and radiation damping are defined for $\omega = 0$ in infinite depth, but they are NaN in finite depth (cf. rule 1.).
3. With forward speed, if the encounter frequency is zero, all outputs are NaN.

(and NaN is used as a fallback value in case of unknown error or bug.)

mancellin:
> is there any case where NaNs are output at low frequencies that is not caught by the logic here?

So to answer your question: that looks good to me, but extract the infinite frequency added mass before if you want it, because you might lose it in the NaN filtering.

jtgrasb:
@akeeste I had a .nc file from Capytaine with some NaNs and was able to try this out. It seems to work as intended, sending a warning message and removing the frequency(s) containing the NaNs.

akeeste:
Approved by @jtgrasb

PR #1600, opened 2025-12-03
Use IEC standard for full directional waves

jtgrasb:
This PR updates the full directional waves (`spectrumImportFullDir`) to follow the IEC standard, which states that a full directional spectrum is defined by an omni-directional spectrum ( $S(f)$ ) and a directional spread ( $D(f,\theta)$ ) and the integral is equal to 1 from -pi to pi. The previous implementation was compatible with waves from Ocean Observatories Initiative (OOI) for the Pioneer Buoy model, but not completely following the IEC standard.

image

I also updated the documentation and developed [two applications cases](https://github.com/WEC-Sim/WEC-Sim_Applications/pull/103) to demonstrate the use of the full directional spectrum and conversion from OOI.


akeeste (approved):
Thanks @jtgrasb, this looks great! I pushed a small commit to clarify a docs line and remove the unused `A` variable in `irregExcFullDirF.m`. The applications both work. One comment in https://github.com/WEC-Sim/WEC-Sim_Applications/pull/103 on perhaps consolidating the two cases which would only affect a docs link here

jtgrasb:
Thanks @akeeste. I made some updates to resolve the conflicts and adjust the documentation for a single application folder. Merging now.

PR #1601, opened 2025-12-09
Fix calcDispPhase for large xy displacements feature

jtgrasb:
There are multiple bug in the large xy displacements feature. 
1. From Falnes, the change in phase with displacement is dependent on the wavenumber, k, x-displacement, and y-displacement. It is not dependent on $\omega$ as was previously being multiplied in `calcDispPhase()`. This PR updates `calcDispPhase()` to calculate the displacement phase as $\phi_{disp} = -k (x cos(\frac{\theta \pi}{180}) + y sin(\frac{\theta \pi}{180}))$.

image

2. The displacement phase previously did not take into account the center of gravity of each body. If the body was not located at the origin, a displacement phase would be incorrectly applied. This PR updates the library block to feed in the relative displacement into the `calcDispPhase()` function. 

I also updated the documentation accordingly and verified the results by comparing to a BEM case with the center of gravity shifted:

image

### DIFF
--- a/docs/user/advanced_features.rst
+++ b/docs/user/advanced_features.rst
@@ -950,7 +950,7 @@ If large lateral displacements (i.e., in the x- or y-direction by the default WE
 
 When :code:`body(i).largeXYDisplacement.option = 1`, the phase of the excitation force exerted on the body is adjusted based upon its displacement as
 
-:math:`\phi_{displacement} = k (x cos(\frac{\theta \pi}{180}) + y sin(\frac{\theta \pi}{180}))`
+:math:`\phi_{displacement} = -k (x cos(\frac{\theta \pi}{180}) + y sin(\frac{\theta \pi}{180}))`
 
 where k is the wavenumber (:code:`waves.wavenumber`), x is the instantaneous displacement relative to the body's center of gravity in the x-direction, y is the instantaneous 
 displacement relative to the body's center of gravity in the y-direction, and :math:`\theta` is the the wave direction in degrees (:code:`waves.direction`). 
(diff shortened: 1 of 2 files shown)


docs/user/advanced_features.rst
```diff
@@ -946,14 +946,15 @@ Large X-Y Displacements
 ^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 By default, the excitation force applied to the modeled body is calculated at the body's CG position as defined in BEM. 
-If large lateral displacements (i.e., in the x or y direction by the default WEC-Sim coordinate system) are expected for the modeled device, it may be desirable to adjust the excitation force to account for this displacment. 
+If large lateral displacements (i.e., in the x- or y-direction by the default WEC-Sim coordinate system) are expected for the modeled device, it may be desirable to adjust the excitation force to account for this displacment. 
 
 When :code:`body(i).largeXYDisplacement.option = 1`, the phase of the excitation force exerted on the body is adjusted based upon its displacement as
 
-:math:`\phi_{displacement} = k \omega x(1)*cos(\frac{\theta \pi}{180}) + x(2).*sin(\frac{\theta \pi}{180})`
+:math:`\phi_{displacement} = k (x cos(\frac{\theta \pi}{180}) + y sin(\frac{\theta \pi}{180}))`
```
akeeste:
```suggestion
:math:`\phi_{displacement} = -k (x cos(\frac{\theta \pi}{180}) + y sin(\frac{\theta \pi}{180}))`

jtgrasb:
@akeeste Thanks for the quick review and adding the tests. I made the added the negative to the documentation, and I also added a negative to the test expected phase (doesn't change result since we wrap to pi) to be accurate (positive displacement means the wave elevation at that position lags the displacement at the origin). Once the tests pass, it is ready to merge from my end!

PR #1603, opened 2025-12-11
Variable hydro with directional waves bugfix

akeeste:
This is a short PR that fixes a a couple small bugs that occur when variable hydro is used with certain directional waves (passive yaw or full directional spectrum). 
- adds the `iH` index to all body.hydroData calls in `bodyClass.m` preprocessing


jtgrasb:
Thanks for the quick fix! This PR looks good, merging now.

PR #1605, opened 2025-12-16
Fix spread in NL instanenous wave elevation

thomash303:
This PR fixes the spread calculation in the instantaneous NL wave elevation. In reference to the same correction made in NL hydro (#1528), where the spread term is moved under the square root, but did not update the same issue in the wave elevation calculation.


akeeste (approved):
Thanks @thomash303 I double checked every other instance of `spread` in all source .m files and documentation .rst files and didn't find any other instances where spread is not under the square root. Approved and merging.

PR #1609, opened 2026-01-20
Update for MoorDyn v2.5

jtgrasb:
This PR updates the mooring and response class to be compatible with MoorDyn v2.5 along with [PR#17 to WEC-Sim/MoorDyn](https://github.com/WEC-Sim/MoorDyn/pull/17). Most importantly, this allows for disabling the graphics console window (`calllib('libmoordyn', 'SetDisableConsole', 1)`) which should prevent GitHub CI issues. Then, we can add the WEC-Sim Applications tests back for MoorDyn cases.


akeeste (approved):
Thanks @jtgrasb. This is working for me. Approved.

#1618 is actually an alternate fix that overlaps with the change here to the `responseClass.m`. That fix makes the `header` a bit cleaner so I am going to defer to that fix instead and remove the change to the responseClass here

akeeste:
Tests passed again. Merging

PR #1614, opened 2026-02-04
Move .h5 files from lfs to git

### LINKED ISSUE #1610: [Developer issue] Git LFS budget frequently hit
The Actions frequently hit the LFS budget and prevent tests from running. In the past 10 years the max file size that git can handle has increased substantially and our h5 file are much smaller now. 

@akeeste action to look at removing all of our LFS files and replacing them with standard git tracking

akeeste:
This PR removes LFS tracking for the examples .h5 files and moves them to normal git tracking. These files are significantly smaller now (~4 MB and 11 MB) and we are hitting the LFS bandwidth on Actions.
- examples/RM3/hydroData/rm3.h5
- examples/OSWEC/hydroData/oswec.h5

The files history is retained fully and intact, so the repo still uses LFS _storage_ but the bandwidth limit to download LFS files using Actions will no longer be hit. 

I committed new versions of each file created using the current version of writeBEMIOH5() on dev so they should be fully up to date.

Process here: https://stackoverflow.com/questions/35011366/move-git-lfs-tracked-files-under-regular-git

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,1 @@
 *.* linguist-language=matlab
-*.h5 filter=lfs diff=lfs merge=lfs -text


akeeste:
A couple ways to review this PR:
- Check that the data in the two .h5 files are identical to the previous version
- Check that rerunning BEMIO for these examples recreates identical h5 files

I screenshotted the `git lfs status` results before the commit was made (below) that confirms the files are moving from an LFS commit hash to a Git commit hash

image

Once merged we can track out Git LFS bandwidth over this month to confirm that we are not using it any longer

jleonqu:
I reviewed this PR following the recommended steps. This is what I did:

* I checked the data in the two .h5 files, they are identical to the previous version
* I reran BEMIO for the examples and I obtained identical h5 files

This is ready to merge.

jleonqu (approved):
I reviewed this PR following the recommended steps. This is what I did:

* I checked the data in the two .h5 files, they are identical to the previous version
* I reran BEMIO for the examples and I obtained identical h5 files
* 
This is ready to merge.

PR #1618, opened 2026-02-11
Prevent problem with leading spaces in the MoorDyn output file header line

luwang00:
Ran into a problem where the leading spaces in the MoorDyn output file header line turning into an empty column name. The problem is fixed by simply removing the leading spaces.


akeeste (approved):
Thanks @luwang00! Approved and merging

PR #1628, opened 2026-03-31
Remove cg difference from variable hydro and improve mask

jtgrasb:
This PR makes 2 updates to fix minor bugs associated with variable hydrodynamics:
- Removes the center of gravity difference as a variable hydro option. Previously, this recorded the difference in center of gravity between each hydrodynamic dataset and the initial one. This was then used as an input to the variable mass block for the variable mass case. However, the hydrodynamic forces need to always be applied at the body's center of gravity, and the rigid transform translates the body to its respective location. By changing the center of gravity, it was altering where/how the hydrodynamics are being applied. Thus, the center of gravity difference input has been removed.
-  The other update is to the mask for the variable hydro selection block. Previously, if you changed the discretization of variable hydrodynamics, it would not delete old lines. This would lead to a messy Simulink subsystem and would cause warnings. The new mask deletes the old lines before adding new ones, cleaning up the subsystem and preventing warnings.
image -> image

Associated with [WEC-Sim Applications PR#109](https://github.com/WEC-Sim/WEC-Sim_Applications/pull/109). Once merged into main, should pull into dev as well. Would be great to have this reviewed before next week's in person meeting if possible.


akeeste (approved):
Thanks @jtgrasb, looks good to me. No changes requested. Merging

PR #1638, opened 2026-05-27
Quality of life improvements

akeeste:
This PR incorporates a variety of quality of life improvements to annoyances, issues, or software behaviors that arose during my variable hydro validation modeling:
- add warning to `axisAngle2RotMat` to catch wrong units (angles greater than pi rad should catch most cases where degrees are incorrectly input instead of radians)
- fix small issues with checking for valid passive yaw directions
- limit the clearing yaw variables in `stopWecSim` to those defined by WEC-Sim
- remove waitbars from BEMIO - these are slow and annoying because they pop up over both MATLAB and non-MATLAB screens. They're largely unnecessary now that BEMIO has been sped up. I converted them to printing updates to the command window which is unobtrusive and faster
- fix issue reloading hydrodata in variable hydro MCR runs
- Utilize body.variableHydro.hydroForceIndexInitial instead of 1 when variable hydro is off to reduce the list of h5 files, hydroStiffness, quadDrag and linearDamping so that the option flag can easily switch between identical cases.
    - e.g. a user loads in 25 depths (in order) between 3m-15m. The initial position uses 9m depth data. The previous behavior would default to 3m data when the varibleHydro.option flag is off. Now it will default to use hydroForceIndexInitial, which will point to the same initial position (9m in this case)


jtgrasb:
Not sure if this should be included here, but I also noticed that the plotBEMIO formatting and notes show up weird when using R2025+. Could consider having if statement depending on MATLAB version or moving notes somewhere else.

akeeste:
> Not sure if this should be included here, but I also noticed that the plotBEMIO formatting and notes show up weird when using R2025+. Could consider having if statement depending on MATLAB version or moving notes somewhere else.

Good catch. That entire plotting structure is quite slow and could likely use an overhaul--though at the cost of much time/effort. Perhaps moving to `nexttile` instead of `subplot`. It looks like `text` could be an option instead of `annotation` too

akeeste:
Tests are passing and I reviewed this PR once again. Merging into dev

Issue #1636, opened 2026-05-25
WEC-Sim bodyClass Tutorial Link Restoration


JJJtt32:
Hi WEC-Sim Developers, I noticed that on the official website's online training materials page, the link to the 7th tutorial file, WEC-Sim_BodyClass.pdf, is currently broken or invalid (returning a 404 error or file-not-found asset glitch).

Since the bodyClass is a vital core component of the WEC-Sim object-oriented multi-body dynamics architecture, this tutorial file is highly essential for users to properly understand how dry mass, moments of inertia, and BEM hydrodynamic coefficients are mapped within the class properties. Could you please check the server assets and update or restore this tutorial link?

Where does the error occur:

Official Training Materials Base Website: (https://wec-sim.github.io/WEC-Sim/main/_static/downloads/7_WEC-Sim_BodyClass.pdf)

jtgrasb:
@JJJtt32 Thanks for letting us know. It looks we have the file on the repo, but it is just named incorrectly. I've made PR #1639 to update the filename. In the meantime, here is the [body class pdf](https://github.com/WEC-Sim/WEC-Sim/blob/main/docs/_static/downloads/7%20WEC-Sim_BodyClass.pdf).

PR #1639, opened 2026-05-27
Fix body class tutorial file name

jtgrasb:
This PR is in response to Issue #1636 and fixes the name of the body class tutorial file so the link in the documentation works.


akeeste:
Thanks @jtgrasb! Approved and merging. The tests are not running on main due to hitting the LFS limit

PR #1648, opened 2026-06-30
Update runner versions

akeeste:
- [x] Update WS version in conf.py and simulationClass.m
- [x] Update versions of all actions used in GitHub Actions
- [ ]  Remove git LFS jobs on Actions


akeeste:
The dispatch event will only run on main. Closing this PR and pushing their directly to fix

PR #1653, opened 2026-08-03
 Simulink slow-down issues-add FOH quick fix after Rate Transition blocks in body-elements library

basaucier:
# Update 
## Description:

This draft PR implements a First-Order Hold (FOH) workaround for a variable-step solver slowdown issue discussed with MathWorks.

## Background:

MathWorks/Sandia discussion on variable-step solver slowdowns identified a discrete-to-continuous interface issue in WEC-Sim. In particular, Rate Transition blocks in radiation/convolution paths can introduce discrete signal changes that cause variable-step physical solvers such as `daessc` to reset and recalculate Jacobians. First-Order Hold (FOH) blocks were suggested as a quick workaround.

## Modified file

`source/lib/WEC-Sim/WECSim_Lib_Body_Elements.slx`

## Change made

This change does not remove any existing `Rate Transition` blocks.

First-Order Hold blocks were inserted immediately downstream of selected `Rate Transition` blocks in the Rigid Body wave-radiation/convolution path. The affected signal paths were rewired from:

`Rate Transition -> downstream block`

to:

`Rate Transition -> First-Order Hold -> downstream block`

Each inserted FOH block uses:

`Ts = simu.dt`

where `simu.dt` is the WEC-Sim simulation time-step/sample-time setting from the input file.

## Scope of this narrowed fix

This narrowed fix adds 7 First-Order Hold blocks in:

`Rigid Body / Hydrodynamic Body / Wave Radiation Forces Calculation`

#### 1. Convolution Integral Calculation / Convolution Variant Subsystem / Convolution Integral Calculation

Full path:

`WECSim_Lib_Body_Elements/Rigid Body/Hydrodynamic Body/Wave Radiation Forces Calculation/SS CI and Constant-Damping-CoeVariant Subsystem/Convolution Integral Calculation/Convolution Variant Subsystem/Convolution Integral Calculation`

Blocks added:

- `FOH_after_Rate_Transition` after `Rate Transition`
- `FOH_after_Rate_Transition1` after `Rate Transition1`

#### 2. Convolution Integral Calculation / Convolution Variant Subsystem / Convolution Integral Surface Calculation

Full path:

`WECSim_Lib_Body_Elements/Rigid Body/Hydrodynamic Body/Wave Radiation Forces Calculation/SS CI and Constant-Damping-CoeVariant Subsystem/Convolution Integral Calculation/Convolution Variant Subsystem/Convolution Integral Surface Calculation`

Blocks added:

- `FOH_after_Rate_Transition` after `Rate Transition`
- `FOH_after_Rate_Transition1` after `Rate Transition1`

#### 3. Parent Convolution Integral Calculation

Full path:

`WECSim_Lib_Body_Elements/Rigid Body/Hydrodynamic Body/Wave Radiation Forces Calculation/SS CI and Constant-Damping-CoeVariant Subsystem/Convolution Integral Calculation`

Block added:

- `FOH_after_Rate_Transition` after `Rate Transition`

#### 4. FIR Filter Calculation

Full path:

`WECSim_Lib_Body_Elements/Rigid Body/Hydrodynamic Body/Wave Radiation Forces Calculation/SS CI and Constant-Damping-CoeVariant Subsystem/FIR Filter Calculation`

Blocks added:

- `FOH_after_Rate_Transition` after `Rate Transition`
- `FOH_after_Rate_Transition1` after `Rate Transition1`

## Library coverage

For this narrowed fix:

- Library FOH blocks added: 7
- A two-body Rigid Body OSWEC model inherits 14 FOH blocks

## Validation summary

Testing was performed using Nate's confirmed OSWEC `daessc` slowdown reproducer.

Configuration:

- Model: `OSWEC`
- Solver: `daessc`
- SolverType: Variable-step
- MaxStep: `0.1`
- RelTol: `1e-3`
- End time: `400 s`

## Current status

Based on the confirmed OSWEC `daessc` reproducer, the Rigid Body radiation/convolution FOH group appears to be the smallest robust fix currently supported by testing.

This notes file documents the narrowed 7-block fix.

## Caveats

This OSWEC case does not exercise Flex Body paths. Flex Body FOH remains follow-on validation with the appropriate application case(s).

Current conclusions are based on repeated timing tests with solver settings and FOH block counts verified after each run.

## Additional documentation:
- ``FOH_quickfix_notes.md``


akeeste:
In reviewing this PR, I tried to isolate how many and where the first-order hold (FOH) blocks are required. I noticed an interesting effect that all FOH blocks in the convolution block could actually be removed and the simulation still sped up. I narrowed this down to actually not adding FOH at all, but removing the `Rate Transition` block at `OSWEC/Flap/Hydrodynamic Body/Wave Radiation Forces Calculation/SS CI and Constant-Damping-CoeVariant Subsystem/Convolution Integral Calculation`. The addition of the `extrapFrad` function serves to handle the convolution force during minor time steps, so its output is continuous and rate transition block is not required. It was actually causing problems by forcing the otherwise continuous signal to be converted back to FiM (fixed in minor) type. 

@basaucier. Can you reset the library and branch to be equal to main, make this one change, and test out the OSWEC case again to ensure this works on another system?

akeeste:
Plan from Adam/Rebekah discussion:
- PR has been updated to go into main
- reset body library to the version on main, re-add the change (removal of the rate transition block in the radiation force function) 
- push library in 2020b to this PR

basaucier:
I reset the library back to the `main` version and applied only the requested single-block change in the Rigid Body radiation/convolution path.

Library change made:
- Removed the parent `Rate Transition` block at:

  WECSim_Lib_Body_Elements/Rigid Body/Hydrodynamic Body/Wave Radiation Forces Calculation/SS CI and Constant-Damping-CoeVariant Subsystem/Convolution Integral Calculation/Rate Transition

- Reconnected the original source signal directly to the downstream destination(s).
- No First-Order Hold blocks are added in this version.

Rationale:
- Additional review suggested that the `extrapFrad` function already provides a continuous output during minor time steps, so this specific Rate Transition block is not required.
- The block appeared to be forcing an otherwise continuous signal back into fixed-in-minor behavior.

akeeste (approved):
Thanks @basaucier. The library change and version look good. The OSWEC case now functions as expected with the DAESSC solver

akeeste:
The tests are failing due to some changes from the previous result. The tolerance is fairly tight and it seems possible that removing the rate transition block could be causing this. Assuming they fail again, i will investigate further before merging

akeeste:
Regression tests are failing to alert us that a result is changing slightly. Based on my investigation of the RM3 regression test in regularCIC waves:
image

and regular waves:
image

The CIC results are actually slightly better in the regularCIC case and nearly identical in the irregular case. I suggest that we update the regression test results. I will do that and then merge this PR

akeeste:
@jtgrasb can you review the tests for this PR too and double check my assessment of the changes before we update the test data? Any reason not to update the correct data?

I have prepared new test data locally that can be pushed and merged any time. Also, I haven't added add it to this PR yet, but it would be nice to update the regression tests to make the data easy to overwrite. it's a painful manual process right now

jtgrasb:
@akeeste @basaucier I reviewed the removal of the rate transition block. Took me a bit to understand why the results change, but here's a summary for others:
- Simulink model has continuous states.
  - This means that there are major (i.e., simu.dt) and minor timesteps (when problem is solved between normal timesteps) even when using a fixed step solver.
- The rate transition block forces the signal to be held during minor timesteps, preventing updated radiation force values at intermediate timesteps.
  - Therefore, the new version without the rate transition block should be marginally more accurate by including minor timesteps.

I don't see any reason not to update the test data, since the results without the rate transition block should be more accurate.
 
Anecdotally, I do see a small speed up locally, but maybe there's a way we can quantify the speed up? 

Also, sounds like the original recommendation was to use first order hold blocks to speed up. Do we know why this wasn't successful?

akeeste:
Thanks @jtgrasb. The primary speed-up comes specifically when using the `daessc` solver. The simulation gets stuck in that case but works fine in other cases. Adding a FOH block was Mathworks recommended solution, but actually the rate transition that motivated the FOH block was not required. When #1476 added extrapolation of the radiation force during minor time steps, the extrapolation acts to convert the discrete CIC block back to continuous. So the rate transition actually works against use by forcing the radiation force back to FiM (fixed in minor). Removing it allows Simulink to directly use the continuous output of the radiation force extrapolation

akeeste:
Tests have passed with updated data. Merging!

PR #1654, opened 2026-08-19
Update Capytaine examples for v3

mancellin:
I've been testing the development version of Capytaine before the release of version 3.

This PR is the only thing I had to update to make the examples in the BEMIO directory compatible with the upcoming version 3 of Capytaine. The examples remains also compatible with 2.3.1, and there is no need to update the large datasets computed with 2.3.1.

In the future, the examples could be updated to make use v3-only features for multibody hydrostatics, but for now keeping examples compatible with 2.3.1 might be reasonable.


akeeste (approved):
Thanks for the update @mancellin!

